PR #15, opened 2020-01-16
Add some of the simple pkg/networkservicemesh/common pieces

edwarnicke:
Based on #14

Please only review pkg/networkservicemes/common subdir.


pkg/networkservicemesh/common/setid/client.go
```diff
@@ -0,0 +1,55 @@
...
+// NewClient - creates a new setId client.
+//             name - name of the client
+//             Iff the current pathSegment name != name && pathsegment.id != connection.Id, set a new uuid for
+//             connection id
+func NewClient(name string) networkservice.NetworkServiceClient {
+	return &idClient{name: name}
+}
+
+func (i *idClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*connection.Connection, error) {
+	if len(request.GetConnection().GetPath().GetPathSegments()) > int(request.GetConnection().GetPath().GetIndex()) &&
```
haiodo: @edwarnicke  Could you use local variable it will be more readable.

conn := request.GetConnection()
path := conn.GetPath()
ind := path.GetIndex()

if len(path.GetPathSegments()) > int(ind) &&
	path.GetPathSegments()[ind].GetName() != i.name &&
	path.GetPath().GetPathSegments()[ind].GetId() != conn.GetId() {
	conn.Id = uuid.New().String()
}

pkg/networkservicemesh/common/setid/server.go
```diff
@@ -0,0 +1,55 @@
...
+// NewServer - creates a new setId server.
+//             name - name of the client
+//             Iff the current pathSegment name != name && pathsegment.id != connection.Id, set a new uuid for
+//             connection id
+func NewServer(name string) networkservice.NetworkServiceServer {
+	return &idServer{name: name}
+}
+
+func (i *idServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+	if len(request.GetConnection().GetPath().GetPathSegments()) > int(request.GetConnection().GetPath().GetIndex()) &&
```
haiodo: Same here.

pkg/networkservicemesh/common/setid/client.go
edwarnicke: Done.

pkg/networkservicemesh/common/setid/server.go
edwarnicke: Done.

pkg/networkservicemesh/common/setid/client.go
```diff
@@ -0,0 +1,57 @@
...
+
+// NewClient - creates a new setId client.
+//             name - name of the client
+//             Iff the current pathSegment name != name && pathsegment.id != connection.Id, set a new uuid for
+//             connection id
+func NewClient(name string) networkservice.NetworkServiceClient {
+	return &idClient{name: name}
+}
+
+func (i *idClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*connection.Connection, error) {
```
fkautz: We take an `opts ...grpc.CallOption` but don't do anything with this. Is this intentional?

edwarnicke: Nope... should have passed it through to next.Client(ctx)... fixed.

PR #16, opened 2020-01-16
Add heal networkservice chain element

edwarnicke:


pkg/networkservicemesh/common/heal/client.go
```diff
@@ -0,0 +1,174 @@
...
+func NewClient(client connection.MonitorConnectionClient, onHeal *networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
+	rv := &healClient{
+		onHeal:            onHeal,
+		requestors:        make(map[string]func()),
+		closers:           make(map[string]func()),
+		reported:          make(map[string]*connection.Connection),
+		client:            client,
+		updateExecutor:    serialize.NewExecutor(),
+		eventReceiver:     nil,                     // This is intentionally nil
+		recvEventExecutor: serialize.NewExecutor(), // This is intentionally nil
```
fkautz: should this be a serialize.NewExecutor() or nil? I suspect the comment is incorrect due to usage further down.

pkg/networkservicemesh/common/heal/client.go
```diff
@@ -0,0 +1,174 @@
...
+			ctx = extend.WithValuesFromContext(timeCtx, ctx)
+			_, err := (*f.onHeal).Close(extend.WithValuesFromContext(timeCtx, ctx), req.GetConnection(), opts...)
+			trace.Log(ctx).Errorf("Attempt to close connection %s during heal resulted in error: %+v", req.GetConnection().GetId(), err)
+			cancelFunc()
+		}
+	})
+	return rv, nil
+}
+
+func (f *healClient) Close(ctx context.Context, conn *connection.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
```
fkautz: opts is never used, is this intentional?

pkg/networkservicemesh/common/heal/client.go
edwarnicke: Its intentionally nil.  populating eventReceiver involves making at *tcp* call... we want to minimize startup latency, so the tcp call happens in a updateExcutor.AsyncExec below...

PR #17, opened 2020-01-16
Added refresh networkservice chain element.

edwarnicke:


pkg/networkservicemesh/common/refresh/client.go
```diff
@@ -0,0 +1,106 @@
...
+	return next.Server(ctx).Close(ctx, conn)
+}
+
+func (t *refreshClient) createTimer(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*time.Timer, error) {
+	expireTime, err := ptypes.Timestamp(request.GetConnection().GetPath().GetPathSegments()[request.GetConnection().GetPath().GetIndex()].GetExpires())
+	if err != nil {
+		return nil, err
+	}
+	duration := time.Until(expireTime)
+	if err != nil {
```
fkautz: This always resolves to false. I think it should be safe to remove L94-96.

PR #18, opened 2020-01-16
Add timeout NetworkServiceServer chain element

edwarnicke:


pkg/networkservicemesh/common/timeout/server.go
```diff
@@ -0,0 +1,103 @@
...
+	return next.Server(ctx).Close(ctx, conn)
+}
+
+func (t *timeoutServer) createTimer(ctx context.Context, request *networkservice.NetworkServiceRequest) (*time.Timer, error) {
+	expireTime, err := ptypes.Timestamp(request.GetConnection().GetPath().GetPathSegments()[request.GetConnection().GetPath().GetIndex()].GetExpires())
+	if err != nil {
+		return nil, err
+	}
+	duration := time.Until(expireTime)
+	if err != nil {
```
fkautz: Same problem as on #17, should be safe to remove L94-96.

PR #28, opened 2020-01-17
Added roundrobin networkservice chain element.

edwarnicke:


pkg/networkservicemesh/common/roundrobin/round_robin_selector.go
```diff
@@ -0,0 +1,60 @@
...
+func newRoundRobinSelector() *roundRobinSelector {
+	return &roundRobinSelector{
+		roundRobin: make(map[string]int),
+	}
+}
+
+func (rr *roundRobinSelector) SelectEndpoint(requestConnection *connection.Connection, ns *registry.NetworkService, networkServiceEndpoints []*registry.NetworkServiceEndpoint) *registry.NetworkServiceEndpoint {
+	if rr == nil {
+		return nil
+	}
```
fkautz: L44-46 feel unnecessary. rr should never be nil.

pkg/networkservicemesh/common/roundrobin/round_robin_selector.go
```diff
@@ -0,0 +1,60 @@
...
+	roundRobin map[string]int
+}
+
+func newRoundRobinSelector() *roundRobinSelector {
+	return &roundRobinSelector{
+		roundRobin: make(map[string]int),
+	}
+}
+
+func (rr *roundRobinSelector) SelectEndpoint(requestConnection *connection.Connection, ns *registry.NetworkService, networkServiceEndpoints []*registry.NetworkServiceEndpoint) *registry.NetworkServiceEndpoint {
```
fkautz: requestConnection is never used. Is this intentional?

pkg/networkservicemesh/common/roundrobin/round_robin_selector.go
edwarnicke: Should never be, agreed.  But weird things happen, best not to panic.  Split the difference by combining with the if on length below it.

pkg/networkservicemesh/common/roundrobin/round_robin_selector.go
edwarnicke: Nope... removed the argument.

PR #38, opened 2020-01-20
Add Spire based security

edwarnicke:


pkg/tools/security/grpcoptions.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// WithSpireDial - use Spire identity for TLS
+func WithSpireDial(ctx context.Context) grpc.DialOption {
+	insecure := viper.GetBool(InsecureEnv)
+	if insecure {
+		return grpc.EmptyDialOption{}
+	}
+	// TODO - decide what to do about error here
+	provider, _ := NewSpireProvider(SpireAgentUnixAddr)
```
lobkovilya: In case of error we can return 
```go
return grpc.WithContextDialer(
    func(i context.Context, s string) (conn net.Conn, e error) {
        return nil, errors.New("some spire error here")
    })
```

For user it will behave like Dial returns error

edwarnicke: Fixed!

PR #40, opened 2020-01-20
Authz OPA server

lobkovilya:
Implement simple policy checker using OPA


pkg/networkservice/common/authorize/server.go
```diff
@@ -28,19 +31,55 @@ import (
...
 
-type authorizeServer struct{}
+type authorizeServer struct {
+	policy *rego.PreparedEvalQuery
+}
 
 // NewServer - returns a new authorization networkservicemesh.NetworkServiceServers
-func NewServer() networkservice.NetworkServiceServer {
-	return &authorizeServer{}
+func NewServer(p *rego.PreparedEvalQuery) networkservice.NetworkServiceServer {
```
edwarnicke: How would we update the policy after startup time?

pkg/networkservice/common/authorize/server.go
```diff
@@ -28,19 +31,55 @@ import (
...
+	if err := check(rs); err != nil {
+		return nil, err
+	}
+
 	return next.Server(ctx).Request(ctx, request)
 }
 
 func (a *authorizeServer) Close(ctx context.Context, conn *connection.Connection) (*empty.Empty, error) {
-	// TODO check authorization
+	rs, err := a.policy.Eval(ctx, rego.EvalInput(conn.GetPath().GetPathSegments()[conn.GetPath().GetIndex()].GetToken()))
```
edwarnicke: Are we only considering the current token?  Would it be possible to reason about more than that?

pkg/networkservice/common/authorize/server.go
edwarnicke: This is interesting: https://www.openpolicyagent.org/docs/latest/management/

pkg/networkservice/common/authorize/server.go
```diff
@@ -28,19 +31,55 @@ import (
...
+
 	return next.Server(ctx).Close(ctx, conn)
 }
+
+func check(rs rego.ResultSet) error {
+	for _, r := range rs {
+		for _, e := range r.Expressions {
+			t, ok := e.Value.(bool)
+			if !ok {
+				return status.Error(codes.Internal, "non boolean expression")
```
fkautz: can we make this error a little more descriptive? it isn't clear that opa returned a non boolean expression.

pkg/networkservice/common/authorize/server.go
fkautz: For a future patch, it would be useful to add support for requesting an eval from an OPA microservice rather than running a local eval. This is beneficial because the OPA microservice may make a decision based using access to a store or database which the NSE shouldn't have access to.

pkg/networkservice/common/authorize/server.go
```diff
@@ -28,19 +35,69 @@ import (
...
 
-type authorizeServer struct{}
+type authorizeServer struct {
+	p opa.Provider
+}
 
 // NewServer - returns a new authorization networkservicemesh.NetworkServiceServers
-func NewServer() networkservice.NetworkServiceServer {
-	return &authorizeServer{}
+func NewServer(p opa.Provider) networkservice.NetworkServiceServer {
```
edwarnicke: Would it be possible to keep the amount of new API to a minimum and just pass in a func() rego.PreparedEvalQuery ?

lobkovilya: Sure, I'll roll back my changes. I misinterpreted Fred's comment about separate service for OPA. I thought we want to obtain policies from some service, but now it is clear that Fred was talking only about evaluation.

pkg/networkservice/common/authorize/server.go
lobkovilya: We can pass here the whole `Request` or `Path`. I think we can do that after adding some real-world policy (not the test one)

pkg/networkservice/common/authorize/server.go
fkautz: Sorry about the confusion, I'll try to be more specific moving forward.

pkg/networkservice/common/authorize/server.go
fkautz: #51 has been created and added to the project board. Resolving this to unblock merging.

pkg/networkservice/common/authorize/server.go
fkautz: Added #52 to cover this.

PR #65, opened 2020-01-26
Local format and lint machinery

haiodo:
Local formatting and linting machinery

### DIFF
--- a/.mk/code_analysis.mk
+++ b/.mk/code_analysis.mk
@@ -1,0 +1,16 @@
+.PHONY: lint-fix
+
+lint-fix:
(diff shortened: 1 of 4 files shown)


.mk/code_analysis.mk
```diff
@@ -0,0 +1,16 @@
+.PHONY: lint-fix
+
```
denis-tingajkin: Is this new line not critical?

edwarnicke:
I'd like to avoid the reintroduction of make machinery to these repos.  They are smaller and simpler and the work you do for them doesn't require that level of complexity.  Its actually much easier to simply learn to use:

go test ./...
go build ./...
golangci-lint run

etc than to introduce a bunch of new scripts and Makefiles.

haiodo:
Ok, sure.

Issue #61, opened 2020-01-25
pkg/core/adapters can not be used with pkg/next


denis-tingajkin:
# Actual behavior
 
pkg/core/adapters  can adapt the server to client or vice versa, but if a server uses `pkg/next` then the adapters will not correctly adapt the server to client and it will panic on called Request/Close

 # Expected behavior

`pkg/core/adapters`  can be used with `pkg/next`.

# Steps to reproduce
1. Adapt server which uses next.Server(...) to client or  vice versa
2. Try to call client.Request() or client.Close()`

**Code example:**
server.go
```
type configServer struct{}

// NewServer - inserts a vppagent *configurator.Config into the GRPC call context.Context
func NewServer() networkservice.NetworkServiceServer {
	return &configServer{}
}

func (c *configServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
	return next.Server(ctx).Request(ctx, request)
}

func (c *configServer) Close(ctx context.Context, conn *connection.Connection) (*empty.Empty, error) {
	return next.Server(ctx).Close(ctx, conn)
}

```

client_test.go
```
func TestClientBasic(t *testing.T) {
	next.NewNetworkServiceClient(NewServerToClient(&configServer{})).Close(nil, nil, nil)
}

```
**Example from sdk-vppagent**:
https://github.com/networkservicemesh/sdk-vppagent/blob/master/pkg/networkservice/vppagent/client.go

## Stacktrace
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9c1110]

goroutine 6 [running]:
testing.tRunner.func1(0xc000140100)
	/usr/local/go/src/testing/testing.go:874 +0x60f
panic(0xa48b00, 0xeba540)
	/usr/local/go/src/runtime/panic.go:679 +0x1e0
github.com/networkservicemesh/sdk/pkg/networkservice/core/adapters.testServer.Close(0xb81160, 0xc0001344b0, 0x0, 0x0, 0x0, 0x0)
	/home/user/go/src/github.com/networkservicemesh/sdk/pkg/networkservice/core/adapters/client_test.go:21 +0x80
[...]

PR #70, opened 2020-01-27
Fix: pkg/core/adapters can not be used with pkg/next

denis-tingajkin:

### Motivation
https://github.com/networkservicemesh/sdk/issues/61


edwarnicke:
Try rebasing... now that things are using networkservicemesh/api not networkservicemesh/networkservicemesh/controlplane/api you will need it :)

denis-tingajkin:
@edwarnicke thanks, the problems are fixed.

pkg/networkservice/core/next/client.go
```diff
@@ -50,6 +52,9 @@ func NewWrappedNetworkServiceClient(wrapper ClientWrapper, clients ...networkser
 
 // NewNetworkServiceClient - chains together clients into a single networkservice.NetworkServiceClient
 func NewNetworkServiceClient(clients ...networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
+	if len(clients) == 0 {
```
edwarnicke: Nice catch :)

pkg/networkservice/core/next/context.go
```diff
@@ -47,6 +51,9 @@ func Server(ctx context.Context) networkservice.NetworkServiceServer {
 	if rv, ok := ctx.Value(nextServerKey).(networkservice.NetworkServiceServer); ok {
 		return rv
 	}
+	if rv, ok := ctx.Value(nextClientKey).(networkservice.NetworkServiceClient); ok {
```
edwarnicke: So I'm curious, why are we converting nextClient to nextServer here?  Perhaps I still need my morning coffee... but I've lost the thread :)

denis-tingajkin: It is needed for the case when we have for example a chain of servers and one adpated client
```
s := next.NewNetworkServiceServer(
something.NewServer(), 
adapters.NewClientToServer(something.NewClient()),
something.NewServer())
```
And we want that this adapted client can correctly call the next server.

Note: This client in `Request()` and `Close()`  uses  `next.Client(ctx)`. So if we will not adapt here we will break a chain.

edwarnicke: Excellent!  Thank you for the patient explanation :)

Issue #74, opened 2020-01-28
Create DNSContext NetworkServiceServer chain elements


edwarnicke:
Please create in pkg/networkservice/connectioncontext/dnscontext/ a NetworkServiceServer chain element appropriate for setting a DNSContext for an Endpoint.

PR #84, opened 2020-02-02
Create DNSContext NetworkServiceServer chain elements

denis-tingajkin:
## Motivation
https://github.com/networkservicemesh/sdk/issues/74


edwarnicke:
Could you rebase on master?  Turns out golangci-lint wasn't running on CI, now it is, and so the vast majority of file changes in this PR should disappear as a result :)

denis-tingajkin:
@edwarnicke 
> Could you rebase on master? 

Done.

> Turns out golangci-lint wasn't running on CI,  now it is

Nice catch! Thanks for the notice! 

Is golangci-lint correctly working with another repos?

pkg/networkservice/connectioncontext/dnscontext/utils.go
```diff
@@ -0,0 +1,49 @@
...
+package dnscontext
+
+import (
+	"github.com/kelseyhightower/envconfig"
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+	"github.com/sirupsen/logrus"
+)
+
+// ConfigFromEnv returns default getter for DNS configs. Builds config from env variables.
+func ConfigFromEnv() GetDNSConfigsFunc {
```
edwarnicke: Lets move the config env bits up to the cmd level.

pkg/networkservice/connectioncontext/dnscontext/server.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+// GetDNSConfigsFunc gets dns configs
+type GetDNSConfigsFunc func(conn *networkservice.Connection) []*networkservice.DNSConfig
```
edwarnicke: Do we anticipate DNSConfig changing dynamically through the life of the server?

denis-tingajkin: I'm not sure that it should be static.
I think it can be useful if the user's DNS configs could be provided from some dynamic thing.
Can you please share your thoughts?

pkg/networkservice/connectioncontext/dnscontext/utils.go
denis-tingajkin: Do you mean that it should be located in repo `networkservicemesh/cmd-...` ?

edwarnicke: Yes...

denis-tingajkin: Thanks, fixed.

pkg/networkservice/connectioncontext/dnscontext/server.go
edwarnicke: I’m not clear when it would be dynamic like this from a Connection... I could see a func() []*networkservice.DNSConfig ... I don’t quite get the example from utils/test ... what am I missing?

denis-tingajkin: I thought that it could be useful for users, but actually I can imagine only one scenario of usage `ConfigFromConnection` it is just a case when the DNS server located as a sidecar in NSE POD. 
It could be used only for testing. 
Finally, I've removed it.

PR #87, opened 2020-02-04
Monitor Server should be passed to clients

haiodo:
Monitor Server should be passed to clients to chain event recieved for connections.


pkg/networkservice/common/monitor/server.go
```diff
@@ -77,43 +85,80 @@ func (m *monitorServer) MonitorConnections(selector *networkservice.MonitorScope
 }
 
 func (m *monitorServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
+	// Pass metrics monitor, so it could be used later in chain.
+	ctx = WithServer(ctx, m)
+
+	connectionClone := proto.Clone(request.GetConnection())
 	conn, err := next.Server(ctx).Request(ctx, request)
 	if err == nil {
```
lobkovilya: Could you please rewrite code in the following way to avoid nesting?
```go
if err != nil {
    return nil, err
}

m.executor.AsyncExec(func() {
	m.connections[conn.GetId()] = conn

	if proto.Equal(connectionClone, conn) {
		return
	}

	if err := m.send(ctx, event); err != nil {
		logrus.Errorf("Error during sending event: %v", err)
	}
})

return conn, err

```

pkg/networkservice/common/monitor/filter.go
```diff
@@ -22,23 +22,37 @@ import (
...
 	return &monitorFilter{
 		selector: selector,
-		MonitorConnection_MonitorConnectionsServer: srv,
+		srv:      srv,
 	}
 }
 
-func (m *monitorFilter) Send(event *networkservice.ConnectionEvent) error {
+func (m *monitorFilter) Send(event *networkservice.ConnectionEvent, current map[string]*networkservice.Connection) error {
+	// Filter connections based on event passed and selector for this filter
```
lobkovilya: That feels like a comment to the whole method

pkg/networkservice/common/monitor/server.go
```diff
@@ -77,43 +85,80 @@ func (m *monitorServer) MonitorConnections(selector *networkservice.MonitorScope
...
-func (m *monitorServer) send(ctx context.Context, event *networkservice.ConnectionEvent) {
-	newMonitors := make([]networkservice.MonitorConnection_MonitorConnectionsServer, len(m.monitors))
-	for _, srv := range m.monitors {
-		select {
-		case <-srv.Context().Done():
-		default:
-			if err := srv.Send(event); err != nil {
-				trace.Log(ctx).Errorf("Error sending event: %+v: %+v", event, err)
+func (m *monitorServer) Send(event *networkservice.ConnectionEvent) error {
+	m.executor.AsyncExec(func() {
```
lobkovilya: This method already called from `AsyncExec` why we need to do it again?

pkg/networkservice/common/monitor/server.go
```diff
@@ -77,43 +85,80 @@ func (m *monitorServer) MonitorConnections(selector *networkservice.MonitorScope
...
+
+		if err := m.send(context.Background(), event); err != nil {
+			logrus.Errorf("Error sending event %v", err)
+		}
+	})
+	return nil
+}
+
+func (m *monitorServer) send(ctx context.Context, event *networkservice.ConnectionEvent) (err error) {
+	m.executor.AsyncExec(func() {
```
lobkovilya: and again?

pkg/networkservice/common/monitor/server.go
```diff
@@ -77,43 +85,80 @@ func (m *monitorServer) MonitorConnections(selector *networkservice.MonitorScope
...
 
-func (m *monitorServer) send(ctx context.Context, event *networkservice.ConnectionEvent) {
-	newMonitors := make([]networkservice.MonitorConnection_MonitorConnectionsServer, len(m.monitors))
-	for _, srv := range m.monitors {
-		select {
-		case <-srv.Context().Done():
-		default:
-			if err := srv.Send(event); err != nil {
-				trace.Log(ctx).Errorf("Error sending event: %+v: %+v", event, err)
+func (m *monitorServer) Send(event *networkservice.ConnectionEvent) error {
```
lobkovilya: Why this `Send` is public?

pkg/networkservice/common/monitor/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/metadata"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/trace"
+)
+
+type testMonitorServer struct {
```
lobkovilya: Let's place all auxiliary test stuff to `pkg/test/...` dir to have `*_test.go` files more neat

pkg/networkservice/common/monitor/server_test.go
```diff
@@ -0,0 +1,268 @@
+package monitor
```
lobkovilya: Please move tests to `monitor_test` package

pkg/networkservice/common/monitor/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+	return &empty.Empty{}, nil
+}
+
+func TestMonitorSendToRightClient(t *testing.T) {
+	myServer := &myServer{}
+
+	ctx := context.Background()
+	ms := NewServer(&myServer.MonitorConnectionServer)
+	mons, ok := ms.(*monitorServer)
+	require.Equal(t, true, ok)
```
lobkovilya: `require.True(t, ok)`

pkg/networkservice/common/monitor/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+			// Context is done, we need to exit
+			logrus.Errorf("Failed to wait for events count %v current value is: %v", count, len(t.events))
+			return
+		case <-t.eventChannel:
+		case <-time.After(1 * time.Second):
+		}
+	}
+}
+
+type myServer struct {
```
lobkovilya: Can we avoid using `my` in the naming?

pkg/networkservice/common/monitor/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+	}()
+}
+
+func (t *testMonitorServer) WaitEvents(ctx context.Context, count int) {
+	for {
+		if len(t.events) == count {
+			logrus.Infof("Waiting for events %v, but has %v", count, len(t.events))
+			break
+		}
+		// Wait 10ms for listeners to activate
```
lobkovilya: Comment is probably outdated, I don't see `10ms` anywhere in the code

pkg/networkservice/common/monitor/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+			&networkservice.MonitorScopeSelector{
+				PathSegments: []*networkservice.PathSegment{{Name: segmentName}},
+			}, t)
+	}()
+}
+
+func (t *testMonitorServer) WaitEvents(ctx context.Context, count int) {
+	for {
+		if len(t.events) == count {
+			logrus.Infof("Waiting for events %v, but has %v", count, len(t.events))
```
lobkovilya: Info line is strange, because `count` and `len(t.events)` are equal if it prints

pkg/networkservice/common/monitor/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+	updateEnv := newUpdateTailServer(updateCounter)
+
+	srv := next.NewWrappedNetworkServiceServer(trace.NewNetworkServiceServer, ms, updateEnv)
+
+	localMonitor := newTestMonitorServer()
+	remoteMonitor := newTestMonitorServer()
+	localMonitor.BeginMonitoring(myServer, "local-nsm")
+	remoteMonitor.BeginMonitoring(myServer, "remote-nsm")
+	// We need to be sure we have 2 clients waiting for events, we could check to have initial transfers for this.
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), time.Second*6000)
```
lobkovilya: Do we really need such big timeout?

pkg/networkservice/common/monitor/server.go
haiodo: Because it implements MonitorConnection_MonitorConnectionsServer

pkg/networkservice/common/monitor/server.go
haiodo: We could not be sure.

pkg/networkservice/common/monitor/server.go
haiodo: Thanks.

pkg/networkservice/common/monitor/server_test.go
haiodo: I can't because test use internals to check.

pkg/networkservice/common/monitor/server.go
```diff
@@ -81,39 +85,47 @@ func (m *monitorServer) Request(ctx context.Context, request *networkservice.Net
...
 		m.executor.AsyncExec(func() {
 			m.connections[conn.GetId()] = conn
+			// Send update only if connection is updated or has metrics pending updates
 			event := &networkservice.ConnectionEvent{
 				Type:        networkservice.ConnectionEventType_UPDATE,
 				Connections: map[string]*networkservice.Connection{conn.GetId(): conn},
 			}
-			m.send(ctx, event)
+			if err = m.send(ctx, event); err != nil {
+				logrus.Errorf("Error during sending event: %v", err)
```
edwarnicke: Would it make sense to use trace.Log(ctx).Errorf here instead?

pkg/networkservice/common/monitor/server.go
```diff
@@ -81,39 +85,47 @@ func (m *monitorServer) Request(ctx context.Context, request *networkservice.Net
...
+	// Pass metrics monitor, so it could be used later in chain.
 	m.executor.AsyncExec(func() {
 		delete(m.connections, conn.GetId())
 		event := &networkservice.ConnectionEvent{
 			Type:        networkservice.ConnectionEventType_DELETE,
 			Connections: map[string]*networkservice.Connection{conn.GetId(): conn},
 		}
-		m.send(ctx, event)
+		if err := m.send(ctx, event); err != nil {
+			logrus.Errorf("Error during sending event: %v", err)
```
edwarnicke: Would it make sense to use trace.Log(ctx).Errorf here instead?

pkg/test/monitor/test_monitor.go
```diff
@@ -14,15 +14,16 @@
...
+// Package monitor provides a TestMonitorClient test class to perform client monitoring based on MonitorConnection call and tests for monitor sdk chain components
 package monitor
 
 import (
 	"context"
-	"github.com/networkservicemesh/sdk/pkg/tools/serialize"
 	"runtime"
 	"time"
 
+	"github.com/networkservicemesh/sdk/pkg/tools/serialize"
```
edwarnicke: Do we know why imports are moving around here?  More interestingly... do we know why linting isn't catching it?

pkg/test/monitor/server_test.go
```diff
@@ -0,0 +1,194 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
```
edwarnicke: Why not just put this in the pkg/networkservice/common/monitor/ directory using package monitor_test ?

pkg/networkservice/common/monitor/server.go
```diff
@@ -81,39 +85,47 @@ func (m *monitorServer) Request(ctx context.Context, request *networkservice.Net
...
 		m.executor.AsyncExec(func() {
 			m.connections[conn.GetId()] = conn
+			// Send update only if connection is updated or has metrics pending updates
 			event := &networkservice.ConnectionEvent{
 				Type:        networkservice.ConnectionEventType_UPDATE,
 				Connections: map[string]*networkservice.Connection{conn.GetId(): conn},
 			}
-			m.send(ctx, event)
+			if sendErr := m.send(ctx, event); sendErr != nil {
+				logrus.Errorf("Error during sending event: %v", sendErr)
```
edwarnicke: Would it make more sense to use trace.Log(ctx).Errorf here?

pkg/test/monitor/test_monitor.go
haiodo: Nope, no idea.

pkg/test/monitor/server_test.go
haiodo: Moved to tests package inside monitor

edwarnicke:
Ah… I was unclear

Let me find an example:
https://github.com/networkservicemesh/sdk/blob/master/pkg/tools/serialize/serialize_test.go#L21

In Golang, you can put your foo_test.go file in the same directory with your code, and simply give it package foo_test, that way you are testing your code from *outside* its package, but you don't have to have a separate tests/ subdir

pkg/networkservice/common/monitor/test_monitor.go
```diff
@@ -0,0 +1,104 @@
...
+	eventChannel chan *networkservice.ConnectionEvent
+	ctx          context.Context
+	Cancel       context.CancelFunc
+	grpc.ServerStream
+	executor  serialize.Executor
+	finalized chan struct{}
+}
+
+// NewTestMonitorClient - construct a new client.
+func NewTestMonitorClient() *TestMonitorClient {
```
edwarnicke: Do you intend TestMonitorClient to be a useful tool for testing things outside of the monitor package?

haiodo: @lobkovilya wanted to use it for some of healing tests AS I remember. I could make it private for now. And change back if it will be required.

edwarnicke: Lets do that :)

pkg/networkservice/common/monitor/server.go
```diff
@@ -81,39 +83,47 @@ func (m *monitorServer) Request(ctx context.Context, request *networkservice.Net
...
+			if sendErr := m.send(ctx, event); sendErr != nil {
+				trace.Log(ctx).Errorf("Error during sending event: %v", sendErr)
+			}
 		})
 	}
 	return conn, err
 }
 
 func (m *monitorServer) Close(ctx context.Context, conn *networkservice.Connection) (*empty.Empty, error) {
+	// Pass metrics monitor, so it could be used later in chain.
```
lobkovilya: What that comment related to?

pkg/networkservice/common/monitor/test_monitor.go
lobkovilya: No, I suggested to put auxiliary classes (fixtures, mock, dummy and etc) to `pkg/test` to have ability use them from another unit tests. Package name should reflect the fact that this is testing util - `monitortestutil` or smth like that. 
I think it would be nice to keep `*_test.go` files without any auxiliary stuff, just unit tests.

pkg/networkservice/common/monitor/test_monitor.go
```diff
@@ -0,0 +1,104 @@
...
+	for {
+		var curLen = 0
+		t.executor.SyncExec(func() {
+			curLen = len(t.Events)
+		})
+		if curLen == count {
+			logrus.Infof("Waiting for Events %v Complete", count)
+			break
+		}
+		// Wait 10ms for listeners to activate
```
lobkovilya: I don't see `10ms` anywhere in the code

pkg/networkservice/common/monitor/test_monitor.go
edwarnicke: I'm fine with doing that *when* we come across a case for which we want to reuse... but lets work with actual rather than imagined cases of that sort.  Keep things as simple as possible :)

pkg/networkservice/common/monitor/server.go
```diff
@@ -81,39 +83,47 @@ func (m *monitorServer) Request(ctx context.Context, request *networkservice.Net
 	if err == nil {
 		m.executor.AsyncExec(func() {
 			m.connections[conn.GetId()] = conn
+			// Send update only if connection is updated or has metrics pending updates
```
edwarnicke: Is this comment still true?

pkg/networkservice/common/monitor/test_monitor.go
lobkovilya: Yes but in the future someone might doesn't notice that `TestMonitorClient` implementation and write own one. So I think we should keep all auxiliary testing stuff in one place.

pkg/networkservice/common/monitor/test_monitor.go
```diff
@@ -0,0 +1,104 @@
...
+		select {
+		case <-ctx.Done():
+			// Context is done, we need to exit
+			logrus.Errorf("Failed to wait for Events count %v current value is: %v", count, curLen)
+			return
+		case <-t.eventChannel:
+		case <-time.After(1 * time.Second):
```
edwarnicke: Why are we pausing for 1 second here?  Which not just wait on a event or ctx.Done()?

pkg/networkservice/common/monitor/test_monitor.go
```diff
@@ -0,0 +1,104 @@
...
+	return nil
+}
+
+// Context - current context to perform checks.
+func (t *TestMonitorClient) Context() context.Context {
+	return t.ctx
+}
+
+// BeginMonitoring - start monitoring inside go routine, use Cancel() to perform stop.
+func (t *TestMonitorClient) BeginMonitoring(server networkservice.MonitorConnectionServer, segmentName string) {
```
edwarnicke: Why not just roll this into the constructor?

pkg/networkservice/common/monitor/test_monitor.go
haiodo: Thanks.

PR #90, opened 2020-02-10
Fix nil pointer exception on call ConnectServer.Request

denis-tingajkin:


pkg/networkservice/common/connect/client_map.go
```diff
@@ -36,7 +36,11 @@ func (c *clientMap) Delete(key string) {
 
 func (c *clientMap) Load(key string) (networkservice.NetworkServiceClient, bool) {
 	value, ok := c.internalMap.Load(key)
-	return value.(networkservice.NetworkServiceClient), ok
```
denis-tingajkin: https://github.com/golang/go/issues/29763

edwarnicke: Ah... in the event we get back a nil, we can't type coerce that... good catch.

PR #91, opened 2020-02-10
Add client dial options to `connectServer`

denis-tingajkin:

## Motivation

It could help to write tests that depend on `connect` package. 
For example to testing `directmemif` package.

### DIFF
--- a/pkg/networkservice/common/connect/server.go
+++ b/pkg/networkservice/common/connect/server.go
@@ -29,6 +29,7 @@ import (
 
 type connectServer struct {
 	ctx                     context.Context
+	dialOptionsSupplier     func() []grpc.DialOption
 	clientFactory           func(ctx context.Context, conn grpc.ClientConnInterface) networkservice.NetworkServiceClient
 	uRLToClientMap          clientMap // key == url as string
 	connectionIDToClientMap clientMap // key == connection.GetId()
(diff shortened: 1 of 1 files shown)


pkg/networkservice/common/connect/server.go
```diff
@@ -44,11 +45,17 @@ type connectServer struct {
 //             connect presumes depends on some previous chain element having set clienturl.WithClientURL so it can know
 //             which client to address.
 func NewServer(clientFactory func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient) networkservice.NetworkServiceServer {
+	return NewServerWithDialOptions(clientFactory, func() []grpc.DialOption { return nil })
+}
+
+// NewServerWithDialOptions - returns a new connect Server with specific client dial options
+func NewServerWithDialOptions(clientFactory func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient, dialOptionsSupplier func() []grpc.DialOption) networkservice.NetworkServiceServer {
```
edwarnicke: Why not just change NewServer to take a dialOptionsSupplier or better yet a ...grpc.DialOption ?

edwarnicke:
Overall this is heading very much in the right direction, just a couple of simple comments on keeping things simple by having NewServer take variatic list of dialOptions at the end.

pkg/networkservice/common/connect/server.go
denis-tingajkin: fixed, thanks

PR #92, opened 2020-02-11
Metrics monitor

haiodo:
Implementation of Metrics chain element, it exposes 

```go
// MetricsMonitor - extension to monitor service to pass metrics into MonitorServer recipients.
// monitor will send events with configured with connection interval, often updates will wait for
// next interval approaching before will be sended.
type MetricsMonitor interface {
	// HandleMetrics - send update to current segment metrics
	HandleMetrics(metrics map[string]string)
}
```

Interface to all next chain elements, and allow to update current path segment metrics and send updates.

PR is based on: https://github.com/networkservicemesh/sdk/pull/87

### DIFF
--- a/pkg/networkservice/common/heal/client.go
+++ b/pkg/networkservice/common/heal/client.go
@@ -22,6 +22,10 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/common/monitor"
+
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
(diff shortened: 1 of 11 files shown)


pkg/networkservice/common/heal/client.go
```diff
@@ -22,6 +22,10 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/sirupsen/logrus"
```
edwarnicke: Why are we mixing metrics into the heal Client?

edwarnicke:
Whatever we do for handling Metrics, it should live in a well defined localized chain element.  It should not be introducing new interfaces (HandleMetrics), it should not require changes to other chain elements unrelated to metrics (heal, etc).  If we drive the metrics from the receipt of a Request, rather than a new timer (interval) it becomes nicely localized, can reuse the existing monitor without additional complication etc.  Let's keep things simple :)

haiodo:
Closing, since not required in concept of Request() only chain element.

Issue #43, opened 2020-01-20
Create NetworkServiceClient chain element to add podName/nodeName/clusterName labels to request


edwarnicke:
Create NetworkServiceClient chain element to add podName/nodeName/clusterName labels to request.

Put it in sdk/pkg/networkservice/common/

denis-tingajkin:
@edwarnicke Can we close this issue?

Issue #44, opened 2020-01-20
Create NetworkServiceRegistryClient to add podName/nodeName/clusterName labels to registrations


edwarnicke:
Create NetworkServiceRegistryClient to add podName/nodeName/clusterName labels to registrations.
Put it in sdk/pkg/registry/common/

PR #93, opened 2020-02-12
clientInfo chain element

lvfxx:
Implementation of #43 
Created clientInfo NetworkServiceClient chain element that takes podName/nodeName/clusterName from environment variables and adds corresponding labels to request

### DIFF
--- a/pkg/networkservice/common/clientinfo/client.go
+++ b/pkg/networkservice/common/clientinfo/client.go
@@ -1,27 +1,27 @@
-package addNames
+package clientinfo
 
 import (
 	"context"
+	"os"
+
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
-
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
-
-	"os"
 )
 
-type addNamesClient struct{}
+var names = map[string]string{"NODE_NAME": "NodeNameKey", "POD_NAME": "PodNameKey", "CLUSTER_NAME": "ClusterNameKey"}
+
+type clientInfo struct{}
 
 // NewClient - creates a new networkservice.NetworkServiceClient chain element that adds pod, node and cluster names to request from corresponding environment variables
 func NewClient() networkservice.NetworkServiceClient {
-	return &addNamesClient{}
+	return &clientInfo{}
 }
 
-func (a *addNamesClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
-	names := map[string]string{"NODE_NAME": "NodeNameKey", "POD_NAME": "PodNameKey", "CLUSTER_NAME": "ClusterNameKey"}
+func (a *clientInfo) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
 	conn := request.GetRequestConnection()
 	if conn.Labels == nil {
 		conn.Labels = make(map[string]string)
@@ -41,6 +41,6 @@ func (a *addNamesClient) Request(ctx context.Context, request *networkservice.Ne
 	return next.Client(ctx).Request(ctx, request, opts...)
 }
 
-func (a *addNamesClient) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
+func (a *clientInfo) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
 	return next.Client(ctx).Close(ctx, conn, opts...)
 }
--- a/pkg/networkservice/common/clientinfo/client.go
+++ b/pkg/networkservice/common/clientinfo/client.go
@@ -14,6 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package clientinfo provides a chain element that adds pod, node and cluster names to request
 package clientinfo
 
 import (
@@ -25,10 +26,18 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 )
 
-const nodeEnv, podEnv, clusterEnv, nodeLabel, podLabel, clusterLabel = "NODE_NAME", "POD_NAME", "CLUSTER_NAME", "NodeNameKey", "PodNameKey", "ClusterNameKey"
+const (
+	nodeEnv      = "NODE_NAME"
+	podEnv       = "POD_NAME"
+	clusterEnv   = "CLUSTER_NAME"
+	nodeLabel    = "NodeNameKey"
+	podLabel     = "PodNameKey"
+	clusterLabel = "ClusterNameKey"
+)
 
 type clientInfo struct{}
 
--- a/pkg/networkservice/common/clientinfo/client.go
+++ b/pkg/networkservice/common/clientinfo/client.go
@@ -1,3 +1,19 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package clientinfo
 
 import (
--- a/pkg/networkservice/common/clientinfo/client.go
+++ b/pkg/networkservice/common/clientinfo/client.go
@@ -31,23 +31,28 @@ import (
 )
 
 const (
-	nodeEnv      = "NODE_NAME"
-	podEnv       = "POD_NAME"
-	clusterEnv   = "CLUSTER_NAME"
-	nodeLabel    = "NodeNameKey"
-	podLabel     = "PodNameKey"
-	clusterLabel = "ClusterNameKey"
+	nodeNameEnv      = "NODE_NAME"
+	podNameEnv       = "POD_NAME"
+	clusterNameEnv   = "CLUSTER_NAME"
+	nodeNameLabel    = "NodeNameKey"
+	podNameLabel     = "PodNameKey"
+	clusterNameLabel = "ClusterNameKey"
 )
 
 type clientInfo struct{}
 
-// NewClient - creates a new networkservice.NetworkServiceClient chain element that adds pod, node and cluster names to request from corresponding environment variables
+// NewClient - creates a new networkservice.NetworkServiceClient chain element that adds pod, node and cluster names
+// to request from corresponding environment variables
 func NewClient() networkservice.NetworkServiceClient {
 	return &clientInfo{}
 }
 
 func (a *clientInfo) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
-	var names = map[string]string{nodeEnv: nodeLabel, podEnv: podLabel, clusterEnv: clusterLabel}
+	var names = map[string]string{
+		nodeNameEnv:    nodeNameLabel,
+		podNameEnv:     podNameLabel,
+		clusterNameEnv: clusterNameLabel,
+	}
 
 	conn := request.GetRequestConnection()
 	if conn.Labels == nil {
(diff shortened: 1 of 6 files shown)


pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+var names = map[string]string{"NODE_NAME": "NodeNameKey", "POD_NAME": "PodNameKey", "CLUSTER_NAME": "ClusterNameKey"}
```
lobkovilya: Please use constants instead of global map

pkg/networkservice/common/clientinfo/client_test.go
```diff
@@ -0,0 +1,115 @@
+package clientinfo
```
lobkovilya: Please move tests to `clientinfo_test` package

pkg/networkservice/common/clientinfo/client_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"context"
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+type clientInfoTestData struct {
```
denis-tingajkin: FYI... you could use anonymous structures for test samples

pkg/networkservice/common/clientinfo/client_test.go
```diff
@@ -0,0 +1,115 @@
...
+	name    string
+	envs    map[string]string
+	request *networkservice.NetworkServiceRequest
+	want    *networkservice.Connection
+}
+
+var tests = []clientInfoTestData{
```
denis-tingajkin: Please let's not use global test samples

pkg/networkservice/common/clientinfo/client_test.go
```diff
@@ -0,0 +1,115 @@
...
+		for name, value := range testData.envs {
+			err := os.Setenv(name, value)
+			if err != nil {
+				t.Errorf("%s: clientInfo.Request() unable to set up environment variable: %v", testData.name, err)
+			}
+		}
+
+		got, _ := server.Request(context.Background(), testData.request)
+		if !reflect.DeepEqual(got, testData.want) {
+			t.Errorf("%s: clientInfo.Request() = %v, want %v", testData.name, got, testData.want)
```
denis-tingajkin: Let's use testify for assertation

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,46 @@
...
+			conn.Labels[labelName] = value
+		} else {
+			logrus.Warningf("Environment variable %s is not set. Skipping.", envName)
+		}
+	}
+	return next.Client(ctx).Request(ctx, request, opts...)
+}
+
+func (a *clientInfo) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
+	return next.Client(ctx).Close(ctx, conn, opts...)
```
denis-tingajkin: Please add labels in case of close

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,46 @@
...
+	conn := request.GetRequestConnection()
+	if conn.Labels == nil {
+		conn.Labels = make(map[string]string)
+	}
+	for envName, labelName := range names {
+		value, exists := os.LookupEnv(envName)
+		if exists {
+			oldValue, isPresent := conn.Labels[labelName]
+			if isPresent {
+				logrus.Warningf("The label %s was already assigned to %s. Overwriting.", labelName, oldValue)
```
denis-tingajkin: Do we really need to replace values here?

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,46 @@
...
+	return &clientInfo{}
+}
+
+func (a *clientInfo) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+	conn := request.GetRequestConnection()
+	if conn.Labels == nil {
+		conn.Labels = make(map[string]string)
+	}
+	for envName, labelName := range names {
+		value, exists := os.LookupEnv(envName)
```
lobkovilya: I saw examples of using `viper.GetBool` in this repo. Do we want to have it as default way to read envs? cc @denis-tingajkin @edwarnicke

denis-tingajkin: Sounds good for me

pkg/networkservice/common/clientinfo/client_test.go
lvfxx: Ok, i will use them

pkg/networkservice/common/clientinfo/client_test.go
lvfxx: Сan you explain why it is bad doing so? If i move them to function's body, linter will complain about method body too long.

pkg/networkservice/common/clientinfo/client_test.go
lvfxx: Ok, i will do it

pkg/networkservice/common/clientinfo/client_test.go
lvfxx: Ok, i will use it

pkg/networkservice/common/clientinfo/client.go
lvfxx: Ok, i'll fix it

pkg/networkservice/common/clientinfo/client_test.go
denis-tingajkin: Can you attach the linter error?

pkg/networkservice/common/clientinfo/client.go
lvfxx: Why do you think we should do it? Connection passed to Close() is expected to contain labels as they are already were injected by preceding Request()
@edwarnicke what do you think?

denis-tingajkin: @lvfxx  because it can be used by next chain elements.

pkg/networkservice/common/clientinfo/client_test.go
lvfxx: No, i can't. 
I used [this]( https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/roundrobin/round_robin_selector_test.go) code as example.

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,73 @@
...
+func (a *clientInfo) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+	var names = map[string]string{nodeEnv: nodeLabel, podEnv: podLabel, clusterEnv: clusterLabel}
+
+	conn := request.GetRequestConnection()
+	if conn.Labels == nil {
+		conn.Labels = make(map[string]string)
+	}
+	for envName, labelName := range names {
+		value, exists := os.LookupEnv(envName)
+		if exists {
```
lobkovilya: Let's try to eliminate nested constructions here

lvfxx: Ok, fixed it.

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,73 @@
...
+
+type clientInfo struct{}
+
+// NewClient - creates a new networkservice.NetworkServiceClient chain element that adds pod, node and cluster names to request from corresponding environment variables
+func NewClient() networkservice.NetworkServiceClient {
+	return &clientInfo{}
+}
+
+func (a *clientInfo) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+	var names = map[string]string{nodeEnv: nodeLabel, podEnv: podLabel, clusterEnv: clusterLabel}
```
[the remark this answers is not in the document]
lvfxx: Ok, fixed it.

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,73 @@
...
+	podEnv       = "POD_NAME"
+	clusterEnv   = "CLUSTER_NAME"
+	nodeLabel    = "NodeNameKey"
+	podLabel     = "PodNameKey"
+	clusterLabel = "ClusterNameKey"
+)
+
+type clientInfo struct{}
+
+// NewClient - creates a new networkservice.NetworkServiceClient chain element that adds pod, node and cluster names to request from corresponding environment variables
```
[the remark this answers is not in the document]
lvfxx: Ok, fixed it.

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+const (
+	nodeEnv      = "NODE_NAME"
```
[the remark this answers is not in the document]
lvfxx: Ok, fixed it.

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+const (
+	nodeNameEnv      = "NODE_NAME"
+	podNameEnv       = "POD_NAME"
+	clusterNameEnv   = "CLUSTER_NAME"
```
jellonek: Can you extract this consts to some more generic package/file, as they will be used also on https://github.com/networkservicemesh/sdk/issues/44

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,78 @@
...
+		nodeNameEnv:    nodeNameLabel,
+		podNameEnv:     podNameLabel,
+		clusterNameEnv: clusterNameLabel,
+	}
+
+	conn := request.GetRequestConnection()
+	if conn.Labels == nil {
+		conn.Labels = make(map[string]string)
+	}
+	for envName, labelName := range names {
```
jellonek: Can you extract whole this labels manipulation as a function to a separate module, so it could be used also in implementation of https://github.com/networkservicemesh/sdk/issues/44 ?
I see that as a func which is getting a map of labels as an arg and returns nothing.
This will allow also to have common test for the same functionality used in two (there and in registry client) places.

pkg/networkservice/common/clientinfo/client.go
edwarnicke: I'm generally highly pro-viper ... typically you want to use viper to read various things at top level (in your main) and use them to create args further down.  In this case, os.LookupEnv is simple and brings no new dependencies, so probably the right choice :)

pkg/networkservice/common/clientinfo/client.go
edwarnicke: @denis-tingajkin Could you say more?

jellonek: As they were already added during initialization - if they were removed/changed later, imo they should not be recreated again during `Close`.
@denis-tingajkin can you show example code which would require readding them there?

denis-tingajkin: @edwarnicke Is it can be useful for metrics?

pkg/networkservice/common/clientinfo/client.go
lvfxx: Ok, did it

pkg/networkservice/common/clientinfo/client.go
lvfxx: Ok, did it

pkg/networkservice/common/clientinfo/client.go
lvfxx: Actually i don't know. Maybe @edwarnicke knows which option is more suitable?

pkg/networkservice/common/clientinfo/clientinfo_test/client_test.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+var testData = []struct {
+	name    string
+	envs    map[string]string
+	request *networkservice.NetworkServiceRequest
+	want    *networkservice.Connection
+}{
+	{
+		"the-labels-map-is-not-present",
```
lobkovilya: Please use spaces instead of dash

lvfxx: Ok, fixed it

pkg/tools/clientinfoutils/clientinfoutils.go
```diff
@@ -0,0 +1,55 @@
...
+func AddClientInfo(labels map[string]string) {
+	names := map[string]string{
+		nodeNameEnv:    nodeNameLabel,
+		podNameEnv:     podNameLabel,
+		clusterNameEnv: clusterNameLabel,
+	}
+	for envName, labelName := range names {
+		value, exists := os.LookupEnv(envName)
+		if !exists {
+			logrus.Warningf("Environment variable %s is not set. Skipping.", envName)
```
jellonek: I think that this should be an error instead of warning, but it's a very minor nit.
Btw. for such things as variable names and their values I would prefer %q as a format type, but it is also very minor.

pkg/tools/clientinfoutils/clientinfoutils.go
```diff
@@ -0,0 +1,55 @@
...
+			continue
+		}
+		oldValue, isPresent := labels[labelName]
+		if isPresent {
+			logrus.Warningf("The label %s was already assigned to %s. Overwriting.", labelName, oldValue)
```
jellonek: Ditto for note about formatting string.

pkg/networkservice/common/clientinfo/client.go
edwarnicke: I'm a bit of mixed mind.  I tend to like @denis-tingajkin 's tendency to code conservatively here.  And I suspect adding the labels in the Close() would be harmless.  I have trouble though thinking of a case where we should have to add them.  At this time I'm ok either way.

edwarnicke:
LGTM, @denis-tingajkin anything you want cleaned up?

pkg/networkservice/common/clientinfo/client.go
denis-tingajkin: @edwarnicke I agree we can add these labels on close when we face a real case

denis-tingajkin:
@edwarnicke In general looks good to me

PR #95, opened 2020-02-12
Add NetworkServiceServer chain element for handling Mechanism Selection

edwarnicke:


pkg/networkservice/common/mechanisms/server.go
```diff
@@ -0,0 +1,81 @@
...
+func (m mechanismsServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
+	if request.GetConnection().GetMechanism() == nil {
+		for _, mechanism := range request.GetMechanismPreferences() {
+			srv, ok := m.mechanisms[mechanism.GetType()]
+			if ok {
+				req := request.Clone()
+				req.GetConnection().Mechanism = mechanism
+				conn, err := srv.Request(ctx, req)
+				if err == nil {
+					request = request.Clone()
```
fkautz: What's the reason for the second clone here?

pkg/networkservice/common/mechanisms/server.go
```diff
@@ -0,0 +1,81 @@
...
+					request.Connection = conn
+					return next.Server(ctx).Request(ctx, request)
```
fkautz: Did you mean to return here? Is L61 only meant to run if no preferences are found?

pkg/networkservice/common/mechanisms/server.go
edwarnicke: You are correct.

pkg/networkservice/common/mechanisms/server.go
edwarnicke: Yes, but your question led me to rewriting in a way that is hopefully clearer.

PR #97, opened 2020-02-13
clientInfo NetworkServiceRegistryClient

lvfxx:
Implementation of #44 
Created clientInfoRegistry NetworkServiceRegistryClient that takes podName/nodeName/clusterName from environment variables and adds corresponding labels to registration
Also added tailRegistryClient to be able to run tests

### DIFF
--- a/pkg/registry/common/clientinfo/client.go
+++ b/pkg/registry/common/clientinfo/client.go
@@ -1,0 +1,54 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package clientinfo provides a registry client that adds pod, node and cluster names to registration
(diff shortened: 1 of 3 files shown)


denis-tingajkin:
I think we need to provide bug fixing as separate PR

pkg/networkservice/common/clientinfo/client.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
```
edwarnicke: Looks like you may need to rebase now that the other PR has been merged... I think these files are from the last PR :)

jellonek:
PTAL on https://github.com/networkservicemesh/sdk/pull/106 - adding tail there should be not needed any more.

edwarnicke:
@lvfxx I think this needs to be rebased, but otherwise looks good to me :)

edwarnicke:
@lvfxx Are you going to follow up on @haiodo 's comments?

pkg/registry/common/clientinfo/client.go
```diff
@@ -0,0 +1,54 @@
...
+	nse := in.NetworkServiceEndpoint
+	if nse.Labels == nil {
+		nse.Labels = make(map[string]string)
+	}
+	clientinfo.AddClientInfo(ctx, nse.Labels)
+	return next.RegistryClient(ctx).RegisterNSE(ctx, in, opts...)
+}
+
+func (a *clientInfoRegistry) BulkRegisterNSE(ctx context.Context, opts ...grpc.CallOption) (registry.NetworkServiceRegistry_BulkRegisterNSEClient, error) {
+	return next.RegistryClient(ctx).BulkRegisterNSE(ctx, opts...)
```
haiodo: Do we need here also ?
clientinfo.AddClientInfo(ctx, nse.Labels)?

haiodo:
I think it is outdated and on internal discussion with @denis-tingajkin  we do not think this changes are required.

PR #98, opened 2020-02-13
Rename clientinfoutils to clientinfo

edwarnicke:
Also switch to using trace.Log(ctx) in clientinfo

### DIFF
--- a/pkg/tools/clientinfo/clientinfo.go
+++ b/pkg/tools/clientinfo/clientinfo.go
@@ -14,13 +14,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package clientinfoutils provides a set of utilities for adding client info to labels map
-package clientinfoutils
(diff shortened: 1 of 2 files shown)


edwarnicke:
@lvfxx @denis-tingajkin @jellonek @lobkovilya Thoughts?

pkg/tools/clientinfo/clientinfo.go
```diff
@@ -43,12 +44,12 @@ func AddClientInfo(labels map[string]string) {
 	for envName, labelName := range names {
 		value, exists := os.LookupEnv(envName)
 		if !exists {
-			logrus.Warningf("Environment variable %s is not set. Skipping.", envName)
+			trace.Log(ctx).Warningf("Environment variable %s is not set. Skipping.", envName)
```
denis-tingajkin: nice catch

PR #99, opened 2020-02-13
Move client_test.go up to the same directory

edwarnicke:
Golang idiomatically puts unittests in the same directory as the module being tested.
To suppor this, a directory can contain both package ${packet} and package ${package}_test
if and only if the file name for ${package}_test ends in *_test.go.


edwarnicke:
@lvfxx @denis-tingajkin @lobkovilya @jellonek Thoughts?

denis-tingajkin:
'Linter' for this case can look like this:
```
package style_test

func TestPkgNames(t *testing.T){
	p, err := filepath.Abs("../../..") // way to root
	require.Nil(t, err)
	var result []string
	filepath.Walk(p, func(path string, info os.FileInfo, err error) error {
		if strings.HasSuffix(info.Name(), "_test.go") {
			bytes, err := ioutil.ReadFile(path)
			require.Nil(t, err)
			pkg := filepath.Base(filepath.Dir(path))
			if !strings.Contains(string(bytes), fmt.Sprintf("package %v_test", pkg)) {
				result = append(result, path)
			}
		}
		return nil
	})
	require.Empty(t, result, "Please use the next pattern for tests pkg names'${package}_test'")
}
```
@edwarnicke  should we consider to add this kind tests on CI?

lobkovilya:
@denis-tingajkin this check doesn't prevent exactly this PR, but the idea is nice

edwarnicke:
@denis-tingajkin I'd be fine with that if it can be narrowly disabled.  99% of the time we want TestCases in _test package... but there are very very rare cases where you can't (say a finalizer test).

denis-tingajkin:
>but there are very very rare cases where you can't (say a finalizer test).

@edwarnicke Can we add exclude rule for this case, for example, add `if file name has not *internal_test.go`?

PR #105, opened 2020-02-16
Add eventchannel networkservice chain element

edwarnicke:
This provides machinery to allow the creation of various MonitorConnection clients/servers backed
by event channels.  This aids in construction of both adapters and the building of tests.

### DIFF
--- a/pkg/networkservice/core/eventchannel/monitor_client.go
+++ b/pkg/networkservice/core/eventchannel/monitor_client.go
@@ -1,0 +1,85 @@
+// Copyright (c) 2020 Cisco and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package eventchannel provides implementations based on event channels of:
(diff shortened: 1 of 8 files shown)


pkg/networkservice/core/eventchannel/monitor_client.go
```diff
@@ -0,0 +1,85 @@
...
+		go func() {
+			<-ctx.Done()
+			m.updateExecutor.AsyncExec(func() {
+				var newFanoutEventChs []chan *networkservice.ConnectionEvent
+				for _, ch := range m.fanoutEventChs {
+					if ch != fanoutEventCh {
+						newFanoutEventChs = append(newFanoutEventChs, ch)
+					}
+				}
+				m.fanoutEventChs = newFanoutEventChs
```
fkautz: a quick question on our convention for fanout.

Do we want to treat it as one word or two? fanout vs fanOut?

pkg/networkservice/core/eventchannel/monitor_connection_client.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (m *monitorConnectionMonitorConnectionsClient) CloseSend() error {
+	return nil
+}
+
+func (m *monitorConnectionMonitorConnectionsClient) Context() context.Context {
+	return m.ctx
+}
+
+func (m *monitorConnectionMonitorConnectionsClient) SendMsg(msg interface{}) error {
```
fkautz: There are several NOP methods here, is this intentional? If not, can we mark them with TODOs?

pkg/networkservice/core/eventchannel/monitor_client.go
edwarnicke: I think of it as one word, but am not strongly opinionated :)

pkg/networkservice/core/eventchannel/monitor_connection_client.go
edwarnicke: These are intentional NOOP.  The interface was designed for an HTTP/2 intermediary.  These are intentionally designed to be *channel* intermediated.  So these methods just have to *not explode* when called.

PR #109, opened 2020-02-17
Refactored the server_test.go to use the new adapters.

edwarnicke:
The intention was the make things simpler and easier to read and understand
by using the existing APIs rather than introducing new machinery.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -188,3 +188,8 @@ issues:
       linters:
         - govet
       text: "lostcancel"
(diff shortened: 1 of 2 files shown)


edwarnicke:
Note, only review the last commit, this is based on #108 which has yet to be merged.

edwarnicke:
@haiodo @lobkovilya Thoughts?

haiodo:
Looks really nice.

PR #111, opened 2020-02-17
Simplify serialize.Executor usage using go idioms

edwarnicke:
Replace Executor.AsyncExec(func()) and Executor.SyncExec(func()) with:

Executor.Exec(f func()) (done chan struct{})

which queues up f() for execution (in order) but return channel done which is closed when f() returns.

This allows for choosing between

exec.Exec(func(){...}) - returns immediately
<-exec.Exec(func(){...}) - returns after f() has returned

very similar to things like time.After(time.Duration)

Also eliminated the eventloop allowing serialize.Executor{} as a zero valued struct to be
used without initialization, in the same way as things like sync.Once{}, sync.Mutex{}, etc.

This did away with serialize.NewExecutor{}

Since we have no eventloop, we have no need to finalize either.


edwarnicke:
@nickolaev @fkautz @haiodo @denis-tingajkin Thoughts?

pkg/tools/serialize/serialize.go
```diff
@@ -19,59 +19,48 @@
...
-		exec()
+	done = make(chan struct{})
+	e.execCh <- func() {
+		f()
 		close(done)
 	}
-	<-done
+	// Drain if a go routine isn't already
+	if atomic.CompareAndSwapUint32(&e.drained, 0, 1) {
+		go func() {
```
haiodo: It is not clear when it will finish.

pkg/tools/serialize/serialize.go
```diff
@@ -19,59 +19,48 @@
...
-	rv := &executor{
-		execCh:      make(chan func(), 100),
-		finalizedCh: make(chan struct{}),
-	}
-	go eventLoop(rv.execCh, rv.finalizedCh)
-	runtime.SetFinalizer(rv, func(f *executor) {
-		close(f.finalizedCh)
+// Exec - guarantees f() will be executed Exclusively and in the Order submitted.
+//        It immediately returns a channel that will be closed when f() has completed execution.
+func (e *Executor) Exec(f func()) (done chan struct{}) {
```
haiodo: I would prefer to use some kind of Submit or leave AsyncExec name, Since Exec is not looks like async one.

edwarnicke: After some consideration, have decided you are correct, also decided to preserve backward compatibility with NewExecutor().  That simplifies the PR a lot.

pkg/tools/serialize/serialize.go
```diff
@@ -19,59 +19,50 @@
...
-		exec()
+// AsyncExec - guarantees f() will be executed Exclusively and in the Order submitted.
+//        It immediately returns a channel that will be closed when f() has completed execution.
+func (e *Executor) AsyncExec(f func()) (done chan struct{}) {
+	// Initialize *once*
+	e.init.Do(func() {
+		e.execCh = make(chan func(), channelSize)
+	})
+	done = make(chan struct{})
+	e.execCh <- func() {
```
haiodo: Should we move e.execCh <- after we do increment? It would be similar to

var wg sync.WaitGroup
for i := 0; i < b.N; i++ {
 	wg.Add(1)
 	go func() { wg.Done() }
 }
wg.Wait()

pkg/tools/serialize/serialize.go
```diff
@@ -19,59 +19,50 @@
...
+		f()
 		close(done)
 	}
-	<-done
+	// Start go routine if we don't have one
+	if atomic.AddInt32(&e.count, 1) == 1 {
+		go func() {
+			for g := range e.execCh {
+				g()
+				if atomic.AddInt32(&e.count, -1) == 0 {
```
haiodo: Is it will be good idea to add runtime.Gosched() call here?

pkg/tools/serialize/serialize.go
edwarnicke: I don't think so.  e.count represents the *unexecuted* func()s we have.  Just taking it off of the channel doesn't means its executed yet.  So we aren't really at e.count == 0 until g() has finished executing.  If someone adds a func() to the queue *while* its executing... then when we atomically deprecate e.count, it won't be zero, and we will go around the loop again :)

edwarnicke: Wait... totally misread... retract the last comment... one moment

edwarnicke: OK... we could increment before e.execCh <- , but it increases the complexity of the code, and in running the Benchmarks doesn't make it any faster:

```go
	count := atomic.AddInt32(&e.count, 1)
	e.execCh <- func() {
		f()
		close(done)
	}
	// Start go routine if we don't have one
	if count == 1 {
		go func() {
			for g := range e.execCh {
				g()
				if atomic.AddInt32(&e.count, -1) == 0 {
					return
				}
			}
		}()
	}
	return done
```

edwarnicke: Since there doesn't *appear* to be a perf benefit, and both seem to be correct, shall we go with the simpler code version?

pkg/tools/serialize/serialize_test.go
```diff
@@ -138,3 +139,20 @@ func TestAsyncExecOneAtATime(t *testing.T) {
 		<-done
 	}
 }
+
+func BenchmarkExecutorAsync(b *testing.B) {
```
denis-tingajkin: As I know `gotestsum` will not run bench tests on CI. Is it expected?

pkg/tools/serialize/serialize.go
```diff
@@ -19,59 +19,50 @@
...
 }
 
-// SyncExec - queues exec for execution and returns *after* exec() has run
-func (t *executor) SyncExec(exec func()) {
-	done := make(chan struct{})
-	t.execCh <- func() {
-		exec()
+// AsyncExec - guarantees f() will be executed Exclusively and in the Order submitted.
+//        It immediately returns a channel that will be closed when f() has completed execution.
+func (e *Executor) AsyncExec(f func()) (done chan struct{}) {
```
denis-tingajkin: Can we return read-only chan `done <- chan struct{}` here?

pkg/tools/serialize/serialize.go
edwarnicke: What would we accomplish with runtime.Gosched() there?  I'm reading this comment as putting the runtime.Gosched() *inside* the if statement... and inside the if statement we are about to return.

pkg/tools/serialize/serialize.go
edwarnicke: Good idea!  Done! ;)

pkg/tools/serialize/serialize.go
haiodo: Actually I'e mean following change:
```go
if atomic.AddInt32(&e.count, 1) == 1 {
	go func() {
		for g := range e.execCh {
			g()
			if atomic.AddInt32(&e.count, -1) == 0 {
				return
			}
		}
	}()
}
e.execCh <- func() {
	f()
	close(done)
}
```
In my experiments all variants looks working fine. So I'm ok having any of them. Moving e.execCh <- is few ms better on my env, not sure it is make any seance :-)

pkg/tools/serialize/serialize.go
haiodo: I've just accidentally found this one:
https://github.com/golang/go/issues/11462

And if we will run code in a container with 1 CPU, it should be better to do cooperative operations from different go routines. And this is just assumption.

pkg/tools/serialize/serialize_test.go
edwarnicke: I'd love to get to running Benchmarks on CI... just not sure what to *do* with them yet in CI  :)
Was using it locally to benchmark differences in implementation :)  On my laptop I'm averaging about 226 ns/op.  Since the function being tested itself is about 20 ns/op... I'm reading this as being basically a ~200ns tax on the function execution... which isn't bad ;)

pkg/tools/serialize/serialize.go
edwarnicke: I just pushed an update.  You are right, this is about 5% faster (at least when I move the channel allocate after as well) :)

pkg/tools/serialize/serialize.go
edwarnicke: Yeah... but it might also slow things down in other environments.  Context switching isn't free.  Generally speaking, run to completion is a lot faster when you can get it :)

pkg/tools/serialize/serialize_test.go
```diff
@@ -138,3 +139,20 @@ func TestAsyncExecOneAtATime(t *testing.T) {
...
+	var exec serialize.Executor
+	var wg sync.WaitGroup
+	for i := 0; i < b.N; i++ {
+		wg.Add(1)
+		exec.AsyncExec(func() { wg.Done() })
```
alex-yust: The benchmark might be more pure without using wg. To achieve this we can schedule b.N-1 executions inside the loop, and do the last execution after the loop using '<-exec.AsyncExec(func() {})' to ensure synchronization.

pkg/tools/serialize/serialize_test.go
denis-tingajkin: >I'd love to get to running Benchmarks on CI... just not sure what to do with them yet in CI :)

I've created the issue for adding bench tests on CI https://github.com/networkservicemesh/sdk/issues/117.

PR #116, opened 2020-02-19
Add discoverServer tests

lvfxx:
Added tests for /pkg/networkservice/common/discover package

### DIFF
--- a/pkg/networkservice/common/discover/server_test.go
+++ b/pkg/networkservice/common/discover/server_test.go
@@ -1,0 +1,153 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/discover/server_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (c *mockNetworkServiceDiscoveryClient) FindNetworkService(context.Context, *registry.FindNetworkServiceRequest, ...grpc.CallOption) (*registry.FindNetworkServiceResponse, error) {
+	return &c.response, nil
+}
+
+var discoveryClientResponse = &registry.FindNetworkServiceResponse{
+	NetworkService: &registry.NetworkService{
+		Name: "secure-intranet-connectivity",
+		Matches: []*registry.Match{
+			{
```
lvfxx: If i move Match with empty source selector to beginning of the Matches array, second test case will fail. Is there any guarantee that Match with empty source selector goes last?

edwarnicke: The matches should be considered in order.  So if the match with empty selector comes first, it should match.   There is no guarantee that Match with empty source selector goes last.

edwarnicke: Filed https://github.com/networkservicemesh/sdk/issues/119 on this.

PR #122, opened 2020-02-21
Add localBypass tests

lvfxx:
Added test for /pkg/networkservice/common/localbypass
Also moved localBypassRegistry to /pkg/registry/common/localbypass


lvfxx:
I'm not entirely sure about [last commit](https://github.com/networkservicemesh/sdk/pull/122/commits/215bedc783cd2bc461e9b5c2ec9ab9167671790d). localBypass registryServer and networkServiceServer look quite strongly connected and i'm not sure if registry Server should be visible beyond package at all?

pkg/networkservice/common/localbypass/server.go
```diff
@@ -23,6 +23,8 @@ import (
 	"net/url"
```
edwarnicke: Thanks for the file rename here :)

Issue #119, opened 2020-02-20
Empty Match does not evaluate properly unless last


edwarnicke:
The contract for Matches in the Network Service is that we evaluate each one from first to last until one matches.  It appears this is not currently the case:

https://github.com/networkservicemesh/sdk/pull/116#discussion_r381220661

PR #125, opened 2020-02-21
Fix match order in discoverCandidatesServer

lvfxx:
Implementation of #119

### DIFF
--- a/pkg/networkservice/common/discover/server_test.go
+++ b/pkg/networkservice/common/discover/server_test.go
@@ -21,29 +21,52 @@ import (
 	"context"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 	"github.com/networkservicemesh/api/pkg/api/registry"
-	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/discover"
 )
 
+type contextKeyType string
+
+const (
+	testDataKey contextKeyType = "testData"
+)
+
+type TestData struct {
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/discover/server_test.go
```diff
@@ -21,133 +21,237 @@ import (
...
+	}
+	return nil
+}
+
 type testNetworkServiceServer struct {
-	t    *testing.T
-	want []*registry.NetworkServiceEndpoint
 }
 
 func (s testNetworkServiceServer) Request(ctx context.Context, in *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
```
haiodo: I think it become more complicated, then it supposed it be. For such scenarios we should have a Network Service accepting function, and function should do assertion inplace.

 Like 
```go
request := ...
server := next.NewNetworkServiceServer(discover.NewServer(discoveryClient), &AssertNetworkService{
Before: func(req networkservice.NetworkServiceRequest ) {
  // do required checks.
},
After: func(conn *connection.Connection)  {
 // do checks on connection
}
})
```

Thoughts?

edwarnicke: You should be able to accomplish checking after with

[checkcontext](https://github.com/networkservicemesh/sdk/tree/master/pkg/networkservice/utils/checks/checkcontext)

```golang
testServer := chain.NewNetworkServiceServer(
   discover.NewServer(discoveryClient),
   checkcontext.NewServer(func(t *testing.T, ctx context.Context),
}
```

If you need to check something about the after you can use: [checkrequest](https://github.com/networkservicemesh/sdk/tree/master/pkg/networkservice/utils/checks/checkrequest) or if you need to check the connection on return you can use [checkconnection](https://github.com/networkservicemesh/sdk/tree/master/pkg/networkservice/utils/checks/checkconnection).  

Finally, if you need to check something about the context *on* return (rather than after) you can use [checkcontextonreturn](https://github.com/networkservicemesh/sdk/tree/master/pkg/networkservice/utils/checks/checkcontextonreturn)

PR #145, opened 2020-02-25
Add setId tests

lvfxx:
Added tests for pkg/networkservice/common/setid package

### DIFF
--- a/pkg/networkservice/common/setid/client_test.go
+++ b/pkg/networkservice/common/setid/client_test.go
@@ -27,109 +27,60 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/setid"
 )
 
-func TestNewClient_SetNewConnectionId(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/setid/client_test.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func TestNewClient_PathSegmentNameEqualClientName(t *testing.T) {
+	connectionID := "54ec77a7-d645-43ca-87bc-ab51765f574a"
+	request := &networkservice.NetworkServiceRequest{
+		Connection: &networkservice.Connection{
+			Id: connectionID,
+			Path: &networkservice.Path{
+				Index: 1,
+				PathSegments: []*networkservice.PathSegment{
```
haiodo: What the difference with passed path segments? I think better to create function to create request.

lvfxx: The difference between requests is only path index and connection id. Factored out into separate function.

PR #148, opened 2020-02-26
Refactor clientInfo networkService tests

lvfxx:


pkg/networkservice/common/clientinfo/client_test.go
```diff
@@ -26,110 +26,115 @@ import (
...
+		"PodNameKey":     "BBB",
+		"ClusterNameKey": "CCC",
+		"SomeOtherLabel": "DDD",
+	}
+
+	err := setEnvs(envs)
+	assert.Nil(t, err)
+
+	client := clientinfo.NewClient()
+	conn, err := client.Request(context.Background(), &networkservice.NetworkServiceRequest{
```
edwarnicke: Lets us https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/utils/checks/checkrequest/client.go#L39 here to check to make sure the labels are set before after the client in the chain, and not mistakenly 'on return'.  Current checking the return only tells us if the labels are set on return (as done here).

lvfxx: Ok, fixed it

edwarnicke:
@lvfxx Any idea why the tests are failing?

lvfxx:
> @lvfxx Any idea why the tests are failing?

I think i need to rebase onto `master`, this flaky test is already fixed

PR #149, opened 2020-02-26
Refactor clientUrl tests

lvfxx:


pkg/networkservice/common/clienturl/server_test.go
```diff
@@ -22,84 +22,108 @@ import (
...
+	testData(ctx).clientURL = clienturl.ClientURL(ctx)
+	return next.Server(ctx).Close(ctx, conn)
+}
+
+func TestAddURLInEmptyContext(t *testing.T) {
+	clientURL := &url.URL{
+		Scheme: "ipv4",
+		Path:   "192.168.0.1",
 	}
+	client := next.NewNetworkServiceServer(clienturl.NewServer(clientURL), &testNetworkServiceServer{})
```
edwarnicke: Why not use https://github.com/networkservicemesh/sdk/tree/master/pkg/networkservice/utils/checks/checkcontext here?

pkg/networkservice/common/clienturl/server_test.go
```diff
@@ -22,84 +22,108 @@ import (
...
+
+	ctx := withTestData(context.Background(), &TestData{})
+	_, err := client.Request(ctx, &networkservice.NetworkServiceRequest{})
+	assert.Nil(t, err)
+	assert.Equal(t, clientURL, testData(ctx).clientURL)
+
+	ctx = withTestData(context.Background(), &TestData{})
+	_, err = client.Close(ctx, &networkservice.Connection{})
+	assert.Nil(t, err)
+	assert.Equal(t, clientURL, testData(ctx).clientURL)
```
edwarnicke: Why not just use https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/clienturl/context.go#L44 plus https://github.com/networkservicemesh/sdk/tree/master/pkg/networkservice/utils/checks/checkcontext ?

pkg/networkservice/common/clienturl/server_test.go
lvfxx: Switched to this utility, thanks

pkg/networkservice/common/clienturl/server_test.go
lvfxx: The reason was to factor out all assertions into test method body to make them obvious. Your solution looks simpler, thanks

PR #152, opened 2020-02-27
Add util functions for tokens

lobkovilya:
## Description
Add two functions:
- `GenerateToken(ctx context.Context, p Provider, expiresAt time.Duration) (string, error)`
- `VerifyToken(token string, x509crt *x509.Certificate) error` 

Cover them with unit tests.

Signed-off-by: Ilya Lobkov <ilya.lobkov@xored.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -201,6 +201,3 @@ issues:
       linters:
         - dupl
       text: "lines are duplicate of"
-    - path: pkg/tools/security/token_test.go
-      linters:
-        - gochecknoinits
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -201,3 +201,6 @@ issues:
       linters:
         - dupl
       text: "lines are duplicate of"
+    - path: pkg/tools/security/token_test.go
+      linters:
+        - gochecknoinits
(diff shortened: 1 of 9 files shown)


.golangci.yml
```diff
@@ -201,3 +201,6 @@ issues:
       linters:
         - dupl
       text: "lines are duplicate of"
+    - path: pkg/tools/security/token_test.go
+      linters:
+        - gochecknoinits
```
denis-tingajkin: Why do we need to disable this linter?

pkg/tools/security/token_test.go
```diff
@@ -0,0 +1,230 @@
...
+	testCA, err = generateCA()
+	if err != nil {
+		panic(err)
+	}
+
+	testTLSCertificate, err = generateKeyPair(spiffeID, "test.com", &testCA)
+	if err != nil {
+		panic(err)
+	}
+}
```
denis-tingajkin: Is it possible to use testify/suite for setup tests? Tesitfy suites can have setup/teardown methods. take a look at https://github.com/stretchr/testify/blob/master/suite/interfaces.go#L20

.golangci.yml
lobkovilya: Fixed

pkg/tools/security/token_test.go
lobkovilya: Fixed

PR #154, opened 2020-03-02
Authn chain elements

lobkovilya:
Depends on #152 

## Description 
Add 2 chain elements:
- `AuthnClient` adds token to the current `PathSegment` (pointed by `Index`)
- `AuthnServer` verifies token  of the current `PathSegment`

### DIFF
--- a/pkg/networkservice/common/authn/server.go
+++ b/pkg/networkservice/common/authn/server.go
@@ -1,0 +1,77 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authn
+
+import (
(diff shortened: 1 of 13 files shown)


denis-tingajkin:
@edwarnicke I think this PR can be useful for **Advanced OPA Policies**. Can you please take a look?

pkg/networkservice/common/authn/server.go
```diff
@@ -0,0 +1,82 @@
...
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/security"
+)
+
+type authnServer struct {
+	p security.Provider
```
edwarnicke, line 37: I really don't want to be promulgating additional APIs around... much of the stuff in the security package isn't needed now that we have grpcoptions.

pkg/networkservice/common/authn/server.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (a *authnServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
+	x509crt, err := peerCertificateFromContext(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Unauthenticated, "unable to get peer's TLS cert from context: %v", err)
+	}
+
+	token := request.GetConnection().GetPath().GetPathSegments()[request.GetConnection().GetPath().GetIndex()].GetToken()
+	if err := security.VerifyToken(token, x509crt); err != nil {
```
edwarnicke, line 55: Can't we do the token verification in OPA?

denis-tingajkin: I think it is a good idea, will do :)

PR #160, opened 2020-03-17
Fix race condition in MonitorConnectionClient

lvfxx:
**Problem:**
Undefined order of execution between `monitorConnectionClient.eventLoop()` goroutine that passes events to `fanoutEventCh` and goroutine that's waiting `<--context.Done()` and closes `fanoutEventCh` leads to different number of `testOnHeal.Request()` calls in `/pkg/networkservice/common/heal/client_test.go`
`TestNewClient_MissingConnectionsOnInit` test that leads to different number (2 or 3) `testOnHeal.Request()` invocations. [Explanation diagram](https://drive.google.com/file/d/1yDwcyOjTQrcif8B_RvndGDDyxH3qzmN3/view?usp=sharing)

**Solutions i see:**
1) `time.Sleep(100 * time.Millisecond)` after sending `INITIAL_STATE_TRANSFER` event to ensure that this event will be handled earlier than context cancellation.

2) Guarantee that chain context cancellation will be handled after connection event, if it was cancelled after connection event sending. One way to do that i see is to emulate the server falling down by closing `eventCh`, not chain context cancelling. Chain context cancelling is rather about client's intention to shut down and logically may be async, but server's falling down is not.

Tried to implement the second one in this PR :)

### DIFF
--- a/pkg/networkservice/common/heal/client_test.go
+++ b/pkg/networkservice/common/heal/client_test.go
@@ -194,6 +194,7 @@ func TestNewClient_MissingConnectionsInInit(t *testing.T) {
 	}
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
 	client := chain.NewNetworkServiceClient(
 		heal.NewClient(ctx, eventchannel.NewMonitorConnectionClient(eventCh), addressof.NetworkServiceClient(onHeal)))
 
@@ -241,5 +242,4 @@ func TestNewClient_MissingConnectionsInInit(t *testing.T) {
 	require.Eventually(t, cond, waitForTimeout, tickTimeout)
 	require.Equal(t, 0, healsRemaining[conns[0].GetId()])
 	require.Equal(t, 0, healsRemaining[conns[1].GetId()])
-	cancelFunc()
 }
--- a/pkg/networkservice/common/heal/client_test.go
+++ b/pkg/networkservice/common/heal/client_test.go
@@ -217,16 +217,18 @@ func TestNewClient_MissingConnectionsInInit(t *testing.T) {
 
 	// we emulate situation that server managed to handle only the first connection
 	// second connection should came in the UPDATE event, but we emulate server's falling down
-	cancelFunc()
+	close(eventCh)
 	// at that point we expect that 'healClient' start healing both 'conn-1' and 'conn-2'
 
-	healedIDs := map[string]bool{}
-
+	healsRemaining := map[string]int{
+		conns[0].GetId(): 1,
+		conns[1].GetId(): 2,
+	}
 	cond := func() bool {
 		select {
 		case r := <-requestCh:
-			if _, ok := healedIDs[r.GetConnection().GetId()]; !ok {
-				healedIDs[r.GetConnection().GetId()] = true
+			if val, ok := healsRemaining[r.GetConnection().GetId()]; ok && val != 0 {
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/heal/client_test.go
```diff
@@ -180,11 +193,15 @@ func TestHealClient_EmptyInit(t *testing.T) {
 			},
 		},
 	}
+
+	cancelFunc()
+	close(eventCh)
```
alex-yust: using 'defer' for such call is preferred, since it guaranties the call for all exit points.
is it a special case?

pkg/networkservice/core/eventchannel/monitor_client.go
```diff
@@ -50,13 +50,16 @@ func NewMonitorConnectionClient(eventCh <-chan *networkservice.ConnectionEvent)
...
+func (m *monitorConnectionClient) MonitorConnections(ctx context.Context, _ *networkservice.MonitorScopeSelector, _ ...grpc.CallOption) (networkservice.MonitorConnection_MonitorConnectionsClient, error) {
 	fanoutEventCh := make(chan *networkservice.ConnectionEvent, 100)
 	m.updateExecutor.AsyncExec(func() {
 		m.fanoutEventChs = append(m.fanoutEventChs, fanoutEventCh)
 		go func() {
 			<-ctx.Done()
 			m.updateExecutor.AsyncExec(func() {
+				if len(m.fanoutEventChs) == 0 {
+					return
+				}
```
alex-yust: Can this check be moved outside the lambda, right after <-ctx.Done()?

pkg/networkservice/core/eventchannel/monitor_client.go
```diff
@@ -50,13 +50,16 @@ func NewMonitorConnectionClient(eventCh <-chan *networkservice.ConnectionEvent)
 	return rv
 }
 
-func (m *monitorConnectionClient) MonitorConnections(ctx context.Context, in *networkservice.MonitorScopeSelector, _ ...grpc.CallOption) (networkservice.MonitorConnection_MonitorConnectionsClient, error) {
+func (m *monitorConnectionClient) MonitorConnections(ctx context.Context, _ *networkservice.MonitorScopeSelector, _ ...grpc.CallOption) (networkservice.MonitorConnection_MonitorConnectionsClient, error) {
 	fanoutEventCh := make(chan *networkservice.ConnectionEvent, 100)
 	m.updateExecutor.AsyncExec(func() {
 		m.fanoutEventChs = append(m.fanoutEventChs, fanoutEventCh)
```
alex-yust: Is it Ok to modify m.fanoutEventChs from multiple go-routines without a lock?
@haiodo , @denis-tingajkin ?

pkg/networkservice/core/eventchannel/monitor_client.go
lvfxx: I think no, because we need synchronized access to `m.fanoutEventChs`, that guaranteed by `m.updateExecutor`

pkg/networkservice/core/eventchannel/monitor_client.go
lvfxx: `m.updateExecutor.AsyncExec()` guarantees exclusive access

pkg/networkservice/common/heal/client_test.go
lvfxx: Good idea, thanks :)

PR #162, opened 2020-03-17
Add proper termination to healClient tests

lvfxx:


lvfxx:
Except `TestNewClient_MissingConnectionsOnInit` test. See #160 about it

pkg/networkservice/common/heal/client.go
```diff
@@ -163,12 +163,11 @@ func (f *healClient) Request(ctx context.Context, request *networkservice.Networ
 	// Set its connection to the returned connection we received
 	req.Connection = rv
 
-	// TODO handle deadline err
-	deadline, _ := ctx.Deadline()
-	duration := time.Until(deadline)
+	// TODO define proper constant timeout
+	timeout := time.Minute
```
edwarnicke: Why 1 minute here?

denis-tingajkin:
```suggestion
	// TODO handle deadline err
	deadline, _ := ctx.Deadline()
	duration := time.Until(deadline)
	f.updateExecutor.AsyncExec(func() {
		f.requestors[req.GetConnection().GetId()] = func() {
			        // TODO rename timeCtx
		                 var  timeCtx context.Context
		                 var  cancelFunc func()
		                 if duration > 0 {
		                                 timeCtx, cancelFunc = context.WithTimeout(f.chainContext, duration)
		                 } else {
		                                 timeCtx, cancelFunc = context.WithCancel(f.chainContext)
		                 }
```

lvfxx:
```suggestion
	deadline, hasDeadline  := ctx.Deadline()
	f.updateExecutor.AsyncExec(func() {
		f.requestors[req.GetConnection().GetId()] = func() {
			        // TODO rename timeCtx
		                 var  timeCtx context.Context
		                 var  cancelFunc func()
		                 if hasDeadline {
		                                 timeCtx, cancelFunc = context.WithDeadline(f.chainContext, deadline)
		                 } else {
		                                 timeCtx, cancelFunc = context.WithCancel(f.chainContext)
		                 }
```
I like your variant, but i wonder why don't just use `context.WithDeadline()`? When use `context.WithDuration()` context for `f.onHeal` may remain valid after request's context deadline expiration due to possibly defferred execution of `f.updateExecutor.AsyncExec(func() {...})`. But with `context.WithDeadline()` it's guaranteed that both contexts have the same cancellation moment.

lvfxx: @edwarnicke any thoughts?

edwarnicke: @lvfxx Excellent question.  Heal *intrinsically* may happen long after the deadline of the original request that created the connection.  I may have a request that comes in with 10 seconds till its deadline, resulting in a connection with an ExpireTime 1 day in the future.  12 hours in, healing is triggered.  We are basically using the initial connection to *hint* at a reasonable deadline (the same duration in the future from when we start healing)... not a source of an absolute deadline.

lvfxx: Ah, got it, thanks :)

lvfxx: I think i should remove constant timeout at all...

PR #166, opened 2020-03-22
Split healClient to healClient and monitorClient

lvfxx:


pkg/tools/heal/healer.go
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package heal provides healClient related tools
+package heal
+
+// Healer - interface for initiating healing process by monitorClient when it detects broken connection
+type Healer interface {
+	StartHeal(connectionID string)
```
edwarnicke: Lets no introduce new interfaces.

pkg/networkservice/common/monitor/client.go
```diff
@@ -0,0 +1,148 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/heal"
+	"github.com/networkservicemesh/sdk/pkg/tools/serialize"
+)
+
+type monitorClient struct {
+	grpcClient    networkservice.MonitorConnectionClient
+	healer        heal.Healer
```
denis-tingajkin: Why monitor client should depend on `heal`l?

lvfxx: It should somehow initiate the healing process when connection breaks down. Another way i see is to pass healing functions from `healClient` within context of connection request.

denis-tingajkin: Can we just use dependency injection?
For example, we can create an interface and pass it into `monitor.Client`
``` go
type ConnectionEventHandler interface {
    Handle(*ConnectionEvent)
}
```
And for `heal` case we need to just implement this interface. 
For example:

```go
type healConnectionEventHandler struct {
}
func (h *healConnectionEventHandler) Handle(e *ConnectionEvent) {
// StartHeal
}
func NewHealConnectionEventHandler() ConnectionEventHandler {
    return &healConnectionEventHandler{}
}
```

lvfxx: As i understood this way we don't need `monitoring` and `reported` maps in `monitor.Client` -- all this job will be done by `heal.Client`, so we get more concrete responsibilities of each client, cool! We also need to add another method to `healConnectionEventHandler` that triggers when connection to `MonitorConnectionServer` breaks down itself. I'll try to implement your suggestion.

edwarnicke: What is the underlying motivation of this split?

denis-tingajkin: nice catch!
@lvfxx Let's create an Issue for this PR.

>What is the underlying motivation of this split?

In short, we want to extract monitoring client from heal pkg to reuse it in different places. For example for `dnscontext.Client` component.

edwarnicke: @denis-tingajkin But why... the amount of reuse is small... but it introduces relatively complex new relationships... I don't mind replicating code in places if it's small and makes it easier to comprehend in one place what's happening.

lvfxx:
Close because this introduces unnecessary complication

PR #168, opened 2020-03-27
Fix time.Stop() may not stop already started refresh

lvfxx:
Handle possible situation when `time.Stop()` doesn't stop already started refresh gotoutine from previous `Request()` or at `Close()` invocation

### DIFF
--- a/pkg/networkservice/common/refresh/client.go
+++ b/pkg/networkservice/common/refresh/client.go
@@ -24,10 +24,9 @@ import (
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
-
-	"github.com/networkservicemesh/api/pkg/api/networkservice"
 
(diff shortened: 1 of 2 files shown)


lvfxx:
Covered this cases by tests in #169

edwarnicke:
Could you say a few words about what problem this is solving that timer.Stop wasn't... I'm not quite following.

pkg/networkservice/common/refresh/client.go
```diff
@@ -24,79 +24,92 @@ import (
...
 		return nil, errors.Wrap(err, "Error calling next")
 	}
 	// Clone the request
 	req := request.Clone()
 	// Set its connection to the returned connection we received
 	req.Connection = rv
-	// Setup the timer with the req containing the returned connection
-	ct, err := t.createTimer(ctx, req, opts...)
+
+	expireTime, err := ptypes.Timestamp(request.GetConnection().GetPath().GetPathSegments()[request.GetConnection().GetPath().GetIndex()].GetExpires())
```
haiodo, line 67: @edwarnicke What do you think about having minimal granularity for token expire time, like 5 seconds.
So we could have just one timer to check expires for every 5 seconds for all connections, in this case it won't be required to create times and could be easy doable with executor AsyncExec and sleep.

PR #169, opened 2020-03-27
Add refreshClient tests

lvfxx:


edwarnicke:
Could you say a few words about what problem this is solving that timer.Stop wasn't... I'm not quite following.

edwarnicke:
Would it make sense to roll the testing into #168

PR #171, opened 2020-03-30
Support for grpc.callback APIs

haiodo:
Support for grpc.callback APIs using bi-directionsl grpc streams.

Serve GRPC via one client to server connection.
So it could be used to perform a server callback to client using one client connection GRPC server.

image

SPEC why it is importsnt is here: https://docs.google.com/document/d/1Lmlcg_qEOQDLdNMCKUL_K-rM2c2lK2lxwKCbPbscCRg/edit?usp=sharing


edwarnicke:
See  https://docs.google.com/document/d/1Lmlcg_qEOQDLdNMCKUL_K-rM2c2lK2lxwKCbPbscCRg/edit?usp=sharing for questions :)

pkg/tools/callback/readme.md
```diff
@@ -0,0 +1,51 @@
...
+
+Server could identify clients from passed metadata values, default identity is 
+using ":authority" passed from client to server.
+
+```go
+md, ok := metadata.FromIncomingContext(serverClient.Context())
+if !ok {
+    logrus.Errorf("Not metadata provided")
+}
+key := s.identityProvider(md)
```
edwarnicke, line 50: Could [peer](https://godoc.org/google.golang.org/grpc/peer) be used here?

haiodo: Changed to pass context to call so it will be possible.

edwarnicke: I guess the question is: could we simply use grpc.Peer instead of metadata.FromIncomingContext(serverClient.Context())

edwarnicke: OK... I get it now.  Authority is just one example here... you could be getting identity from the TLS information as well using a different IdentityProvider.

haiodo: from TLS yes, we could use grpc.Peer to get what information. Added a test to check this exact case.

pkg/tools/callback/readme.md
```diff
@@ -0,0 +1,51 @@
...
+callbackServer := callback.NewServer(callback.IdentityByAuthority)
+//
+callback.RegisterCallbackServiceServer(server, callbackServer)
+...
+_ = server.Serve(listener)
+```
+
+And then require to call client.
+```go
+nsmClientGRPC, _ := callbackServer.NewClient(context.Background(), clientID)
```
edwarnicke, line 30: Its looks like what we have here is a callbackServer.NewClient(...) that is returning a grpc.ClientConnInterface

Given that, might it not make more sense to mirror more closely what we already have API wise from grpc?

perhaps something like
```golang
cc := callbackServer.DialContex(ctx,target)
nsmClient := networkservice.NetworkServiceClient(cc)
```

Honestly I'd prefer if we could figure it out to have something like:
```golang
cc := callback.DialContex(ctx,target,WithCallbackServer(...))
nsmClient := networkservice.NetworkServiceClient(cc)
```
Where target can be in some form of a callback URL that indicates we are doing callback, WithCallBackServer(...) provides whatever parameters we need to start the CallBackServer if its not present... and if the target is *not* a callback URL we just fall through to the normal grpc.DialContext(...)

This approach would then allow folks to take advantage of callback if and only if they got a callback URL, and otherwise follow more or less the same pattern they use now for GRPC clients.  That said... all of the details of how to do that are not yet clear in my head.

haiodo: Sounds good, will check and update PR.

pkg/tools/callback/readme.md
```diff
@@ -0,0 +1,67 @@
...
+...
+_ = server.Serve(listener)
+```
+
+And then require to call client.
+```go
+target := "callback:{client-authority}"
+// If target is not in a list of callback server targets, 
+// it will perform grpc.DialContext to connect to passed target
+nsmClientGRPC, err := grpc.DialContext(context.Background(), target, callback.WithCallbackDialer(server.callbackServer, target), grpc.WithInsecure())
```
edwarnicke, line 33: Does this require grpc.WithInsecure()?

edwarnicke: Or to put it differently... can I use grpc.Peer from within the callback client?  It would be quite handy for the client using the callback to be able to get the correct AuthInfo for its peer :)

haiodo: Without insecure, it will be grpc error showing it require authentication, SSL could be used for this connection as well, but since it is passed using already authenticated connection I suppose better to not use it. It require individual grpc server on client to handle callbacks from particular server, so not sure if additional authentication or identification is required. Potentially we could enhance solution and have just one grpc for callbacks on client and connections to different servers, but current implementation doesn't support so.

edwarnicke: We will need to be able to get peer information for clients to properly compute tokens... is there any way to have this properly *set* the peer information?

haiodo: I've only found a way to retrieve client certificate.

haiodo: https://github.com/networkservicemesh/sdk/pull/171/files#diff-bcd6aa47c15ba3f1dd29728755ef360bR51

edwarnicke: OK... looking at the code more closely... this will in fact correctly handle the case of using the TLS in TLS if need be... or some simpler workarounds for Token Generation.

Issue #172, opened 2020-03-31
next::Client doesn't work with chained elements


lazyniv:
# Motivation

When we pass the chained client element into next.NewNetworkServiceClient(), unexpected behaviour may occur in next.Request -- it will be impossible to make request through all the clients in the chain.

denis-tingajkin:
Fixed by https://github.com/networkservicemesh/sdk/pull/173

PR #173, opened 2020-03-31
Add the possible to make chain of chains via pkg/next

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/172

### DIFF
--- a/pkg/networkservice/core/next/client.go
+++ b/pkg/networkservice/core/next/client.go
@@ -41,7 +41,7 @@ type ClientChainer func(...networkservice.NetworkServiceClient) networkservice.N
 
 // NewWrappedNetworkServiceClient chains together clients with wrapper wrapped around each one
 func NewWrappedNetworkServiceClient(wrapper ClientWrapper, clients ...networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
-	rv := &nextClient{}
+	rv := &nextClient{clients: make([]networkservice.NetworkServiceClient, 0, len(clients))}
 	for _, c := range clients {
 		rv.clients = append(rv.clients, wrapper(c))
 	}
@@ -59,7 +59,7 @@ func NewNetworkServiceClient(clients ...networkservice.NetworkServiceClient) net
 }
 
 func (n *nextClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
-	if n.index == 0 {
+	if n.index == 0 && ctx != nil {
 		if nextParent := Client(ctx); nextParent != nil {
 			n.nextParent = nextParent
 		}
@@ -71,7 +71,7 @@ func (n *nextClient) Request(ctx context.Context, request *networkservice.Networ
 }
 
 func (n *nextClient) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
-	if n.index == 0 {
+	if n.index == 0 && ctx != nil {
 		if nextParent := Client(ctx); nextParent != nil {
 			n.nextParent = nextParent
 		}
--- a/pkg/networkservice/core/next/client.go
+++ b/pkg/networkservice/core/next/client.go
@@ -28,8 +28,9 @@ import (
 )
 
 type nextClient struct {
-	clients []networkservice.NetworkServiceClient
-	index   int
+	clients    []networkservice.NetworkServiceClient
+	index      int
+	nextParent networkservice.NetworkServiceClient
 }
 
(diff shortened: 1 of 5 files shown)


edwarnicke:
I'm not following the problem we are trying to solve here... could you say more?

It's unclear to me why the existing code does not allow you to use one chain as an element in another.

denis-tingajkin:
> we are trying to solve here... could you say more?

We are trying to use chains inside chains. For example
```go
next.NewServer(next.NetServer(somepkg.NewServer()), somepkg.NewServer())
```

> It's unclear to me why the existing code does not allow you to use one chain as an element in another.

The problem related to two things:
1) `tailServer`, `tailClient` elements  are not using `next.Client()`/`next.Servers()` (the chains brokes here)
2) Inner chain elements rewrites value of`nextServer`/`nextClient`

edwarnicke:
Oh... that is nicely subtle.  Good catch.

Question thought... might it not be simpler and clearer to simply do something like replacing

https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/core/next/server.go#L64

```golang
return n.servers[n.index].Request(withNextServer(ctx, nil), request)
```

with 
```golang
return n.servers[n.index].Request(ctx, request)
```

That way any 'nextServer' coming from the parent would show up for the child?

denis-tingajkin:
@edwarnicke good idea. Thanks, fixed.

pkg/networkservice/core/next/context.go
```diff
@@ -48,6 +48,9 @@ func withNextServer(parent context.Context, next networkservice.NetworkServiceSe
 // Server -
 //   Returns the Server networkservice.NetworkServiceServer to be called in the chain from the context.Context
 func Server(ctx context.Context) networkservice.NetworkServiceServer {
+	if ctx == nil {
```
edwarnicke: Question... why return nil here instead of tailServer?

pkg/networkservice/core/next/client.go
```diff
@@ -54,23 +53,31 @@ func NewNetworkServiceClient(clients ...networkservice.NetworkServiceClient) net
 	if len(clients) == 0 {
 		return &tailClient{}
 	}
-	return NewWrappedNetworkServiceClient(notWrapClient, clients...)
+	return NewWrappedNetworkServiceClient(func(client networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
```
edwarnicke: This change is very slick :)

pkg/networkservice/core/next/client.go
```diff
@@ -40,11 +41,9 @@ type ClientChainer func(...networkservice.NetworkServiceClient) networkservice.N
...
 // NewWrappedNetworkServiceClient chains together clients with wrapper wrapped around each one
 func NewWrappedNetworkServiceClient(wrapper ClientWrapper, clients ...networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
-	rv := &nextClient{
-		clients: clients,
-	}
-	for i := range rv.clients {
-		rv.clients[i] = wrapper(rv.clients[i])
+	rv := &nextClient{}
+	for _, c := range clients {
+		rv.clients = append(rv.clients, wrapper(c))
```
edwarnicke: Good idea to make a new array here.

pkg/networkservice/core/next/context.go
denis-tingajkin: actually it can be nil in two cases:
1. At the first call of method request/close with `ctx = nil`. In this case, we looking for the next parent chain and value can be nil
2. Call this method outside of the chain. I.e. user not used `next.NewServer()`

denis-tingajkin: Finally, for the second case I think it will be better just throw nil panic exception. Fixed by https://github.com/networkservicemesh/sdk/pull/173/commits/1d5588cc84d32b48e825132a8f5912d8c3102afd

edwarnicke: Why would we panic if called outside of a chain?

denis-tingajkin: 1. We have analogies here:
https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/core/trace/context.go#L43
https://github.com/networkservicemesh/sdk-vppagent/blob/master/pkg/networkservice/vppagent/context.go#L50
...
2. I think when the user tries to use nil context to get the value it is an invalid case. And we should not work with invalid states.

denis-tingajkin: @edwarnicke Is it make sense?

edwarnicke: That's fair :)

PR #181, opened 2020-04-08
Only log the first stack trace verbosely

edwarnicke:


pkg/tools/spanhelper/span_helper.go
```diff
@@ -108,6 +108,7 @@ type SpanHelper interface {
 	LogObject(attribute string, value interface{})
 	LogValue(attribute string, value interface{})
 	LogError(err error)
+	LogErrorf(format string, err error)
```
denis-tingajkin, line 111: Not sure about this API. Because we impose a restriction on the format string that it should have only one format arg. It can be not clear for users.
Probably we can use something like this:
```
LogErrorf(msg string, err error) 
```

PR #199, opened 2020-04-14
Return callback tests

haiodo:
Pr to restore callback tests and find why they could fail.


fkautz:
You may need to resubmit this. :)

PR #207, opened 2020-04-20
Add journaling to streaming NATS

fkautz:
Creates a journal and streams it to streaming NATS (STAN) as a json entry.

The journal stores IP src,dst,event type (request,close) and path information.

The journal will be used to repair IPAM and can be used for additional observability capabilities.


pkg/networkservice/common/journal/server.go
```diff
@@ -0,0 +1,117 @@
...
+	Action      string
+	Path        *networkservice.Path
+}
+
+type journalServer struct {
+	journalID string
+	nats      stan.Conn
+}
+
+func (srv journalServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
```
edwarnicke, line 57: Do you want the receiver here to be journalServer or *journalServer ?

PR #211, opened 2020-04-20
Fix windows build/test

fkautz:
There will still be other Linux-isms but this will get the build/test cycle working.

Signed-off-by: Frederick F. Kautz IV <fkautz@alumni.cmu.edu>

### DIFF
--- a/pkg/networkservice/common/mechanisms/kernel/server.go
+++ b/pkg/networkservice/common/mechanisms/kernel/server.go
@@ -19,10 +19,11 @@ package kernel
 
 import (
 	"context"
+	"strconv"
+
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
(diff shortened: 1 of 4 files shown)


pkg/networkservice/common/mechanisms/kernel/server.go
```diff
@@ -33,12 +32,12 @@ type mechanismsServer struct{}
 
 func (m mechanismsServer) Request(ctx context.Context, req *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
 	conn := req.Connection
-	inode := uint64(0)
+	inode := uintptr(0)
```
edwarnicke, line 35: Why uintptr here?

fkautz: It's guaranteed to hold a pointer based on the platform. This will be important if we see 32 bit platforms. E.g. most Raspberry PIs are 32 bit.

Issue #200, opened 2020-04-14
Generalize authorize chain element


edwarnicke:
The current chain element looks *only* at the last token in the path chain.  This is insufficient for the full spectrum of policy we are likely to want to enforce.

Expand on the authorize chain element by providing as its 'input' object:

1.  The Connection
2.  The TLSInfo (as retrieved by peer.FromContext for Server or grpc.Peer for the client)
3.  Operation - One of Request/Close
4.  Role - One of Client/Endpoint

edwarnicke:
Related to #51

lazyniv:
How do you think, do we need to add all of the TLSInfo.State content into the OPA input? Or we need to add only a X509SVID certificate from TLSInfo.State. I think that in our cases we only need a pem-encoded X509SVID certificate to verify the signature of the jwt token using the [built-in OPA functions](https://www.openpolicyagent.org/docs/latest/policy-reference/#using-pem-encoded-x-509-certificate). Of course we also need to add the spiffeID from this certificate to OPA input

edwarnicke:
@lazyniv I think you are right on wrt just using the pem-encoded X509SVID :)

PR #225, opened 2020-04-24
Prepare input for OPA

lazyniv:
Now OPA input provides the following objects:
1. All information about connection (usage: input.connection.[...]. For example, a token from zero path segment we can get as input.connection.path.path_segments[0].token)
2. The TLSInfo (as retrieved by peer.FromContext for Server or grpc.Peer for the Client) which contains the following data: 
	*	"certificate"  --  is  a  pem  encoded  x509cert  (usage:  input.auth_info.certificate)
	*	"spiffe_id"  --  is  a  spiffeID  from  SVIDX509Certificate  (usage:  input.auth_info.spiffe_id)
3. "operation"  --  one  of  request/close  (usage:  input.operation)
4. "role"  --  one  of  client/endpoint  (usage:  input.role)

An example of using OPA input for the case of token signature verification

```
package test 
default allow = false    
allow { 
   token := input.connection.path.path_segments[0].token  
   cert := input.auth_info.certificate  
   io.jwt.verify_es256(token, cert) # signature verification  
}
```

Motivation #200 

Signed-off-by: Dmitry Vlasov <dmitry.vlasov@xored.com>

### DIFF
--- a/pkg/tools/opautils/opainput.go
+++ b/pkg/tools/opautils/opainput.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Cisco and/or its affiliates.
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
--- a/pkg/tools/opautils/opainput.go
+++ b/pkg/tools/opautils/opainput.go
@@ -17,8 +17,8 @@
 // Now we can consider the OPA input as a map with the keys:
 //    1) "connection" -- we can use all information about connection in OPA (e.g. input.connection.path.path_segments[0].token)
 //    2) "auth_info" with sub keys:
-//          -"certificate" -- is a pem encoded x509cert (access: input.auth_info.certificate)
-//			-"spiffe_id" -- is a spiffeID from SVIDX509Certificate (access: input.auth_info.spiffe_id)
+//          i) "certificate" -- is a pem encoded x509cert (access: input.auth_info.certificate)
+//			ii) "spiffe_id" -- is a spiffeID from SVIDX509Certificate (access: input.auth_info.spiffe_id)
 //    3) "operation" -- one of request/close (access: input.operation)
 //    4) "role" -- one of client/endpoint (access: input.role)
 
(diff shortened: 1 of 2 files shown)


edwarnicke:
```
package test 
default allow = false    
allow { 
   token := input.connection.path.path_segments[0].token  
   cert := input.auth_info.certificate  
   io.jwt.verify_es256(token, cert) # signature verification  
}```

Before the line:
```
cert := input.auth_info.certificate
```

We need something to *validate* the cert.

File: pkg/tools/opautils/opainput.go [hunk 1]
```diff
@@ -0,0 +1,125 @@
...
+	return peerCert
+}
+
+func convertConnectionToMap(connection *networkservice.Connection) (map[string]interface{}, error) {
+	jsonConn, err := json.Marshal(connection)
+	if err != nil {
+		return nil, err
+	}
+	rv := make(map[string]interface{})
+	if err := json.Unmarshal(jsonConn, &rv); err != nil {
```
edwarnicke, line 121: I am really curious why the unmarshal here :)

pkg/tools/opautils/opainput.go
```diff
@@ -0,0 +1,125 @@
...
+	pemcert := pemEncodingX509Cert(cert)
+
+	rv := map[string]interface{}{
+		"connection": connectionAsMap,
+		"auth_info": map[string]interface{}{
+			"certificate": pemcert,
+			"spiffe_id":   spiffeID,
+		},
+		"operation": operation,
+		"role":      role,
```
edwarnicke, line 95: Tell me what you are thinking about the "role" parameter :)

pkg/tools/opautils/opainput.go
lazyniv: I thought, that we need to convert all policy input data to a map[string]interface{}

pkg/tools/opautils/opainput.go
lazyniv: So I don't understand what the authorization client chain element should do. Please can you explain some usage cases when we need to use authorization client? Authz server is handle policies by using connection and tls info, but what authz client should do?

pkg/tools/opautils/opainput.go
edwarnicke: Ah, I see :)

pkg/tools/opautils/opainput.go
edwarnicke: Client-side authorization is used to decide if you trust who is providing you the Network Service.

Client-side authorization would also use OPA to do authorization based on the TLSInfo and Connection... but in this case its the Connection *returned* from its Request call.

lazyniv:
@edwarnicke I have implemented OPA input, what do you think about operation and role parameters? In my implementation we pass these parameters to OPA input as strings in authorization chain element. What do you think about of this approach?

lazyniv:
@edwarnicke What do you think about operation and role parameters? In my implementation of OPA input we just pass these parameters to OPA as strings in authorization chain element. What do you think about of this approach?

Issue #196, opened 2020-04-14
Investigate race condition in eventchannel


edwarnicke:
https://github.com/networkservicemesh/sdk/pull/195/checks?check_run_id=584264737

denis-tingajkin:
Should be fixed by https://github.com/networkservicemesh/sdk/pull/226

PR #226, opened 2020-04-26
Fix race condition in eventchannel

denis-tingajkin:
## Motivation

https://github.com/networkservicemesh/sdk/issues/196

### DIFF
--- a/pkg/networkservice/core/eventchannel/monitor_connection_server.go
+++ b/pkg/networkservice/core/eventchannel/monitor_connection_server.go
@@ -18,6 +18,7 @@ package eventchannel
 
 import (
 	"context"
+
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc/metadata"
--- a/pkg/networkservice/core/eventchannel/monitor_connection_server.go
+++ b/pkg/networkservice/core/eventchannel/monitor_connection_server.go
@@ -18,8 +18,6 @@ package eventchannel
 
 import (
 	"context"
-	"sync"
-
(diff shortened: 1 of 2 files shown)


pkg/networkservice/core/eventchannel/monitor_connection_server.go
```diff
@@ -48,9 +49,15 @@ func NewMonitorConnectionMonitorConnectionsServer(ctx context.Context, eventCh c
 }
 
 func (m *monitorConnectionMonitorConnectionsServer) Send(event *networkservice.ConnectionEvent) error {
+	m.m.Lock()
+	defer m.m.Unlock()
+	if m.done {
+		return errors.New("monitor connection server is done")
+	}
 	select {
```
haiodo, line 57: I do not think it is a good idea to have lock hold in select.

pkg/networkservice/core/eventchannel/monitor_connection_server.go
```diff
@@ -39,23 +37,15 @@ func NewMonitorConnectionMonitorConnectionsServer(ctx context.Context, eventCh c
...
-	go func() {
-		<-ctx.Done()
-		// Funnel through Send here to trigger close in *one* place to avoid race conditions
-		_ = rv.Send(&networkservice.ConnectionEvent{})
-	}()
 	return rv
 }
 
 func (m *monitorConnectionMonitorConnectionsServer) Send(event *networkservice.ConnectionEvent) error {
-	select {
```
denis-tingajkin, line 51: The root cause is this `select`.  Actually two goroutines can read and close `eventCh` in one moment.

edwarnicke: Elegant solution :)

Issue #185, opened 2020-04-10
Create DNSContext NetworkServiceClient chain element


denis-tingajkin:
## Description

Previously we had added DNSContext server chain element: https://github.com/networkservicemesh/sdk/issues/74

We need also to create in pkg/networkservice/connectioncontext/dnscontext/ a NetworkServiceClient chain element appropriate for getting a DNSContext for clients.

This part was blocked by https://github.com/networkservicemesh/sdk/issues/48

PR #227, opened 2020-04-28
Create DNSContext NetworkServiceClient chain element

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/185

### DIFF
--- a/pkg/networkservice/connectioncontext/dnscontext/client.go
+++ b/pkg/networkservice/connectioncontext/dnscontext/client.go
@@ -32,33 +32,39 @@ import (
 )
 
 type dnsContextClient struct {
-	cancelMonitoring context.CancelFunc
-	chainContext     context.Context
-	monitorContext   context.Context
-	coreFilePath     string
-	getCallOptions   func() []grpc.CallOption
-	dnsConfigManager dnscontext.Manager
-	monitorClient    networkservice.MonitorConnectionClient
-	executor         serialize.Executor
+	cancelMonitoring    context.CancelFunc
+	chainContext        context.Context
+	monitorContext      context.Context
+	coreFilePath        string
+	resolveConfigPath   string
+	defaultNameServerIP string
+	monitorCallOptions  []grpc.CallOption
+	dnsConfigManager    dnscontext.Manager
+	monitorClient       networkservice.MonitorConnectionClient
+	executor            serialize.Executor
 }
 
 // NewClient creates a new DNS client chain component. Setups all DNS traffic to the localhost. Monitors DNS configs from connections.
-func NewClient(chainContext context.Context, coreFilePath, resolveConfigPath string, monitorClient networkservice.MonitorConnectionClient, getMonitorCallOptions func() []grpc.CallOption) networkservice.NetworkServiceClient {
+func NewClient(monitorClient networkservice.MonitorConnectionClient, options ...DNSOption) networkservice.NetworkServiceClient {
 	c := &dnsContextClient{
-		coreFilePath:     coreFilePath,
-		chainContext:     chainContext,
-		dnsConfigManager: dnscontext.NewManager(),
-		monitorClient:    monitorClient,
-		getCallOptions:   getMonitorCallOptions,
+		chainContext:        context.Background(),
+		dnsConfigManager:    dnscontext.NewManager(),
+		monitorClient:       monitorClient,
+		defaultNameServerIP: "127.0.0.1",
+		resolveConfigPath:   "/etc/resolv.conf",
+		coreFilePath:        "/etc/coredns/Corefile",
 	}
-	if r, err := dnscontext.OpenResolveConfig(resolveConfigPath); err != nil {
-		logrus.Errorf("DnsContextClient: can not load resolve config file. Path: %v. Error: %v", resolveConfigPath, err.Error())
+	for _, o := range options {
+		o.apply(c)
+	}
+	if r, err := dnscontext.OpenResolveConfig(c.resolveConfigPath); err != nil {
+		logrus.Errorf("DnsContextClient: can not load resolve config file. Path: %v. Error: %v", c.resolveConfigPath, err.Error())
 	} else {
 		c.dnsConfigManager.Store("", &networkservice.DNSConfig{
 			SearchDomains: r.Value(dnscontext.AnyDomain),
 			DnsServerIps:  r.Value(dnscontext.NameserverProperty),
 		})
-		r.SetValue(dnscontext.NameserverProperty, "127.0.0.1")
+		r.SetValue(dnscontext.NameserverProperty, c.defaultNameServerIP)
 		if err := r.Save(); err != nil {
 			logrus.Errorf("DnsContextClient: can not update resolve config file. Error: %v", err.Error())
 		}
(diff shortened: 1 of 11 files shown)


denis-tingajkin:
Here failed two test jobs on windows and osx. PR is not ready for review but be free for comment.

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -0,0 +1,126 @@
...
+	monitorContext   context.Context
+	coreFilePath     string
+	getCallOptions   func() []grpc.CallOption
+	dnsConfigManager dnscontext.Manager
+	monitorClient    networkservice.MonitorConnectionClient
+	executor         serialize.Executor
+}
+
+// NewClient creates a new DNS client chain component. Setups all DNS traffic to the localhost. Monitors DNS configs from connections.
+func NewClient(chainContext context.Context, coreFilePath, resolveConfigPath string, monitorClient networkservice.MonitorConnectionClient, getMonitorCallOptions func() []grpc.CallOption) networkservice.NetworkServiceClient {
```
edwarnicke, line 46: This is a lot of parameters... are there good presumable defaults for many of them?  If so, it might make sense to 
think about something like the WithXYZ pattern we see in grpc.  The goal is to preserve the excellent degree of configuratibility we have, but also make default use simple.  Thoughts?

pkg/tools/dnscontext/manager.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+)
+
+// Manager provides API for storing/deleting dnsConfigs. Can represent the configs in caddyfile format.
+// Can be used from different goroutines
+type Manager interface {
```
edwarnicke, line 29: Is there a reason we are pulling this out into a separate thing?

pkg/networkservice/connectioncontext/dnscontext/client.go
denis-tingajkin: @edwarnicke 

> Thoughts?

Here we are passing `chainContext context.Context`. Probably we can just look up into context values to use specific configuration. For example, if I need to use specific configuration then I can do something like this:
```go
dnscontext.NewClient(dnscontext.WithCorefilePath(ctx, "mydir/Corefile"), monitorClient) 
```
Or If I want to use default values I could just pass context without values
```go
dnscontext.NewClient(ctx, monitorClient) 
```
Do you mind with this pattern?

pkg/tools/dnscontext/manager.go
denis-tingajkin: > Is there a reason we are pulling this out into a separate thing?

Yes, there I extract DNS tools to make the structure of `connectioncontext/dnscontext` pkg more clear. For example, pkg `connectioncontext/dnscontext` has only server, client and test files, and all depend on tools located in pkg `tools/...`.

pkg/networkservice/connectioncontext/dnscontext/client.go
haiodo: Please use `dnscontext.NewClient(ctx, monitorClient, DnsOption ...options)` I think it would be good.

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -0,0 +1,126 @@
...
+		getCallOptions:   getMonitorCallOptions,
+	}
+	if r, err := dnscontext.OpenResolveConfig(resolveConfigPath); err != nil {
+		logrus.Errorf("DnsContextClient: can not load resolve config file. Path: %v. Error: %v", resolveConfigPath, err.Error())
+	} else {
+		c.dnsConfigManager.Store("", &networkservice.DNSConfig{
+			SearchDomains: r.Value(dnscontext.AnyDomain),
+			DnsServerIps:  r.Value(dnscontext.NameserverProperty),
+		})
+		r.SetValue(dnscontext.NameserverProperty, "127.0.0.1")
```
haiodo, line 61: Should we support IPv6 here as well, or it is not required?

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	return c
+}
+
+func (c *dnsContextClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+	rv, err := next.Client(ctx).Request(ctx, request, opts...)
+	if err != nil {
+		return nil, err
+	}
+	c.monitorContext, c.cancelMonitoring = context.WithCancel(c.chainContext)
```
haiodo, line 74: Should it be inside <- c.executor.AsyncExec?

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -0,0 +1,126 @@
...
+		c.executor.AsyncExec(c.monitorConfigs)
+		return
+	}
+	for {
+		if c.monitorContext.Err() != nil {
+			return
+		}
+		event, err := steam.Recv()
+		if err != nil {
+			c.executor.AsyncExec(c.monitorConfigs)
```
haiodo, line 100: I think at least logging of err are required.

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (c *dnsContextClient) handleEvent(event *networkservice.ConnectionEvent) {
+	switch event.GetType() {
+	case networkservice.ConnectionEventType_INITIAL_STATE_TRANSFER:
+		for k, v := range event.Connections {
+			c.dnsConfigManager.Store(k, v.GetContext().GetDnsContext().GetConfigs()...)
+		}
+	case networkservice.ConnectionEventType_UPDATE:
+		for k, v := range event.Connections {
+			c.dnsConfigManager.Remove(k)
```
haiodo, line 118: Did a Remove are required, or did Store not overwrite value?

denis-tingajkin: Remove did not require. Fixed.

pkg/networkservice/connectioncontext/dnscontext/client.go
denis-tingajkin: Thanks, fixed.

pkg/networkservice/connectioncontext/dnscontext/client.go
denis-tingajkin: Good catch, fixed.

pkg/networkservice/connectioncontext/dnscontext/client.go
denis-tingajkin: Thanks, fixed.

pkg/networkservice/connectioncontext/dnscontext/client.go
denis-tingajkin: @haiodo , @edwarnicke got it. Thanks, fixed.

PR #233, opened 2020-05-04
Allow to pass list of URLs via env and command line

haiodo:
Allow to pass list of URLs via env and command line


pkg/tools/flags/url_slice_value.go
```diff
@@ -0,0 +1,187 @@
...
+func NewURLSliceValue(p *[]*url.URL) pflag.Value {
+	v := &urlSliceValue{
+		values:  p,
+		changed: false,
+	}
+	*p = []*url.URL{}
+	return v
+}
+
+func readAsCSV(val string) ([]*url.URL, error) {
```
edwarnicke, line 50: Interesting... by reading as csv you basically make both the apparent unitary case:
unix://foo
And the multiple case
unix://foo,10.10.10.1:9111 valid.  Correct?

haiodo: Yes.

haiodo:
@edwarnicke PR depend on fix in https://github.com/networkservicemesh/sdk/pull/234 to prevent concurrency from happening. Could try to return tests to pass.

haiodo:
@fkautz could you please check, thanks.

PR #238, opened 2020-05-06
Moved grpcutils to a saner errCh approach rather than playing multicontext monty

edwarnicke:


pkg/tools/grpcutils/listen_and_serve.go
```diff
@@ -24,28 +24,25 @@ import (
...
+	errCh := make(chan error, 1)
 
 	// Serve
 	go func() {
 		// Create listener
 		network, target := urlToNetworkTarget(address)
 		ln, err := net.Listen(network, target)
 		if err != nil {
-			log.Entry(ctx).Fatalf("failed to listen on %q: %v", address, err)
+			log.Entry(ctx).Errorf("failed to listen on %q: %v", address, err)
```
denis-tingajkin, line 45: Should we pass error to the errCh here?
```suggestion
			log.Entry(ctx).Errorf("failed to listen on %q: %v", address, err)
			errCh <- err
			return
```

pkg/tools/grpcutils/listen_and_serve.go
```diff
@@ -24,28 +24,25 @@ import (
...
 
-// ListenAndServe listens on address with server.  Returns a context which will be canceled in the event that
-// server.Serve(listener) returns an error.  The resulting error can then be retrieve from the returned context with.
-// grpcutils.Err(ctx)
-func ListenAndServe(ctx context.Context, address *url.URL, server *grpc.Server) context.Context {
-	ctx, cancel := context.WithCancel(ctx)
-	ctx = errctx.WithErr(ctx)
+// ListenAndServe listens on address with server.  Returns an chan err  which will
+// receive an error and then be closed in the event that server.Serve(listener) returns an error.
+func ListenAndServe(ctx context.Context, address *url.URL, server *grpc.Server) chan error {
```
denis-tingajkin, line 36: Let's return read-only error channel to prevent closing by the consumer.

pkg/tools/grpcutils/listen_and_serve.go
edwarnicke: Done! :)

pkg/tools/grpcutils/listen_and_serve.go
edwarnicke: Done! :)

Issue #232, opened 2020-05-01
Add registry memory chain components


denis-tingajkin:
## Motivation

The main idea to create chain components to be able to build analogies registries like `nsmd-k8s` via SDK and be not dependent on Kubernetes staff.

Memory registries could be useful for the next goals:
1. Testing cmd application which depends on `nsmd-k8s`. For example `cmd-nsmgr`.
2. Building **nsm registries** application which not depend on Kubernetes.

PR #240, opened 2020-05-06
Prepare pkg registry/core/next for registry memory chain elements

denis-tingajkin:

## Motivation

1. Fixed a couple of issues related to registry next pkg
2. Added support for `NsmRegistryClient`/`NsmRegistryServer`
3. https://github.com/networkservicemesh/sdk/issues/232


pkg/registry/core/next/context.go
```diff
@@ -1,4 +1,6 @@
-// Copyright (c) 2020 Cisco Systems, Inc.
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+//Copyright (c) 2020 Cisco Systems, Inc.
```
haiodo, line 3: Indenting, Also I think it should start with Cisco, since it was first.

pkg/registry/core/next/discovery_client.go
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
```
haiodo, line 2: Same here.

pkg/registry/core/next/context.go
denis-tingajkin: Thanks, fixed

pkg/registry/core/next/discovery_client.go
denis-tingajkin: Thanks, fixed

denis-tingajkin:
@edwarnicke, @haiodo Here I faced with a lot of duplicate changes... 
**Question:** Can we implement a simple go code generator for `next` pattern to avoid problems with `next` pkgs code support? 

For example, the generator on each passed type will generate a source file with `next` logic for all type methods.  
gen.go file can looks someting like this:
```go
// go:generate go run pkg/core/next/generator.go -type=registry.NsmRegistryServer -output=pkg/registry/core/next 
```
What do you think? Please share your thoughts.

denis-tingajkin:
Question https://github.com/networkservicemesh/sdk/pull/240#issuecomment-624756702 still actual. @fkautz Do you mind if we will use go code generation for pkg `next`?

edwarnicke:
> @edwarnicke, @haiodo Here I faced with a lot of duplicate changes...
> **Question:** Can we implement a simple go code generator for `next` pattern to avoid problems with `next` pkgs code support?
> 
> For example, the generator on each passed type will generate a source file with `next` logic for all type methods.
> `gen.go` file can look something like this:
> 
> ```go
> // go:generate ${GEN TOOL} -type=registry.NsmRegistryServer -output=pkg/registry/core/next 
> ```
> 
> What do you think? Please share your thoughts.

That could be very interesting... what are your thoughts on the particulars?

denis-tingajkin:
@edwarnicke Here I see two options on how to implement it.

Let's consider the current structure of pkg next:

```
- next<T> .go // implements interface T and have logic for select next chain element.
- tail<T>.go // returns result if previous chain elements did not return the result.
- context<T>.go // has context specific functions like Get<T>(context.Context)...
```

Where T is an interface for chainings. 

## Solutions via code generation:

### 1. Use exist go code generators

This option will allow us to generate only common parts (I.e. we will need to implement `tail<T>` and `next<T>` parts). Because the generator tool is not know how it should generate the implementation of a specific interface method.
#### Steps to add type for pkg/next
1. Generate a common part for type T.
2. Implement `tail<T>`
3. Implement `next<T>`

### 2. Use the custom tool for generating

We can add pkg for our `sdk/tools` directory. 
For example `sdk/tools/nextgen`.
And implement a simple code generating tool. 

#### Tool input: 
1. Type for generating.
2. Output path.

#### Tool output:
1. generated go code file.

#### How it will work?

1. For finding all of the public methods we can use pkg `reflect`.
2. For generating `next<T>` part we will use the pattern based on https://github.com/networkservicemesh/sdk/blob/fd5d675dedcd1ab455c609f496c8b03ca5e20764/pkg/networkservice/core/next/client.go#L61

3. For `tail<T>` part the tool could use the next strategy:
3.1. If the output params of generating method have one of the input params then return the input parameter as a result.
3.2. If input and output parameters of the generating method don't match then return nill for all output parameters.
  3.2.1 If one of the output parameters is an error then return the error 'T not found'.

#### Tool usage example

```
 go:generate sdk/tools/nextgen/generator.go -type=registry.NsmRegistryServer -output=pkg/registry/core/next 
```

PR #244, opened 2020-05-14
WIP: Add join registry chain elements

denis-tingajkin:

## Motivation

Registry `join` chain elements add a possible to aggregate a few registry clients to one. It could be useful for `cmd-nsmgr`.

Depends on https://github.com/networkservicemesh/sdk/pull/241


denis-tingajkin:
This PR is blocked, reason: I think at first we can simplify registry API https://github.com/networkservicemesh/api/issues/39 to reduce code lines of this patch.

Issue #237, opened 2020-05-05
Sometimes unit test 'TestNewClient_StopRefreshAtClose' has goroutine leak


denis-tingaikin:
## Context

https://github.com/networkservicemesh/sdk/pull/227/checks?check_run_id=645773960

## Failure logs

```
##[error]    leaks.go:78: found unexpected goroutines:
        [Goroutine 62 in state chan send, with github.com/networkservicemesh/sdk/pkg/networkservice/common/refresh_test.TestNewClient_StopRefreshAtClose.func1 on top of the stack:
        goroutine 62 [chan send]:
        github.com/networkservicemesh/sdk/pkg/networkservice/common/refresh_test.TestNewClient_StopRefreshAtClose.func1(0x18d8e20, 0xc000154120, 0xc000180280, 0x0, 0x0, 0x0, 0x1180c0b, 0xc000154140, 0xc000154120)
        	/Users/runner/runners/2.169.1/work/sdk/sdk/pkg/networkservice/common/refresh/client_test.go:97 +0x78
        github.com/networkservicemesh/sdk/pkg/networkservice/common/refresh_test.(*testRefresh).Request(0xc000010280, 0x18d8e20, 0xc000154120, 0xc000180280, 0x0, 0x0, 0x0, 0xc000154120, 0x18d21a0, 0xc000108b70)
        	/Users/runner/runners/2.169.1/work/sdk/sdk/pkg/networkservice/common/refresh/client_test.go:67 +0x96
        github.com/networkservicemesh/sdk/pkg/networkservice/core/next.(*nextClient).Request(0xc000108b70, 0x18d8f60, 0xc000152100, 0xc000180280, 0x0, 0x0, 0x0, 0x106989c, 0xc000152100, 0xc000048c00)
        	/Users/runner/runners/2.169.1/work/sdk/sdk/pkg/networkservice/core/next/client.go:70 +0x5fb
        github.com/networkservicemesh/sdk/pkg/networkservice/common/refresh.(*refreshClient).Request(0xc00000ee20, 0x18d8f60, 0xc000152100, 0xc000180280, 0x0, 0x0, 0x0, 0xc0001c8218, 0x1092790, 0xc0001c81c0)
        	/Users/runner/runners/2.169.1/work/sdk/sdk/pkg/networkservice/common/refresh/client.go:53 +0xaa
        github.com/networkservicemesh/sdk/pkg/networkservice/common/refresh.(*refreshClient).createTimer.func1()
[...]

PR #248, opened 2020-05-22
Fix data races in RefreshClient

lvfxx:
Implementation of #237 
Fixed data race between refresh goroutines and close calls/new requests from heal. In some situations there wasn't enough to call `timer.Stop()` to really stop refresh goroutine. I've added context for each refresh and moved timer creation to synced block to fix it. 
Also fixed invalid second request initialization in `TestNewClient_StopRefreshAtAnotherRequest`, that also led to data race. 
Harmless refactoring.

### DIFF
--- a/pkg/networkservice/common/refresh/client.go
+++ b/pkg/networkservice/common/refresh/client.go
@@ -35,16 +35,18 @@ import (
 )
 
 type refreshClient struct {
-	connectionTimers map[string]*time.Timer
-	executor         serialize.Executor
+	connectionTimers       map[string]*time.Timer
+	timerContextCancellers map[string]func()
+	executor               serialize.Executor
 }
 
 // NewClient - creates new NetworkServiceClient chain element for refreshing connections before they timeout at the
 // endpoint
 func NewClient() networkservice.NetworkServiceClient {
 	rv := &refreshClient{
-		connectionTimers: make(map[string]*time.Timer),
-		executor:         serialize.Executor{},
+		connectionTimers:       make(map[string]*time.Timer),
+		timerContextCancellers: make(map[string]func()),
+		executor:               serialize.Executor{},
 	}
 	return rv
 }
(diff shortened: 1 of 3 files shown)


pkg/networkservice/common/refresh/client.go
```diff
@@ -35,16 +35,18 @@ import (
...
 
 // NewClient - creates new NetworkServiceClient chain element for refreshing connections before they timeout at the
 // endpoint
 func NewClient() networkservice.NetworkServiceClient {
 	rv := &refreshClient{
-		connectionTimers: make(map[string]*time.Timer),
-		executor:         serialize.Executor{},
+		connectionTimers:       make(map[string]*time.Timer),
+		timerContextCancellers: make(map[string]func()),
+		executor:               serialize.Executor{},
```
denis-tingajkin, line 49: We do not need to assign value for field `executor`.

pkg/networkservice/common/refresh/client.go
```diff
@@ -59,15 +61,24 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
...
 	req.Connection = rv
 	// Setup the timer with the req containing the returned connection
-	ct, err := t.createTimer(ctx, req, opts...)
+
+	expire, err := t.getExpireDuration(request)
 	if err != nil {
 		return nil, errors.Wrapf(err, "Error creating timer from Request.Connection.Path.PathSegment[%d].ExpireTime", request.GetConnection().GetPath().GetIndex())
 	}
 	t.executor.AsyncExec(func() {
+		if ctx.Err() != nil {
```
denis-tingajkin, line 70: Why do we need to check request context here?

lvfxx: For each scheduled refresh request (created in `createTimer`) we set cancel function and save it to the map. When a new request from heal or stop call with the same `connId` will reach `AsyncExec` block, it cancels refresh's context. Then when refresh will try to schedule its next call, it sees that context was cancelled and terminates itself. This check must be placed in `AsyncExec` block to ensure that new request and refresh goroutines will not get access to context concurrently.

pkg/networkservice/common/refresh/client.go
lvfxx: I'll remove it, thanks.

pkg/networkservice/common/refresh/client.go
```diff
@@ -78,23 +89,31 @@ func (t *refreshClient) Close(ctx context.Context, conn *networkservice.Connecti
 			timer.Stop()
 			delete(t.connectionTimers, conn.GetId())
 		}
+		if canceller, ok := t.timerContextCancellers[conn.GetId()]; ok {
+			canceller()
+			delete(t.timerContextCancellers, conn.GetId())
```
denis-tingajkin, line 94: Should we also delete the entry from the map connectionTimers?

pkg/networkservice/common/refresh/client.go
denis-tingajkin: As I can see here using the request context. Should to use a special `refresh context` for refresh goals?
For example you need to create a refreshing context on first Request call and reuse this context in the refresh timer. Is it make sense?

pkg/networkservice/common/refresh/client.go
lvfxx: It is deleted two lines above

pkg/networkservice/common/refresh/client.go
```diff
@@ -59,25 +59,37 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
...
-			canceller()
+		// check if it is refresh request
+		if refreshCtx := refreshContext(ctx); refreshCtx != nil {
+			// refresh was canceled
+			if refreshCtx.Err() != nil {
+				return
+			}
+			// we reuse non-canceled refresh context for the next refresh request
+			timer := t.createTimer(ctx, req, expire, opts...)
+			t.connectionTimers[req.GetConnection().GetId()] = timer
```
denis-tingajkin, line 76: [minor] `req.GetConnection().GetId()` reused many times, you can create local variable for it. 
```suggestion
			id := req.GetConnection().GetId()
			t.connectionTimers[id] = timer
```

PR #252, opened 2020-05-25
Add endpoint url for identification

haiodo:
Fix localbypass to use endpoint url for identification.

Depenes on:
https://github.com/networkservicemesh/api/pull/40


edwarnicke:
@haiodo Should work after a rebase :)

denis-tingajkin:
As I understand `URL` field will be used for callback API in case of NSE. 

@haiodo  Should we also create a new SDK chain NetworkServiceServer which will fill field `URL`?

PR #255, opened 2020-05-26
Remove unused 'role' and 'operation' fields from OPA input

denis-tingajkin:


denis-tingajkin:
I'm not sure that we can correctly check these fields at this moment. Probably we can return this later.

PR #256, opened 2020-05-26
NSM manager registry chain + memory reg fixes.

haiodo:
+ Chain element for NSMgr registry chain
+ Fixes for bulk register tail and stuff.

### DIFF
--- a/pkg/networkservice/common/localbypass/server_test.go
+++ b/pkg/networkservice/common/localbypass/server_test.go
@@ -23,7 +23,7 @@ import (
 	"net/url"
 	"testing"
 
-	localbypass_reg "github.com/networkservicemesh/sdk/pkg/registry/common/localbypass"
+	registrylocalbypass "github.com/networkservicemesh/sdk/pkg/registry/common/localbypass"
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
@@ -76,7 +76,7 @@ func (s testNetworkServiceServer) Close(ctx context.Context, _ *networkservice.C
 
 func TestNewServer_NSENotPresented(t *testing.T) {
 	defer goleak.VerifyNone(t)
-	localBypassRegistryServer := localbypass_reg.NewServer()
+	localBypassRegistryServer := registrylocalbypass.NewServer()
 	localBypassNetworkServiceServer := localbypass.NewServer(localBypassRegistryServer)
 	server := next.NewNetworkServiceServer(localBypassNetworkServiceServer, &testNetworkServiceServer{})
 	request := &networkservice.NetworkServiceRequest{
@@ -98,7 +98,7 @@ func TestNewServer_NSENotPresented(t *testing.T) {
 
 func TestNewServer_UnixAddressRegistered(t *testing.T) {
 	defer goleak.VerifyNone(t)
-	localBypassRegistryServer := localbypass_reg.NewServer()
+	localBypassRegistryServer := registrylocalbypass.NewServer()
 	localBypassNetworkServiceServer := localbypass.NewServer(localBypassRegistryServer)
 	server := next.NewNetworkServiceServer(localBypassNetworkServiceServer, &testNetworkServiceServer{})
 	request := &networkservice.NetworkServiceRequest{
@@ -129,7 +129,7 @@ func TestNewServer_UnixAddressRegistered(t *testing.T) {
 
 func TestNewServer_NonUnixAddressRegistered(t *testing.T) {
 	defer goleak.VerifyNone(t)
-	localBypassRegistryServer := localbypass_reg.NewServer()
+	localBypassRegistryServer := registrylocalbypass.NewServer()
 	localBypassNetworkServiceServer := localbypass.NewServer(localBypassRegistryServer)
 	server := next.NewNetworkServiceServer(localBypassNetworkServiceServer, &testNetworkServiceServer{})
 	request := &networkservice.NetworkServiceRequest{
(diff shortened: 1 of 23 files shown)


pkg/registry/nsmgr/server.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/networkservicemesh/api/pkg/api/registry"
+
+	"github.com/networkservicemesh/sdk/pkg/registry/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/serialize"
+)
+
+type nsmRegistryServer struct {
+	registry.NetworkServiceRegistryServer
```
denis-tingajkin, line 31: Why do we need this field?

pkg/registry/nsmgr/server.go
```diff
@@ -0,0 +1,100 @@
...
+	serialize.Executor
+	manager *registry.NetworkServiceManager
+}
+
+// NewServer - construct a new NSM aware NSE registry server
+// Server is using latest NSM registration and update endpoints accordingly.
+func NewServer(manager *registry.NetworkServiceManager) registry.NetworkServiceRegistryServer {
+	return &nsmRegistryServer{
+		manager: manager,
```
denis-tingajkin, line 40: Do we allow the nil manager here?

pkg/registry/memory/nse_memory_server.go
```diff
@@ -32,6 +32,15 @@ type memoryNetworkServeRegistry struct {
...
 func (m *memoryNetworkServeRegistry) RegisterNSE(ctx context.Context, registration *registry.NSERegistration) (*registry.NSERegistration, error) {
+	nseRegistration, err := m.processRegistration(registration)
+	if err != nil {
+		return nseRegistration, err
+	}
+
+	return next.NetworkServiceRegistryServer(ctx).RegisterNSE(ctx, registration)
+}
+
+func (m *memoryNetworkServeRegistry) processRegistration(registration *registry.NSERegistration) (*registry.NSERegistration, error) {
```
denis-tingajkin, line 43: Do we really need this method? Can we in bulk server call `m.RegisterNSE(...)` ?

pkg/registry/nsmgr/server.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+// RegisterNSE - register NSE with current manager associated.
+func (n *nsmRegistryServer) RegisterNSE(ctx context.Context, registration *registry.NSERegistration) (*registry.NSERegistration, error) {
+	// update manager to be proper one.
+	<-n.AsyncExec(func() {
+		registration.NetworkServiceManager = n.manager
+	})
```
denis-tingajkin, line 49: Why do we need to use synchronization?

pkg/registry/nsmgr/server.go
```diff
@@ -0,0 +1,100 @@
...
+	result, err := next.NetworkServiceRegistryServer(ctx).RegisterNSE(ctx, registration)
+	if result != nil {
```
denis-tingajkin, line 51: Should we check `err` here?

pkg/registry/nsmgr/server.go
haiodo: removed.

pkg/registry/nsmgr/server.go
haiodo: NewServer is not return error, so I could not return error here. But I think manager should be passed.

pkg/registry/nsmgr/server.go
haiodo: Yes, we don't. Removed.

pkg/registry/memory/nse_memory_server.go
haiodo: We can't since it will loose stream capability.

pkg/registry/nsmgr/server.go
haiodo: thanks.

pkg/registry/common/localbypass/server.go
```diff
@@ -21,42 +21,56 @@ package localbypass
...
 	"github.com/networkservicemesh/api/pkg/api/registry"
-
-	"github.com/networkservicemesh/sdk/pkg/registry/core/next"
 )
 
-type localBypassRegistry struct {
-	sockets localbypass.SocketMap
+// LocalBypass - return interafce to return localbypass endpoints URIs
+type LocalBypass interface {
+	LoadEndpoint(endpointName string) (*url.URL, bool)
```
denis-tingajkin, line 33: [minor] Should we rename it to `LoadPass` ?

pkg/registry/core/next/discovery_server.go
```diff
@@ -40,7 +40,9 @@ type nextDiscoveryServer struct {
 func NewWrappedDiscoveryServer(wrapper DiscoveryServerWrapper, servers ...registry.NetworkServiceDiscoveryServer) registry.NetworkServiceDiscoveryServer {
 	rv := &nextDiscoveryServer{servers: make([]registry.NetworkServiceDiscoveryServer, 0, len(servers))}
 	for _, c := range servers {
-		rv.servers = append(rv.servers, wrapper(c))
+		if c != nil {
```
denis-tingajkin, line 43: Do we ignore with this check that SDK developer doing something wrong?

pkg/networkservice/common/localbypass/server_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"net/url"
 	"testing"
 
+	localbypass_reg "github.com/networkservicemesh/sdk/pkg/registry/common/localbypass"
```
denis-tingajkin, line 26: I would suggest this name
```suggestion
	registrylocalbypass "github.com/networkservicemesh/sdk/pkg/registry/common/localbypass"
```

pkg/registry/core/next/discovery_server.go
haiodo: There was no such check, but it simplify usage of this kind of wrapper servers a lot.

pkg/networkservice/common/localbypass/server_test.go
haiodo: Thanks.

pkg/registry/common/localbypass/server.go
haiodo: Thanks.

edwarnicke:
Can this be broken into smaller pieces?  There seem to be several things in motion here.

Issue #201, opened 2020-04-14
Default Policy examples


edwarnicke:
We need default OPA Policies based upon the update in input found in #200

Among the examples we'd want:

1. **Tokens valid** - checks the validity of some or all of the tokens in the path.
1. **No Tokens in path expired** - none of the tokens in the chain are expired.
1. **Last Token matches TLSInfo**  - checks to make sure that the Last Token (JWT) is signed by the same cert as found in TLSInfo.
1. **Token.Aud for token n in path matches Token.Sub for n+1 token in chain**

lazyniv:
I think, we can consider the following use case example of tokens validation(1)

So, for example of tokens validation we can consider the case of matching clients and endpoints by their SPIFFE IDs. The map of matching(endpoint’s spiffe id -> list of allowed clients by spiffe id) can be stored on NSMgrs. The policy of matching also can be stored on NSMgrs. So let’s consider the case more detailed:

1) Client generates jwt token with the following claims:
	1) src spiffe id -- client’s spiffe id
	2) dst spiffe id -- spiffe id of the endpoint the client wants

2) Then this token should be sign by provided x.509 certificate

3) Then this token should be validated on NSMgr(which placed on the same node as a client) by provided policy using OPA(in our case it’s just a matching by spiffe ids) 

4) Then we get the policy decision from the OPA and do something with this request (e.g. reject)

edwarnicke:
> So, for example of tokens validation we can consider the case of matching clients and endpoints by their SPIFFE IDs. The map of matching(endpoint’s spiffe id -> list of allowed clients by spiffe id) can be stored on NSMgrs. The policy of matching also can be stored on NSMgrs. So let’s consider the case more detailed:

No... an Endpoint *itself* applies its own policy.  The NSMgr is not responsible for that policy.  The NSMgr may have certain kinds of policy applicable to its own role... but client/endpoint permissiblity isn't among them.

>1. Client generates jwt token with the following claims:
>
>src spiffe id -- client’s spiffe id
>dst spiffe id -- spiffe id of the endpoint the client wants
>
[...]

Issue #51, opened 2020-01-21
OPA should support multiple tokens in the chain


fkautz:
OPA currently only reasons on a single token in the chain. It will be useful to have OPA support multiple tokens so that it can reason about elements deeper in the chain.

Context: https://github.com/networkservicemesh/sdk/pull/40#discussion_r369232544

denis-tingaikin:
@fkautz Currently we have a few policies that are using multi tokens :

1. https://github.com/networkservicemesh/sdk/blob/main/pkg/tools/opa/policies/tokens_chained.rego
2. https://github.com/networkservicemesh/sdk/blob/main/pkg/tools/opa/policies/tokens_expired.rego

Let me know if you have something other in mind and then we can reopen this issue :)

PR #257, opened 2020-05-27
OPA: Add tokens expired policy

denis-tingajkin:

- [X] **Depends on** https://github.com/networkservicemesh/sdk/pull/259

## Motivation

https://github.com/networkservicemesh/sdk/issues/51
https://github.com/networkservicemesh/sdk/issues/201 // This PR implements **2. No Tokens in path expired - none of the tokens in the chain are expired.**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -172,6 +172,10 @@ issues:
       linters:
         - funlen
       text: "Function 'TestMonitor' is too long"
+    - path: pkg/tools/opa/tokens_expired_policy.go
+      linters:
+        - gosec
+      text: "G101: Potential hardcoded credentials"
     - path: pkg/networkservice/utils/checks/checkerror/server_test.go
       linters:
         - dupl
--- a/pkg/tools/opa/tokens_expired_policy.go
+++ b/pkg/tools/opa/tokens_expired_policy.go
@@ -15,3 +15,28 @@
 // limitations under the License.
 
 package opa
+
+const tokenExpiredPolicy = `
+package policies
+
+default tokens_expired = false
+
+tokens_expired {
+    token := input.path_segments[_].token
+    [_, payload, _] := io.jwt.decode(token)
+    now > payload.exp
+}
+
+now = t {
+	 ns := time.now_ns()
+     t := ns / 1e9
+}`
+
+// WithTokensExpiredPolicy returns default policy for checking tokens expiration
+func WithTokensExpiredPolicy() AuthorizationPolicy {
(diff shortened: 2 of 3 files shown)


pkg/networkservice/common/authorize/server.go
```diff
@@ -33,63 +30,35 @@ import (
 )
 
 type authorizeServer struct {
-	p *rego.PreparedEvalQuery
+	policies []opa.AuthorizationPolicy
 }
 
 // NewServer - returns a new authorization networkservicemesh.NetworkServiceServers
-func NewServer(p *rego.PreparedEvalQuery) networkservice.NetworkServiceServer {
+func NewServer(policies ...opa.AuthorizationPolicy) networkservice.NetworkServiceServer {
```
denis-tingajkin, line 37: Here I suggest using a set of policies when can be used default policies such as `tokens expired policy`, `tokens valid`. 
Example: 
```go
s := next.NewServer(authorize.NewServer(opa.WithTokensExpiredPolicy(), opa.WithSomePolicy()...))
```
@haiodo , @edwarnicke , @fkautz  What do you think? Please share your thoughts.

edwarnicke: I like where this is going *a lot*.

When you bubble this up to the endpoint and client chains.. you could basically then let folks pass in an 'authzServer' or 'authzClient' into the constructors for those chains :)

pkg/networkservice/chains/endpoint/server.go
```diff
@@ -55,12 +56,12 @@ type endpoint struct {
 //             - name - name of the NetworkServiceServer
 //             - tokenGenerator - token.GeneratorFunc - generates tokens for use in Path
 //             - additionalFunctionality - any additional NetworkServiceServer chain elements to be included in the chain
-func NewServer(name string, authzPolicy *rego.PreparedEvalQuery, tokenGenerator token.GeneratorFunc, additionalFunctionality ...networkservice.NetworkServiceServer) Endpoint {
+func NewServer(name string, authzPolicies []opa.AuthorizationPolicy, tokenGenerator token.GeneratorFunc, additionalFunctionality ...networkservice.NetworkServiceServer) Endpoint {
```
edwarnicke, line 59: Could we simply make this an authzPolicy 'NetworkServiceServer' rather than a list of opa policies?

So instead of passing authzPolicies, I'd pass authorize.NewServer(authzPolicies...) 

Makes the system nice and general :)

PR #259, opened 2020-05-29
OPA: Add policy conception

denis-tingajkin:
Signed-off-by: denis-tingajkin <denis.tingajkin@xored.com>

## Motivation

1. Added abstraction layer for rego query executor
```go 
package opa

type AuthorizationPolicy interface {
	Check(ctx context.Context, input interface{}) error
}
func WithPolicyFromSource(source, query string, checkQuery CheckQueryFunc)AuthorizationPolicy
func WithPolicyFromFile(path, query string, checkQuery CheckQueryFunc)AuthorizationPolicy
```
2.  Added a possibility to use a set of authorization policies for pkg `authorize`.
```go
package authorize

func NewServer(policies ...opa.AuthorizationPolicy) networkservice.NetworkServiceServer
```
3. Added tests to cover.

PR #261, opened 2020-05-30
OPA: Add tokens valid policy

denis-tingajkin:
**Depends on #259**

## Motivation

https://github.com/networkservicemesh/sdk/issues/201 // This PR Implements **1. Tokens valid - checks the validity of some or all of the tokens in the path.**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -172,10 +172,6 @@ issues:
       linters:
         - funlen
       text: "Function 'TestMonitor' is too long"
-    - path: pkg/tools/opa/tokens_expired_policy.go
-      linters:
-        - gosec
-      text: "G101: Potential hardcoded credentials"
     - path: pkg/networkservice/utils/checks/checkerror/server_test.go
       linters:
         - dupl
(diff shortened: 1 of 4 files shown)


pkg/tools/security/token.go
```diff
@@ -37,11 +36,3 @@ func GenerateToken(ctx context.Context, p Provider, expiresAt time.Duration) (st
 
 	return jwt.NewWithClaims(jwt.SigningMethodES256, claims).SignedString(crt.PrivateKey)
 }
-
-// VerifyToken verifies JWT 'token' using x509.Certificate
-func VerifyToken(token string, x509crt *x509.Certificate) error {
```
denis-tingajkin, line 42: NOTE: Now we can verify tokens via OPA.

.golangci.yml
```diff
@@ -172,6 +172,10 @@ issues:
       linters:
         - funlen
       text: "Function 'TestMonitor' is too long"
+    - path: pkg/tools/opa/tokens_valid_policy.go
```
edwarnicke, line 175: Looks like there's a conflict in this file that needs resolving.

Tip on gosec

You can use the comment
```go
 // #nosec
```
to mark lines for gosec to ignore, and thus avoid having to change the .golangci.yml file :)

denis-tingajkin: thanks! Fixed.

PR #264, opened 2020-05-31
skip hostname verification of x509 certs in spiffeutils

fkautz:


pkg/tools/spiffeutils/grpcoptions.go
```diff
@@ -67,5 +68,6 @@ func SpiffeCreds(peer TLSPeer, timeout time.Duration) grpc.ServerOption {
 	if err != nil {
 		return grpc.EmptyServerOption{}
 	}
+	tlsConfig.InsecureSkipVerify = true
```
denis-tingajkin, line 71: As I understand InsecureSkipVerify=true should be used only in testing.
Do we plan to use `SpiffeCreds` only in testing?
If not then I think we can add options to this method something like `WithInsecure()`

fkautz: Typically for most TLS use cases require it when connecting from a client. The hostname is verified against the certificate. However, this ends up failing in SPIFFE. The official SPIFFE golang repo also sets this skip for the same reason.

denis-tingajkin: ok, got it.

PR #265, opened 2020-06-01
Authorization Client: Add possible to generate tokens

denis-tingajkin:

### DIFF
--- a/pkg/networkservice/common/authorize/client.go
+++ b/pkg/networkservice/common/authorize/client.go
@@ -33,18 +33,29 @@ import (
 )
 
 type authorizeClient struct {
-	provider security.Provider
+	provider        security.Provider
+	tokenExpiration time.Duration
 }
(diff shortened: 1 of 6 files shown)


pkg/networkservice/common/authorize/client.go
```diff
@@ -27,19 +32,37 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 )
 
-type authorizeClient struct{}
+type authorizeClient struct {
+	provider security.Provider
+}
+
+const (
+	defaultTokenExpiration = 5 * time.Minute
```
haiodo, line 40: As I remember we discussed something like 15 minutes, should we make this timeout configurable somehow?

denis-tingajkin: Sure, fixed

PR #266, opened 2020-06-02
Revert "Authorization Client: Add possible to generate tokens (#265)"

edwarnicke:
authorize client has nothing to do with token authorization, please see
that work already done in updatpath:

https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/updatepath/client.go

authorize client need to *check* that the tlsInfo/Connection returned from a Client call passes policies, the exact
mirror image of what server does.

Also deleting the pkg/tools/security package as its presence seems to be a source of confusion.

This reverts commit 7c14bd73ab14a2a46c89282bbdf8f98a19d5b720.


edwarnicke:
Revert #265

denis-tingajkin:
ok, I thought that we will use `pkg/tools/security` but looks like it is outdated and was deleted by this PR. Finally got, that this comment  https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/authorize/client.go#L38  related to OPA policies :)

PR #271, opened 2020-06-02
Retire spiffeutils

edwarnicke:


pkg/tools/opa/opainput.go
```diff
@@ -43,17 +41,12 @@ func PreparedOpaInput(ctx context.Context, model interface{}) (map[string]interf
...
 	if cert != nil {
-		spiffeID, err = spiffeutils.SpiffeIDFromX509(cert)
-		if err != nil {
-			return nil, err
-		}
 		pemcert = pemEncodingX509Cert(cert)
 	}
 	result["auth_info"] = map[string]interface{}{
 		"certificate": pemcert,
-		"spiffe_id":   spiffeID,
```
denis-tingajkin, line 56: Why do we need to get rid of spiffe_id for OPA input?

edwarnicke: Yes, the idea is to get it from the Cert.

The reason for getting rid of spiffe_id here is that we want to limit the 'surface' of NSM to spiffe.  Right now for example:
1. Tokens in path are completely agnostic to spiffe.  We provide a token generator that generates them that *can* be used to pass them in.
2. cmd's use the spiffev2 X509Source to get a cert, and from there use it like any other cert.

I *love* spiffe, and think its the way... but I also like good modularity when its inexpensive to get :)

Thoughts?

denis-tingajkin: I think it is right direction :)

PR #275, opened 2020-06-03
Rework localbypass to use endpoint URL 

haiodo:
1. To improve testability and use url of endpoints.
2. Reverse logic to be from registry to make it more easy testable.


pkg/registry/common/localbypass/server.go
```diff
@@ -21,24 +21,36 @@ package localbypass
...
 	"github.com/golang/protobuf/ptypes/empty"
-	"github.com/networkservicemesh/api/pkg/api/registry"
 
 	"github.com/networkservicemesh/sdk/pkg/registry/core/next"
+
+	"github.com/networkservicemesh/api/pkg/api/registry"
 )
 
+// LocalBypass - return interafce to return localbypass endpoints URIs
+type LocalBypass interface {
```
edwarnicke, line 34: Why are we defining a new API here?

pkg/networkservice/common/localbypass/server.go
```diff
@@ -48,34 +42,23 @@ type localBypassServer struct {
 //                        while maintaining the NewServer pattern for use like anything else in a chain.
 //                        The value in *server must be included in the registry.NetworkServiceRegistryServer listening
 //                        so it can capture the registrations.
-func NewServer(registryServer *registry.NetworkServiceRegistryServer) networkservice.NetworkServiceServer {
-	rv := &localBypassServer{}
-	*registryServer = localbypass.NewNetworkServiceRegistryServer(rv)
+func NewServer(reg localbypass.Registry) networkservice.NetworkServiceServer {
```
edwarnicke, line 45: Why are we passing the registry *in* here when before we were passing it *out* ?

haiodo:
@edwarnicke @denis-tingajkin found more correct way for localbypass.

pkg/networkservice/common/localbypass/server.go
```diff
@@ -21,61 +21,43 @@ package localbypass
...
-//                        (and thus a pointer) *registry.NetworkServiceRegistryServer is a double pointer.  Meaning it
-//                        points to a place that points to a place that implements registry.NetworkServiceRegistryServer
-//                        This is done so that we can return a registry.NetworkServiceRegistryServer chain element
-//                        while maintaining the NewServer pattern for use like anything else in a chain.
-//                        The value in *server must be included in the registry.NetworkServiceRegistryServer listening
-//                        so it can capture the registrations.
-func NewServer(registryServer *registry.NetworkServiceRegistryServer) networkservice.NetworkServiceServer {
-	rv := &localBypassServer{}
-	*registryServer = localbypass.NewNetworkServiceRegistryServer(rv)
+func NewServer(storage *memory.Storage) networkservice.NetworkServiceServer {
```
edwarnicke, line 40: Could we please *not* multiple the number of APIs wandering around?  I'm actually not at all happy about memory.Storage currently... I'd prefer not to propogate it to additional packages.

The original pattern involved *returning* the registry piece to the caller via the caller providing a place to provide it.  That kept us to our stock APIs.  Whatever it is you need to do with that registry chain element, it was registry chain element.  Now a developer has to learn about a memory.Storage.

edwarnicke: What if we simply had localbypass.NewServer keep its signature but write a pointer to its internally created memory registry server.  That way you don't have to test the registry bit (its tested already).

haiodo: Both variants not looking good, let's wait for registry simplification to eliminate localbypass at all.

denis-tingajkin: >Could we please not multiple the number of APIs wandering around? I'm actually not at all happy about memory.Storage currently... I'd prefer not to propogate it to additional packages.

I think we'll get rid of `memory.Storage` when registry API simplification will be completed.

PR #277, opened 2020-06-04
OPA: Add tokens chain policy

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/201
This PR Implements **Token.Aud for token n in path matches Token.Sub for n+1 token in chain**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,8 @@ linters-settings:
     suggest-new: true
   dupl:
     threshold: 150
+  funlen:
+    Lines: 100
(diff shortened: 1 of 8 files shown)


.golangci.yml
```diff
@@ -27,6 +27,8 @@ linters-settings:
     suggest-new: true
   dupl:
     threshold: 150
+  funlen:
+    Lines: 100
```
edwarnicke, line 31: LOL... nice touch adopting the new kernel line limit :)

PR #284, opened 2020-06-08
Update for registry next and adapters

haiodo:
Adapters should be respectfull for next items in context after conversion from client to server and back.

+ a bit of cleanups and structuring


denis-tingajkin:
Super interesting how to reproduce the problem what you are fixing in this patch. Can you share steps to reproduce or the chain of your elements where the problem is reproducing?

PR #288, opened 2020-06-08
Client option factory and fixes

haiodo:
+ Allow callback from connect chain element
* Fix tcp support for connect
+ Allow to return tcp address for grpc utils.

### DIFF
--- a/pkg/networkservice/common/connect/server.go
+++ b/pkg/networkservice/common/connect/server.go
@@ -21,6 +21,9 @@ package connect
 
 import (
 	"context"
+	"net/url"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/grpcutils"
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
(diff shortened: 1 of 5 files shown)


pkg/networkservice/common/connect/server.go
```diff
@@ -36,6 +39,19 @@ type connectServer struct {
...
 	connectionIDToClientMap clientMap // key == connection.GetId()
+	dialOptionFactories     []func(ctx context.Context, request *networkservice.NetworkServiceRequest, clientURL *url.URL) []grpc.DialOption
+}
+
+type dialOptionFactory struct {
+	factory func(ctx context.Context, request *networkservice.NetworkServiceRequest, clientURL *url.URL) []grpc.DialOption
+	grpc.EmptyDialOption
+}
+
+// WithDialOptionFactory - define a dial option factory, will be called on moment to create dial
```
edwarnicke, line 50: Could you give an example of where we need a DialOptionFactory rather than just a DialOption?

edwarnicke:
This looks like several things (all of them good) glommed together.  Could you please break this up into the PRs for each of those things?

pkg/networkservice/common/connect/server.go
haiodo: https://github.com/networkservicemesh/cmd-nsmgr/pull/8/files#diff-52369f926d2c6c933dce1bfafc6cb6efR79

haiodo: Factory is required, since we do not know a endpoint uri until Request will be processed, so for callbacks we need to modify dial options.

PR #293, opened 2020-06-09
Add authn options

denis-tingajkin:

## Motivation

This constructor may be useful for forwarder, nsmgr, nse applications

### DIFF
--- a/pkg/networkservice/common/authorize/server.go
+++ b/pkg/networkservice/common/authorize/server.go
@@ -34,10 +34,12 @@ type authorizeServer struct {
 }
 
 // NewServer - returns a new authorization networkservicemesh.NetworkServiceServers
-func NewServer(policies ...opa.AuthorizationPolicy) networkservice.NetworkServiceServer {
-	return &authorizeServer{
-		policies: policies,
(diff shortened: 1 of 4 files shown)


pkg/networkservice/common/authorize/server.go
```diff
@@ -40,6 +40,18 @@ func NewServer(policies ...opa.AuthorizationPolicy) networkservice.NetworkServic
 	}
 }
 
+// NewServerWithDefaultPolicies -  returns a new authorization networkservicemesh.NetworkServiceServers with default OPA policies
+func NewServerWithDefaultPolicies(additionalPolicies ...opa.AuthorizationPolicy) networkservice.NetworkServiceServer {
```
denis-tingajkin, line 44: @edwarnicke Can we use this constructor for cmd-nsmgr, cmd-forwarder, cmd-icmp-responder?

edwarnicke: Rather than this (which breaks pattern) how about we introduce a With pattern here, and have the normal server default to these policies?

denis-tingajkin: good idea, fixed

PR #302, opened 2020-06-10
connect server calls next in the chain rather than returning

fkautz:


haiodo:
@fkautz @edwarnicke 
Let's wait a bit, 
I've almost complete with complex refactoring fixing for connect chain element.

pkg/networkservice/common/connect/server.go
```diff
@@ -84,14 +84,25 @@ func NewServer(clientFactory func(ctx context.Context, cc grpc.ClientConnInterfa
 func (c *connectServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
 	client, found := c.connectionIDToClientMap.Load(request.GetConnection().GetId())
 	if found {
-		return client.Request(ctx, request)
+		conn, err := client.Request(ctx, request)
+		if err != nil {
+			return nil, err
+		}
+		request.Connection = conn
```
edwarnicke, line 91: I don't think we actually want to pass the connection through here... we want to at most pass through the ConnectionContext (and that may be subject to some filtering).

fkautz: I created #306 to track filtering and validation of connection context

fkautz:
> @fkautz @edwarnicke
> Let's wait a bit,
> I've almost complete with complex refactoring fixing for connect chain element.

I'd prefer to get this in. If it gets replaced, that's fine. This is blocking some of the work I'm doing in icmp-responder.

edwarnicke:
We will need to come back and make this a bit more flexible around filtering etc.  But for know this unsticks.

PR #304, opened 2020-06-11
Connect chain rework

haiodo:
+ support close of clients and connection, in progress reference counting.
+ few proper tests
- re-request tests are pending


edwarnicke:
We need to find a way of using Callback that doesn't require it running throughout the code base.
I *like* the callback approach... but it needs to remain a simple thing that can be applied *only* in main via standard dialOptions etc, with its server started in main.  Forcing it all the way up and down the code base is going to be an ongoing problem.

pkg/networkservice/common/connect/server.go
```diff
@@ -65,80 +97,204 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
-	client, _ = c.uRLToClientMap.LoadOrStore(u.String(), client)
-	client, _ = c.connectionIDToClientMap.LoadOrStore(request.GetConnection().GetId(), client)
-	if _, err = client.Request(ctx, request); err != nil {
+	// Will decrease pending and close connection if required
+	defer deferFunc()
+
+	// We have connection ready,
+	var conn *networkservice.Connection
+	conn, err = ce.client.Request(ctx, request)
+	if err != nil {
```
edwarnicke, line 145: Question: I am pretty sure that grpc redials itself if there's an issue (ie, it doesn't fail forever).  What circumstances do you expect in which we would need to reconstitute the whole client?

pkg/networkservice/common/connect/server.go
```diff
@@ -34,27 +35,58 @@ import (
...
+	dialOptions         []grpc.DialOption
+	clientFactory       func(ctx context.Context, conn grpc.ClientConnInterface) networkservice.NetworkServiceClient
+	clientMap           map[string]*clientEntry // key == url as string
+	connections         map[string]*clientEntry // Connection map is required to close using connection id.
+	dialOptionFactories []func(ctx context.Context, clientURL *url.URL) []grpc.DialOption
+	executor            serialize.Executor
+	dialer              func(ctx context.Context, target string, opts ...grpc.DialOption) (conn grpc.ClientConnInterface, cancelFunc func(), err error)
 }
 
 type dialOptionFactory struct {
```
edwarnicke, line 66: We need to find a way to use Callback without having to make odd code changes up and down our codebase (like DialOptFactory).

haiodo: It is not for callback but for testing, I think I could remove it and use grpc endpoint instead.

pkg/networkservice/common/connect/server.go
haiodo: Not sure I understand correctly, grpc created not here, but at line 281, this is handling of client response, and if it failed we need to remove connection stuff. For grpc I've expect only initial dial could fail.

edwarnicke: My concern is that we wind up removing the client no matter what the failure is... it could be as simple as a policy rejection of that specific Request...

pkg/networkservice/common/connect/server.go
edwarnicke: Let's look into that.

pkg/networkservice/common/connect/server.go
```diff
@@ -65,80 +97,204 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
+			if request.Connection.Id != "" {
+				// Delete from CE
+				delete(ce.connections, request.Connection.Id)
+				// Delete from global list
+				delete(c.connections, request.Connection.Id)
+			}
+		})
 		return nil, err
 	}
+	// we succeed, update connection and remove pending state
```
edwarnicke, line 157: I don't quite follow why we have pending state...

pkg/networkservice/common/connect/server.go
```diff
@@ -65,80 +97,204 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
+	})
 	if found {
-		c.connectionIDToClientMap.LoadOrStore(conn.GetId(), client)
-		return client.Close(ctx, conn)
+		// Wait until connection is ready, or context timeout
+		select {
+		case <-ctx.Done():
+			// We failed, we need to remove pending state and close connection if we are last one.
+			<-c.executor.AsyncExec(func() {
+				ce.pending--
```
edwarnicke, line 226: Why are we reference counting?

pkg/networkservice/common/connect/server.go
haiodo: pending state is when we in request, but not have connection yet.
So I'e +1 pending on Any Request/Close, and -1 on defer from calls. And in case I've had connection I've add it to map of connection, and client is not closed/removed until we had connections or pending. I could rename it to activeRequests

pkg/networkservice/common/connect/server.go
haiodo: Agree, let's cleanup on close only.

pkg/networkservice/common/connect/server.go
haiodo: Removed pending state, use connections instead.

pkg/networkservice/common/connect/server.go
```diff
@@ -34,22 +36,39 @@ import (
...
-	dialOptions             []grpc.DialOption
-	clientFactory           func(ctx context.Context, conn grpc.ClientConnInterface) networkservice.NetworkServiceClient
-	uRLToClientMap          clientMap // key == url as string
-	connectionIDToClientMap clientMap // key == connection.GetId()
-	dialOptionFactories     []func(ctx context.Context, request *networkservice.NetworkServiceRequest, clientURL *url.URL) []grpc.DialOption
+	ctx                 context.Context
+	dialOptions         []grpc.DialOption
+	clientFactory       func(ctx context.Context, conn grpc.ClientConnInterface) networkservice.NetworkServiceClient
+	clientMap           map[string]*clientEntry // key == url as string
+	connections         map[string]*clientEntry // Connection map is required to close using connection id.
```
edwarnicke, line 60: 'clientMap' and 'connections' are both better names :)
Would it make sense for 'clientMap' -> 'clients' to match 'connections' ?

pkg/networkservice/common/connect/server.go
```diff
@@ -65,101 +84,205 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
+			ce, connectionExists = c.clientMap[clientURI.String()]
+		}
+		if !connectionExists {
+			// try to find using connection id
+			ce, connectionExists = c.connections[conn.Id]
+		}
+		if !connectionExists {
+			// Since there is no connection, we need to create it and mark as pending as well
+			// We could also be sure we one who will open connection.
+			if clientURI == nil {
```
edwarnicke, line 185: Would it perhaps make more sense to make this check up front in findOrCreate rather than inside the AsyncExec?

pkg/networkservice/common/connect/server.go
```diff
@@ -65,101 +84,205 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
+			return nil, errors.Errorf("context timeout")
+		case <-ce.ready:
+			// all is fine, just return
+			if ce.err != nil {
+				<-c.executor.AsyncExec(func() {
+					delete(ce.connections, conn.Id)
+					c.closeClient(ctx, ce)
+				})
+				return nil, ce.err
+			}
```
edwarnicke, line 223: If you return here for the non-error case, then you can drop the else below (which is a bit more go-like)

pkg/networkservice/common/connect/server.go
```diff
@@ -34,22 +36,39 @@ import (
...
 
 type dialOptionFactory struct {
-	factory func(ctx context.Context, request *networkservice.NetworkServiceRequest, clientURL *url.URL) []grpc.DialOption
+	factory func(ctx context.Context, clientURL *url.URL) []grpc.DialOption
 	grpc.EmptyDialOption
 }
 
 // WithDialOptionFactory - define a dial option factory, will be called on moment to create dial
-func WithDialOptionFactory(factory func(ctx context.Context, request *networkservice.NetworkServiceRequest, clientURL *url.URL) []grpc.DialOption) grpc.DialOption {
+func WithDialOptionFactory(factory func(ctx context.Context, clientURL *url.URL) []grpc.DialOption) grpc.DialOption {
```
edwarnicke, line 71: Why do we need DialOptionFactory?

pkg/networkservice/common/connect/server.go
```diff
@@ -65,101 +84,205 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
+				clientURI:   clientURI,
+				ready:       make(chan bool, 1),
+				connections: map[string]*networkservice.Connection{},
+			}
+			// Put/Update connection
+			c.clientMap[clientURI.String()] = ce
+		}
+		ce.connections[conn.Id] = conn
+	})
+	if ce.err != nil {
```
edwarnicke, line 199: I don't quite follow why we are doing this here rather than at the point we are assigning non-nil values to c.err

pkg/networkservice/common/connect/server.go
```diff
@@ -65,101 +84,205 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
+				return nil, ce.err
+			}
+		}
+	} else {
+		// Dial and create client connection
+		if err := c.createClient(ctx, ce); err != nil {
+			// If we failed to create client, we failed to dial, so no need to close,
+			// we need to mark it as error one and all clients pending will return errors, and last one will remove entry
+			// Mark connection as ready to use
+			<-c.executor.AsyncExec(func() {
```
edwarnicke, line 231: I don't quite follow why we are wrapping this in an AsyncExec for the whole server.  It looks like the ce objects are marked as ready by their creator, and thus we don't need to get exclusivity at the server level for this operation.

pkg/networkservice/common/connect/server.go
```diff
@@ -65,101 +84,205 @@ func WithDialOptionFactory(factory func(ctx context.Context, request *networkser
...
+				ce.err = err
+				// Mark as ready but with error
+				ce.markAsReady(false)
+				c.closeClient(ctx, ce)
+			})
+			return nil, err
+		}
+
+		// Mark connection as ready to use
+		<-c.executor.AsyncExec(func() {
```
edwarnicke, line 241: I don't quite follow why we are wrapping this in an AsyncExec for the whole server. It looks like the ce objects are marked as ready by their creator, and thus we don't need to get exclusivity at the server level for this operation.

pkg/networkservice/common/connect/server.go
haiodo: Sounds good.

pkg/networkservice/common/connect/server.go
haiodo: In this case it will be required to have two AsyncExecs, since this check is for case then client URI is not passed with context and there is no existing connection. 
This is useful if we want to Heal/Close if we had only Connection object with Id, and do not store client url from context. If we will always require clientURI to be passed with context, in this case connectServer.connections are not required at all, and could be removed and code will be more simple.

pkg/networkservice/common/connect/server.go
haiodo: Thanks.

pkg/networkservice/common/connect/server.go
haiodo: callback.WithCallbackDialer requires target uri to select a proper callback connection, but in case of client we know target uri only inside Request, so we need a way to construct a proper callback.WithCallbackDialer() grpc option.

pkg/networkservice/common/connect/server.go
haiodo: Simplified, but it will require if clientURI is not passed to check for connections for it.

pkg/networkservice/common/connect/server.go
haiodo: Fixed.

pkg/networkservice/common/connect/server.go
haiodo: We access shared structures inside closeClient.

pkg/networkservice/common/connect/server.go
edwarnicke: Would WithCallbackDialer be reworked to not require the uri ?  Dialer gets a target which could be parsed to such a uri anyway, correct?

pkg/networkservice/common/connect/server.go
```diff
@@ -56,8 +56,8 @@ type connectServer struct {
 	ctx                 context.Context
 	dialOptions         []grpc.DialOption
 	clientFactory       func(ctx context.Context, conn grpc.ClientConnInterface) networkservice.NetworkServiceClient
-	clientMap           map[string]*clientEntry // key == url as string
-	connections         map[string]*clientEntry // Connection map is required to close using connection id.
+	clients             map[string]*clientEntry // key == url as string
+	connections         map[string]*url.URL     // Connection map is required to close using connection id.
```
edwarnicke, line 60: This seems to be more of a clientURLs than a connections ...

pkg/networkservice/common/connect/server.go
```diff
@@ -168,40 +178,22 @@ func (c *connectServer) Close(ctx context.Context, conn *networkservice.Connecti
...
 
-func (c *connectServer) findOrCreateClient(ctx context.Context, conn *networkservice.Connection) (ce *clientEntry, err error) {
+func (c *connectServer) findOrCreateClient(ctx context.Context, clientURI *url.URL, conn *networkservice.Connection) (ce *clientEntry, err error) {
 	connectionExists := false
-	clientURI := clienturl.ClientURL(ctx)
 	<-c.executor.AsyncExec(func() {
-		if clientURI != nil {
-			ce, connectionExists = c.clientMap[clientURI.String()]
-		}
+		ce, connectionExists = c.clients[clientURI.String()]
```
edwarnicke, line 184: I like this change a lot.

pkg/networkservice/common/connect/server.go
haiodo: Yes, I'm still not fully sure should it be separate chain element and request clientURI to be part of context always.

pkg/networkservice/common/connect/server.go
```diff
@@ -34,22 +36,39 @@ import (
...
+	ready           chan bool // Flag for connection is establishing
+	connections     map[string]*networkservice.Connection
+	err             error
+	closeConnection func() error
+}
+
+func (ce *clientEntry) markAsReady(state bool) {
+	ce.ready <- state
+	close(ce.ready)
+}
```
edwarnicke, line 53: Couldn't we just make ce.ready a chan struct{} and simply close it?

haiodo: Yes, we can :-)

pkg/networkservice/common/connect/server.go
edwarnicke: Say more.

PR #310, opened 2020-06-15
Fix Data race in next.

haiodo:
Fix Data race in next.


edwarnicke:
Good catch.  Very subtle issue.

PR #317, opened 2020-06-17
connect.Close() must call client.Close before next.Server(ctx).Close

edwarnicke:
Otherwise, we don't get correct cleanup in xconnects

### DIFF
--- a/pkg/networkservice/common/connect/server.go
+++ b/pkg/networkservice/common/connect/server.go
@@ -155,17 +155,9 @@ func (c *connectServer) Close(ctx context.Context, conn *networkservice.Connecti
 		// In case of error, no pending state is propogated.
 		return nil, err
 	}
-	// Call next Close before calling client Close
-	_, err = next.Server(ctx).Close(ctx, conn)
 
 	// Call client close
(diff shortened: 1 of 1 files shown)


pkg/networkservice/common/connect/server.go
```diff
@@ -155,26 +155,21 @@ func (c *connectServer) Close(ctx context.Context, conn *networkservice.Connecti
 		// In case of error, no pending state is propogated.
 		return nil, err
 	}
-	// Call next Close before calling client Close
```
haiodo, line 158: I've done this with purpose, since if remote nsmgr/endpoint will die, we need to cleanup on Close all our next chains in any case.

edwarnicke: OK... after a long discussion on slack, I think that we landed at:

Close is responsible for calling downstream Closes in the chain, even if it encounters an error in cleaning up its own state (in this case, client.Close() is cleaning up its own state).

To that end, I've updated this PR to call next.Server(ctx).Close() even if client.Close() returns an error.  @haiodo Does this handle your concerns?

PR #320, opened 2020-06-18
Fix: nextwrap ignores inner client find result 

denis-tingaikin:

## Motivation

nextwrap ns/nse clients currently ignore inner client find result


pkg/registry/core/nextwrap/ns_registry_client.go
```diff
@@ -40,10 +41,13 @@ func (r *nextNetworkServiceWrappedClient) Register(ctx context.Context, in *regi
 }
 
 func (r *nextNetworkServiceWrappedClient) Find(ctx context.Context, in *registry.NetworkServiceQuery, opts ...grpc.CallOption) (registry.NetworkServiceRegistry_FindClient, error) {
-	_, err := r.client.Find(ctx, in, opts...)
-	if err != nil {
+	client, err := r.client.Find(ctx, in, opts...)
+	if err != nil && err != io.EOF {
 		return nil, err
 	}
+	if client != nil {
```
haiodo, line 48: I think we need to have aggregator from two streams here.

haiodo:
```go
type nsAggrClient struct {
	grpc.ClientStream
	ctx     context.Context
	streams []registry.NetworkServiceRegistry_FindClient
}

func (nsAggr *nsAggrClient) Recv() (*registry.NetworkService, error) {
	for {
		if nsAggr.streams == nil {
			return nil, io.EOF
		}
		res, err := nsAggr.streams[0].Recv()
		if err == io.EOF {
			nsAggr.streams = nsAggr.streams[1:]
			continue
		}
		return res, nil
	}
	return nil, nil
}

func (nsAggr *nsAggrClient) Context() context.Context {
	return nsAggr.ctx
}

func (r *nextNetworkServiceWrappedClient) Find(ctx context.Context, in *registry.NetworkServiceQuery, opts ...grpc.CallOption) (registry.NetworkServiceRegistry_FindClient, error) {
	clientClient, err := r.client.Find(ctx, in, opts...)
	if err != nil {
		return nil, err
	}
	var nextClient registry.NetworkServiceRegistry_FindClient
	nextClient, err = next.NetworkServiceRegistryClient(ctx).Find(ctx, in, opts...)
	if err != nil {
		return nil, err
	}
	return &nsAggrClient{
		ctx:     ctx,
		streams: []registry.NetworkServiceRegistry_FindClient{clientClient, nextClient},
	}, nil
}
```

Thoughts?

denis-tingajkin: @haiodo thanks, I think it is more correct direction. I'll improve ur suggestion to support monitoring cases.

denis-tingajkin:
NOTE:

Pkg `aggregate` can be used in chains to aggregate a few find clients, for example
```go
	c1 := myFindClient{}
	c2, err := next.NetworkServiceEndpointRegistryClient(ctx).Find(ctx, in, opts...)
	return aggregate.NewNetworkServiceEndpointFindClient(c1, c2), nil
```

denis-tingajkin:
@haiodo for this case we should apply the fix for registries pkg next to allow next streaming. The tail element of streaming should be the source. Also, I think we should get rid of nextwrap pkg because the source stream should be in tail we should not wrap them or aggregate.

denis-tingajkin:
@edwarnicke we've little bit discussed with @haiodo about your suggestion to use valid next chain (A(B(C))) and we found that we have a case when stream source client can be in the middle of the chain. For this case, I've added `streamsource` pkg that solves this problem.  Do you have any concerns about this approach?

edwarnicke:
> @edwarnicke we've little bit discussed with @haiodo about your suggestion to use valid next chain (A(B(C))) and we found that we have a case when stream source client can be in the middle of the chain. For this case, I've added `streamsource` pkg that solves this problem. Do you have any concerns about this approach?

I don't understand why having a 'streamsource' in the middle of the chain is a problem... it feels like we are starting to massively overcomplicated a fairly simply thing:

If I have a chain: chain(A,B,C) that has a method that returns a stream (say StreamReturningMethod) then:

pkg/registry/core/streamsource/ns_registry.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type nsClient struct{}
+
+func (n *nsClient) Register(ctx context.Context, in *registry.NetworkService, opts ...grpc.CallOption) (*registry.NetworkService, error) {
+	return next.NetworkServiceRegistryClient(ctx).Register(ctx, in, opts...)
+}
+
+func (n *nsClient) Find(ctx context.Context, in *registry.NetworkServiceQuery, opts ...grpc.CallOption) (registry.NetworkServiceRegistry_FindClient, error) {
+	streams := networkServiceStreams(ctx)
```
haiodo, line 37: Should we in case there is 0,1 just return without wrapping?

pkg/registry/core/streamsource/nse_registry.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type nseClient struct{}
+
+func (n *nseClient) Register(ctx context.Context, in *registry.NetworkServiceEndpoint, opts ...grpc.CallOption) (*registry.NetworkServiceEndpoint, error) {
+	return next.NetworkServiceEndpointRegistryClient(ctx).Register(ctx, in, opts...)
+}
+
+func (n *nseClient) Find(ctx context.Context, in *registry.NetworkServiceEndpointQuery, opts ...grpc.CallOption) (registry.NetworkServiceEndpointRegistry_FindClient, error) {
+	streams := networkServiceEndpointStreams(ctx)
```
haiodo, line 37: Same here, in case 0,1 we should not wrap with aggregator I suppose.

pkg/registry/core/aggregate/ns_server.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+func (c *nsAggregateClient) Context() context.Context {
+	return c.ctx
+}
+
+// NewNetworkServiceFindClient aggregates few NetworkServiceRegistry_FindClient to single  NetworkServiceRegistry_FindClient
+func NewNetworkServiceFindClient(clients ...registry.NetworkServiceRegistry_FindClient) registry.NetworkServiceRegistry_FindClient {
+	d := int32(0)
```
haiodo, line 76: We could check for 0,1 before or here to not wrap in case it is not really required.

denis-tingajkin: thanks, fixed

pkg/registry/core/streamsource/ns_registry.go
denis-tingajkin: thanks, fixed

pkg/registry/core/streamsource/nse_registry.go
denis-tingajkin: thanks, fixed

edwarnicke:
I don't see why sourcing in the middle of the chain is actually a problem.

Say I have chain(A,B,C)

I call chain(A,B,C).Find(...) and get back chain(A.Client,B.Client,C.Client)

I call chain(A.Client,B.Client,C.Client).Recv()

Normal Case(Chain Tail(C) is source):

A.Client.Recv() calls B.Client.Recv() calls C.Client.Recv(), C.Recv() returns a NSE entry.

Middle Source Case(Chain element B is source):

A.Client.Recv() calls B.Client.Recv(), B.Client has an NSE it's sourcing, so simply returns that NSE.

Its quite the same as any other chain processing where something occurs int the middle of the chain that causes early chain termination.

Thoughts?

haiodo:
Let's not close it until we will find a example/usage scenario when we need 2 sources at once.

denis-tingajkin:
@edwarnicke It sounds valid for me. So let's just wait, probably we can find any valid example where we need to support two or more source streams.

edwarnicke:
> Let's not close it until we will find a example/usage scenario when we need 2 sources at once.

I'm curious what you think the issue is for 2 sources at once here?  Is it a question of event order?

edwarnicke:
@denis-tingaikin Can this be closed?

denis-tingaikin:
@edwarnicke I think yes :)

Issue #297, opened 2020-06-10
Improve performance of chain of adapters + next


denis-tingajkin:
## Description

Chain of next + adapters elements can produce extra adapters per each of the chain elements. 

For example: 
If we have a chain of
```
server->adapted client->server->server
```
then it can be represented on call Request/Register/Close:
```
server->adapted client->adapted server->adapted server
```

We need to investigate and fix extra adaptation.

PR #321, opened 2020-06-19
Fix performance for next and adapter chains

denis-tingajkin:
## Motivation

https://github.com/networkservicemesh/sdk/issues/297

With this patch per each adapter in chain, adapter will be called once(in master version twice)

### DIFF
--- a/pkg/networkservice/common/updatepath/updatepath_test.go
+++ b/pkg/networkservice/common/updatepath/updatepath_test.go
@@ -41,7 +41,7 @@ func TestChain(t *testing.T) {
 	want := &networkservice.Connection{
 		Id: "conn-2",
 		Path: &networkservice.Path{
-			Index: 0,
+			Index: 2,
 			PathSegments: []*networkservice.PathSegment{
 				{
 					Name:    "nsc-1",
@@ -73,5 +73,5 @@ func TestChain(t *testing.T) {
 
 	got, err := server.Request(context.Background(), request)
 	assert.Equal(t, want, got)
-	assert.Equal(t, nil, err)
+	assert.NoError(t, err)
 }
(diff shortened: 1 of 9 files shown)


pkg/networkservice/common/updatepath/updatepath_test.go
```diff
@@ -41,7 +41,7 @@ func TestChain(t *testing.T) {
 	want := &networkservice.Connection{
 		Id: "conn-2",
 		Path: &networkservice.Path{
-			Index: 0,
+			Index: 2,
```
denis-tingajkin, line 44: @edwarnicke I found that in a master version by some reason index did not increment. Can you please verify this line change?

pkg/networkservice/core/adapters/server_to_client.go
```diff
@@ -36,9 +38,27 @@ func NewServerToClient(server networkservice.NetworkServiceServer) networkservic
...
-	return s.server.Request(ctx, in)
+	conn, err := s.server.Request(ctx, in)
+	if err != nil {
+		return nil, err
+	}
+	if in == nil {
+		in = &networkservice.NetworkServiceRequest{}
+	}
+	in.Connection = conn
+	if !next.Done(ctx) {
```
denis-tingajkin, line 49: NOTE: I've added flag `done` for pkg next to support case when we have a chain of 
s -> adapted c - > s and for example client did broke the chain, then we should not call next server.

pkg/networkservice/common/updatepath/updatepath_test.go
denis-tingajkin: As I can see here is used to update servers and looks like here should be 2. Looks like this patch also fixed issue with this test :)

pkg/networkservice/common/updatepath/updatepath_test.go
```diff
@@ -73,5 +73,5 @@ func TestChain(t *testing.T) {
 
 	got, err := server.Request(context.Background(), request)
 	assert.Equal(t, want, got)
-	assert.Equal(t, nil, err)
+	assert.NoError(t, err)
```
edwarnicke, line 76: Thanks for fixing this :)

Issue #307, opened 2020-06-13
Registry: add resources cleanup chain element


denis-tingajkin:
## Motivation

With the new registry model, we need to add a chain element to manage NSM resources(NSs,NSEs). This component should do next things:

1. Deleting NS if the registry has not NSE entries that provide this NS.
2. Deleting NSE if achieved expiration time.

edwarnicke:
> Deleting NS if the registry has not NSE entries that provide this NS.
We definitely don't want to do this.  You can easily imagine someone pre-provisioning an NS for which they do not have NSEs for example.

> Deleting NSE if achieved expiration time.
Yep, something similar to the timeout chain element.

I'd also add to the list for both NS and NSE:
- refresh - a chain element that refreshes registration at 1/3 of the time between initial registration and expiration

denis-tingajkin:
>I'd also add to the list for both NS and NSE:

>refresh - a chain element that refreshes registration at 1/3 of the time between initial registration and expiration

@edwarnicke sounds good. Can we add this element as a separate issue/PR?

edwarnicke:
@denis-tingajkin I'd prefer it :)  Modularity and small thematic PRs for the win!

PR #322, opened 2020-06-20
Implement registry expire chain element

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/307

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -174,6 +174,10 @@ issues:
         - govet
       text: "lostcancel"
     # Sometimes test functions just get long :(
+    - path: .*
+      linters:
+        - golint
+      text: "returns unexported type"
     - path: pkg/networkservice/common/monitor/server_test.go
       linters:
         - funlen
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -174,10 +174,6 @@ issues:
         - govet
       text: "lostcancel"
     # Sometimes test functions just get long :(
-    - path: .*
-      linters:
-        - golint
-      text: "returns unexported type"
     - path: pkg/networkservice/common/monitor/server_test.go
       linters:
         - funlen
--- a/pkg/registry/common/expire/common.go
+++ b/pkg/registry/common/expire/common.go
@@ -24,19 +24,13 @@ import (
 
 const defaultPeriod = time.Second * 5
 
-func defaultNowFunc() func() int64 {
-	return func() int64 {
-		return time.Now().Unix()
-	}
-}
-
-func getExpiredNSEs(nseMap map[string]*registry.NetworkServiceEndpoint, now int64) []*registry.NetworkServiceEndpoint {
+func getExpiredNSEs(nseMap map[string]*registry.NetworkServiceEndpoint) []*registry.NetworkServiceEndpoint {
 	var list []*registry.NetworkServiceEndpoint
 	for _, v := range nseMap {
 		if v.ExpirationTime == nil {
 			continue
 		}
-		if now >= v.ExpirationTime.Seconds {
+		if time.Until(time.Unix(v.ExpirationTime.Seconds, int64(v.ExpirationTime.Nanos))) <= 0 {
 			list = append(list, v)
 		}
 	}
--- a/pkg/registry/common/expire/common.go
+++ b/pkg/registry/common/expire/common.go
@@ -1,0 +1,36 @@
+// Copyright (c) 2020 Cisco Systems, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package expire
+
+import (
+	"time"
+
+	"github.com/networkservicemesh/api/pkg/api/registry"
+)
+
+func removeExpiredNSEs(nseMap map[string]*registry.NetworkServiceEndpoint) []*registry.NetworkServiceEndpoint {
+	var list []*registry.NetworkServiceEndpoint
+	for _, v := range nseMap {
+		if v.ExpirationTime == nil {
+			continue
+		}
+		if time.Now().UnixNano() >= v.ExpirationTime.Seconds {
+			list = append(list, v)
+		}
+	}
+	return list
+}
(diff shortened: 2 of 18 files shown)


pkg/registry/common/expire/ns_server.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (n *nsServer) Register(ctx context.Context, request *registry.NetworkService) (*registry.NetworkService, error) {
+	n.once.Do(n.monitor)
+	if n.monitorErr != nil {
+		return nil, n.monitorErr
+	}
+	<-n.executor.AsyncExec(func() {
+		n.nss[request.Name] = request
+	})
+	return n.server.Register(ctx, request)
```
haiodo, line 100: You should remove n.nss if server register will fail.

pkg/registry/common/expire/nse_server.go
```diff
@@ -0,0 +1,77 @@
...
+	nses      map[string]*registry.NetworkServiceEndpoint
+	nsesMutex sync.Mutex
+	period    time.Duration
+	once      sync.Once
+	server    registry.NetworkServiceEndpointRegistryServer
+}
+
+func (n *nseServer) Register(ctx context.Context, nse *registry.NetworkServiceEndpoint) (*registry.NetworkServiceEndpoint, error) {
+	n.once.Do(n.monitor)
+	n.nsesMutex.Lock()
```
haiodo, line 38: Why not use async synchronizers here?

haiodo: Also you need to remove from list in case n.sever.Register return error.

pkg/registry/memory/ns_seever_test.go
```diff
@@ -1,3 +1,19 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
```
haiodo, line 1: Is it right file name?

pkg/registry/memory/ns_server.go
```diff
@@ -13,43 +29,43 @@ import (
...
-	n.networkServices.Store(ns.Name, ns)
 	n.executor.AsyncExec(func() {
-		n.eventChannelsLocker.Lock()
+		n.networkServices[ns.Name] = ns
 		for _, ch := range n.eventChannels {
 			ch <- ns
 		}
-		n.eventChannelsLocker.Unlock()
 	})
 	return next.NetworkServiceRegistryServer(ctx).Register(ctx, ns)
```
haiodo, line 45: We need to cleanup if next return error.

pkg/registry/memory/nse_server.go
```diff
@@ -13,82 +29,82 @@ import (
...
-	n.networkServiceEndpoints.Store(nse.Name, nse)
 	n.executor.AsyncExec(func() {
-		n.eventChannelsLocker.Lock()
+		n.networkServiceEndpoints[nse.Name] = nse
 		for _, ch := range n.eventChannels {
 			ch <- nse
 		}
-		n.eventChannelsLocker.Unlock()
 	})
 	return next.NetworkServiceEndpointRegistryServer(ctx).Register(ctx, nse)
```
haiodo, line 45: Same here, cleanup if next return error.

pkg/registry/common/expire/common.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/networkservicemesh/api/pkg/api/registry"
+)
+
+func removeExpiredNSEs(nseMap map[string]*registry.NetworkServiceEndpoint) []*registry.NetworkServiceEndpoint {
+	var list []*registry.NetworkServiceEndpoint
+	for _, v := range nseMap {
+		if v.ExpirationTime == nil {
+			continue
+		}
+		if time.Now().UnixNano() >= v.ExpirationTime.Seconds {
```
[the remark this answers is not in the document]
denis-tingajkin: Thanks, fixed

pkg/registry/common/expire/ns_server.go
denis-tingajkin: thanks, fixed

pkg/registry/common/expire/ns_server.go
```diff
@@ -0,0 +1,136 @@
...
+				for _, service := range nse.NetworkServiceName {
+					n.nsCounter[service]++
+				}
+			})
+		}
+	}()
+	go func() {
+		for {
+			var list []*registry.NetworkService
+			n.executor.AsyncExec(func() {
```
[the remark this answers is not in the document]
denis-tingajkin: The reason than we have a channel with nsm resources(we read it) and we also need to able to check nsm resources by period.

pkg/registry/memory/ns_server.go
denis-tingajkin: nice catch, fixed

pkg/registry/memory/nse_server.go
denis-tingajkin: nice catch, fixed

pkg/registry/common/expire/ns_server.go
```diff
@@ -0,0 +1,146 @@
...
+		})
+		if err != nil {
+			n.monitorErr = err
+			return
+		}
+		for event := range registry.ReadNetworkServiceEndpointChannel(c) {
+			nse := event
+			n.Lock()
+			n.nses[nse.Name] = nse
+			for _, service := range nse.NetworkServiceNames {
```
haiodo, line 64: It looks not really correct, if for some of NetworkServices it would be a lot of updates, n.nsCounter for it will be big enough so service will be newer removed, since it is removed if and only if there is no network services. 
So I would recommend to handle a list of endpoints implementing a network service instead of having an counter.

denis-tingajkin: Good catch, I've added a check that incrementing should be only in case if nse does not exist in nses map.

.golangci.yml
```diff
@@ -174,6 +174,10 @@ issues:
         - govet
       text: "lostcancel"
     # Sometimes test functions just get long :(
+    - path: .*
+      linters:
+        - golint
+      text: "returns unexported type"
```
edwarnicke, line 180: I don't quite follow what this is doing... it looks like it may be overly general... but I can't tell.  Also, the comment about it goes with the next exception.

denis-tingajkin: oops! Yes, you are right, it is wrong to change. Fixed!

denis-tingajkin:
@edwarnicke golint change is removed, thanks for your attention

pkg/registry/memory/ns_server.go
```diff
@@ -62,26 +64,41 @@ func (n *networkServiceRegistryServer) Find(query *registry.NetworkServiceQuery,
...
 					}
 					if err := s.Send(event); err != nil {
 						return err
 					}
 				}
+				return nil
+			}
+		}
+		for {
+			err := notifyChannel()
```
haiodo, line 96:
```suggestion
			err := processChannel(query, s, eventCh)
```

haiodo:
@denis-tingajkin  Please proceed comments with next PR.

PR #326, opened 2020-06-22
Do not return io.EOF for tail server registry chain elements

denis-tingajkin:

## Motivation

@haiodo has found during nsmgr + registry grpc testing that we should get rid of using io.EOF for our adapters and next stream chain elements because it works wrong with grpc.

### DIFF
--- a/pkg/registry/memory/nse_server.go
+++ b/pkg/registry/memory/nse_server.go
@@ -1,3 +1,19 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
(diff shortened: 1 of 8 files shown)


pkg/registry/memory/nse_server.go
```diff
@@ -45,44 +61,37 @@ func (n *networkServiceEndpointRegistryServer) Find(query *registry.NetworkServi
...
-				case event := <-eventCh:
-					if matchutils.MatchNetworkServiceEndpoints(query.NetworkServiceEndpoint, event) {
-						if err := s.Send(event); err != nil {
-							return
-						}
+		eventCh := make(chan *registry.NetworkServiceEndpoint, n.eventChannelSize)
+		n.eventChannelsLocker.Lock()
+		n.eventChannels = append(n.eventChannels, eventCh)
+		n.eventChannelsLocker.Unlock()
+		err := sendAllMatches(query.NetworkServiceEndpoint)
```
haiodo, line 68: Should we combine next.NetworkServiceEndpointRegistryServer(s.Context()).Find(query, s) in case of watch somehow?

PR #328, opened 2020-06-22
Check endpoint and return error

haiodo:
Check endpoint and return error


pkg/networkservice/common/roundrobin/server.go
```diff
@@ -63,6 +63,9 @@ func (s *selectEndpointServer) withClientURL(ctx context.Context, conn *networks
 	if clienturl.ClientURL(ctx) == nil {
 		candidates := discover.Candidates(ctx)
 		endpoint := s.selector.selectEndpoint(candidates.NetworkService, candidates.Endpoints)
+		if endpoint == nil {
```
denis-tingajkin, line 66: I would also recommend adding a unit test to cover this branch

PR #332, opened 2020-06-22
Introduce simple server.WithCallbackDialer() that does not require arguments

edwarnicke:
In many places in our code, we need to be able to pass in a simple grpc.DialOption
rather than a function that can be called to return them.  This makes that possible.

### DIFF
--- a/pkg/tools/callback/server.go
+++ b/pkg/tools/callback/server.go
@@ -27,6 +27,9 @@ import (
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/grpcutils"
+	"github.com/networkservicemesh/sdk/pkg/tools/log"
 )
 
 // ClientListener - inform server about new client is arrived or disconnected.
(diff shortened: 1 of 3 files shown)


pkg/tools/callback/server.go
```diff
@@ -21,12 +21,20 @@ import (
...
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/grpcutils"
+	"github.com/networkservicemesh/sdk/pkg/tools/log"
+)
+
+const (
+	callbackPrefix = "callback:"
```
denis-tingajkin, line 37:
```suggestion
	callbackPrefix = "callback"
```

pkg/tools/callback/server.go
```diff
@@ -99,8 +99,7 @@ func (s *serverImpl) HandleCallbacks(serverClient CallbackService_HandleCallback
 
 	ctx, cancelFunc := context.WithCancel(serverClient.Context())
 	defer cancelFunc()
-
-	handleID := fmt.Sprintf("callback:%v", clientID)
+	handleID := fmt.Sprintf("%s%v", callbackPrefix, clientID)
```
denis-tingajkin, line 102:
```suggestion
	handleID := fmt.Sprintf("%s:%v", callbackPrefix, clientID)
```

pkg/tools/callback/server.go
haiodo: I do not think it is required, it could be any prefix. I've choose one with ':'

PR #351, opened 2020-06-29
Fix TODO comment "Reconsider match_selector as a part of 'tools'"

denis-tingaikin:

## Motivation

Fixed TODO comment https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/discover/match_selector.go#L28


haiodo:
@denis-tingajkin  should only isSubset to be moved to tools? I do not think tools should know about endpoints.

denis-tingaikin:
PR is outdated.

PR #354, opened 2020-06-29
Add a sync.Map for map[string]networkservice.NetworkServiceClient

edwarnicke:
Also provides a refcounting version


pkg/tools/clientmap/refcount.go
```diff
@@ -0,0 +1,90 @@
...
+type RefcountMap struct {
+	Map
+}
+
+// Store sets the value for a key.
+// Increments the refcount
+func (r *RefcountMap) Store(key string, value networkservice.NetworkServiceClient) {
+	client := &refcountClient{client: value}
+	r.Map.Store(key, client)
+	for count := int32(0); !atomic.CompareAndSwapInt32(&client.count, count, count+1); count++ {
```
fkautz, line 52: Would it be more appropriate to use this?

https://golang.org/pkg/sync/atomic/#AddInt32

PR #355, opened 2020-06-30
Rework connect chain element to be simpler, clearer, and more performant.

edwarnicke:


edwarnicke:
@haiodo Have a look at the DESIGN.md.  I attempt to explain my thinking there.  I believe this is both functionally correct, also easier to follow, and has better concurrency characteristics.

Also, your TestParallel is a vicious vicious test that caught a huge amount of bugs in my attempted implementations :)

pkg/networkservice/common/connect/client.go
```diff
@@ -0,0 +1,69 @@
...
+func NewClient(cancel context.CancelFunc) networkservice.NetworkServiceClient {
+	return &connectClient{
+		cancel: cancel,
+	}
+}
+
+func (c *connectClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+	var conn *networkservice.Connection
+	var err error
+	<-c.executor.AsyncExec(func() {
```
denis-tingajkin, line 47: Can we get rid of using `executor` here? Because we are not using async here and look like we just creating additional goroutine and just join.
```suggestion
```

pkg/networkservice/common/connect/client.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		conn, err = next.Client(ctx).Request(ctx, clientRequest)
+		c.connection = conn
+	})
+	return conn, err
+}
+
+func (c *connectClient) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
+	var err error
+	<-c.executor.AsyncExec(func() {
```
denis-tingajkin, line 61: The same here. NOTE: mutex more effective for cases when needed just sync access.

pkg/networkservice/common/connect/server.go
```diff
@@ -1,5 +1,3 @@
-// Copyright (c) 2020 Doc.ai and/or its affiliates.
```
denis-tingajkin, line 1: Can we remove previous copyright holders if the file is reworked? 🤔

pkg/networkservice/common/connect/server.go
```diff
@@ -21,247 +19,112 @@ package connect
...
-				connections: map[string]*networkservice.Connection{},
-			}
-			// Put/Update connection
-			c.clients[clientURI.String()] = ce
+func (c *connectServer) client(ctx context.Context, conn *networkservice.Connection) networkservice.NetworkServiceClient {
+	// Fast path, if we have a client for this conn.GetId(), use it
+	client, _ := c.clientsByID.Load(conn.GetId())
+
+	// If we didn't find a client, we fall back to clientURL
+	if client == nil {
```
denis-tingajkin, line 93: Imagine the case when client called two times Reuqest in parallel in this case this condition for one client can be true twice. When expected that the first call added to 'cache' map client and the second call should just load this value. I also think that this problem can be detected by a test with few requests in separate goroutines.

pkg/networkservice/common/connect/server.go
```diff
@@ -21,247 +19,112 @@ package connect
...
-	dialOptions   []grpc.DialOption
-	clientFactory func(ctx context.Context, conn grpc.ClientConnInterface) networkservice.NetworkServiceClient
-	clients       map[string]*clientEntry // key == url as string
-	connections   map[string]*url.URL     // Connection map is required to close using connection id.
-	executor      serialize.Executor
+	ctx               context.Context
+	clientFactory     func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
+	clientDialOptions []grpc.DialOption
+	clientsByURL      clientmap.RefcountMap // key == clientURL.String()
+	clientsByID       clientmap.Map         // key == client connection ID
```
denis-tingajkin, line 40: I think to prevent race conditions we need to use mutexes.
```suggestion
	clients            map[string]*clientEntry // key == url as string		clientsByURL      clientmap.RefcountMap // key == clientURL.String()
	connections        map[string]*url.URL  
	lockers            map[string]sync.Mutex
```

pkg/networkservice/common/connect/server.go
edwarnicke: We actually have tests with Requests in different go routines :) (10 of them).

You are correct that we could, in theory, have client == nil be true here for multiple nearly simultaneous Requests with the same Conn.Id(). See comment below for why we don't care :)

pkg/networkservice/common/connect/server.go
```diff
@@ -21,247 +19,112 @@ package connect
...
+			// maps.
+			client = chain.NewNetworkServiceClient(
+				NewClient(func() {
+					c.clientsByID.Delete(conn.GetId())
+					c.clientsByURL.Delete(clientURL.String())
+				}),
+				client,
+			)
+			var loaded bool
+			client, loaded = c.clientsByID.LoadOrStore(conn.GetId(), client)
```
edwarnicke, line 120: Even if we get near simultaneous initial requests for the same Connection.Id causing two separate cases where client == nil in line 93... they would both hit this point, where precisely *one* of them would win on the LoadOrStore.  One would be Stored, the other would be Loaded.  If you see the comment below about why we have the if loaded {...} its actually cleaning up after the clientByURL refcount for that case.

So even in the *extremely* unlikely case of nearly simultaneous initial Requests... we maintain correctness, at a price much much smaller than global mutexes.  And in the common case without this collision, enormously smaller than global mutexes.

pkg/networkservice/common/connect/server.go
edwarnicke: We actually don't need global mutexes to avoid race conditions (see comments below).  I *think* we produce correct behavior, even in the case of nearly simultaneous initial Requests for the same connection.ID.

pkg/networkservice/common/connect/client.go
edwarnicke: You are correct, mutex is the better choice here as we simply need exclusivity.

I had somehow convinced myself we also needed to preserve order (which executor does and mutex does not)... but as I tried to write the explanation of why, I found I was mistaken :)

pkg/networkservice/common/connect/server.go
edwarnicke: Fixed :)

pkg/networkservice/common/connect/server.go
edwarnicke: Also updated the comments in the file to reflect all of this to make sure its clear to subsequent readers :)

denis-tingajkin: Looks good to me with the flag `loaded` it will work correctly :)

pkg/networkservice/common/connect/server.go
denis-tingajkin: Definitely you are right looks like global mutex can be more costly for those cases.

PR #356, opened 2020-06-30
Fix: refresh client should set expire_time before call the next client

denis-tingajkin:

## Description

During testing in registry-memory found that refresh client chain element is working incorrectly with grpc. It should set expiration_time for nse before call the next client.

### DIFF
--- a/pkg/registry/common/refresh/nse_registry_client.go
+++ b/pkg/registry/common/refresh/nse_registry_client.go
@@ -31,15 +31,15 @@ import (
 )
 
 type refreshNSEClient struct {
-	client                    registry.NetworkServiceEndpointRegistryClient
-	nsesMutex                 sync.Mutex
-	nseCancels                map[string]context.CancelFunc
-	retryDelay                time.Duration
-	defaultExpirationDuration time.Duration
+	client                registry.NetworkServiceEndpointRegistryClient
+	nsesMutex             sync.Mutex
+	nseCancels            map[string]context.CancelFunc
+	retryDelay            time.Duration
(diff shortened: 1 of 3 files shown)


pkg/registry/common/refresh/nse_registry_client.go
```diff
@@ -31,15 +31,15 @@ import (
...
+	nsesMutex             sync.Mutex
+	nseCancels            map[string]context.CancelFunc
+	retryDelay            time.Duration
+	defaultExpiryDuration time.Duration
 }
 
-func (c *refreshNSEClient) setDefaultExpiration(duration time.Duration) {
-	c.defaultExpirationDuration = duration
+func (c *refreshNSEClient) setDefaultExpiryDuration(duration time.Duration) {
+	c.defaultExpiryDuration = duration
```
edwarnicke, line 42: Why make this a function rather than doing it inline?

pkg/registry/common/refresh/options.go
```diff
@@ -20,7 +20,7 @@ import "time"
 
 type configurable interface {
```
edwarnicke, line 21: Why an interface here?

pkg/registry/common/refresh/nse_registry_client.go
denis-tingajkin: You are right we can just remove this inline function )

pkg/registry/common/refresh/options.go
denis-tingajkin: I did add this for support server/client but actually pkg refresh has the only client, so we can just remove this interface.

PR #358, opened 2020-07-01
Rework refresh.

edwarnicke:


pkg/networkservice/common/connect/server.go
```diff
@@ -56,14 +56,14 @@ func (c *connectServer) Request(ctx context.Context, request *networkservice.Net
...
+	request.GetConnection().Context = clientConn.Context
+
 	// Carry on with next.Server
 	conn, err := next.Server(ctx).Request(ctx, request)
 	if err != nil {
 		return nil, err
 	}
-	// Copy Context and Path from client to response from server
-	conn.Context = clientConn.Context
-	conn.Path = clientConn.Path
```
denis-tingajkin, line 66: Interesting... Why we don't need to set a path?

pkg/networkservice/common/refresh/client.go
```diff
@@ -55,76 +54,67 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
...
-				timer.Stop()
-			}
-			if canceller, ok := t.refreshCancellers[id]; ok {
-				canceller()
-			}
-			// add refresh context to request context
-			refreshCtx, cancelFunc := context.WithCancel(context.Background())
-			newCtx := withRefreshContext(ctx, refreshCtx)
+		// Create refreshRequest
+		refreshRequest := request.Clone()
```
denis-tingajkin, line 66: I think to avoid potential race condition we need to move this cloning out of async exec section.

pkg/networkservice/common/connect/server.go
edwarnicke: You are correct, we should set the path.  Fixed!

pkg/networkservice/common/refresh/client.go
edwarnicke: You are correct, fixed in update :)

pkg/networkservice/common/refresh/client.go
```diff
@@ -55,76 +54,67 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
...
+			})
+		})
+		t.timers[request.GetConnection().GetId()] = timer
 	})
 	return rv, nil
 }
 
 func (t *refreshClient) Close(ctx context.Context, conn *networkservice.Connection, _ ...grpc.CallOption) (*empty.Empty, error) {
 	t.executor.AsyncExec(func() {
-		if timer, ok := t.connectionTimers[conn.GetId()]; ok {
```
haiodo, line 100: @edwarnicke I think we need to call cancel for refreshCtx, for more fast stop of Requesting in case of Close. 

Also possible we need to check on Close if Request is already in progress and wait until it will be canceled. Thoughts?

edwarnicke: If we weren’t in the executor you’d be right.  But because the executor guarantees exclusivity and order of execution, either 

1.  This invocation in Close comes first, in which case the timer is stopped and the timers invocation of executor never happens
2. The timer’s execution of executor happens first *and it completes* before we get to the Closes executor invocation here.  

Either way, executing cancel has no effect.

edwarnicke: No wait, you are correct, we could race executor is called, and the timer’s use could get in line while we are Closing.  Will fix.

pkg/networkservice/common/refresh/client.go
```diff
@@ -55,76 +56,72 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
...
+				if ok && currentTimer != timer {
+					cancel()
+					return
+				}
+				// Resend request if the refreshCtx is not Done.  Note: refreshCtx should only be done if
+				// t.ctx has been canceled (usually because the clientCtx has been canceled
+				select {
+				case <-refreshCtx.Done():
+				default:
+					if _, err := t.Request(refreshCtx, refreshRequest, opts...); err != nil {
```
haiodo, line 100: Do we block all other requests by doing Request inside executor?

edwarnicke: Good catch.  Fixed!

pkg/networkservice/common/refresh/client.go
edwarnicke: Fixed.

PR #359, opened 2020-07-01
Rework heal.

edwarnicke:
Heal had a few issues:

1.  It was doing an expensive init in the constructor
2.  The eventLoop could be clearer and simpler
3.  There unecessary information being kept at struct{} level
4.  It was attempting to Close things after the clientContext was Done.
    This is pretty much guaranteed to fail, as when the clientContext is Done, so typically is
    the grpc.ClientConn for the underlying connection.  If we want to handle graceful Close
    when shutting down a client, we probably need to do that in a separate chain element.
    This is why closers were removed and TestHealClient_MonitorClose was deleted.

### DIFF
--- a/pkg/networkservice/common/heal/client.go
+++ b/pkg/networkservice/common/heal/client.go
@@ -19,10 +19,11 @@ package heal
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/golang/protobuf/ptypes/empty"
-	"github.com/sirupsen/logrus"
+	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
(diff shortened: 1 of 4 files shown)


pkg/networkservice/common/heal/client.go
```diff
@@ -62,95 +61,86 @@ type healClient struct {
...
+					delete(f.connections, conn.GetId())
 				}
 			}
 
-			for id, request := range f.requestors {
-				if _, ok := f.reported[id]; !ok {
-					request()
+			for id, heal := range f.heals {
+				if _, ok := f.connections[id]; !ok {
+					heal()
```
haiodo, line 135: @edwarnicke I do not think it is a good idea to do a heal on every update for every connection, 
Assume just one is added we receive update, and we do a bunch of requests for any of others.

I think we need to heal only on delete.

Update's case fo monorepo will propogate event to previous client, and in case of context changes( DNS change for example), every client was able to immediately react, but now we will need to wait, so for Updates it is also doesn't make seance to call Heal.

Initial transfer also not require to call heal, if and only if we do not have connections on received list.

Thoughts?

edwarnicke: I agree that healing every connection on every update would not be a good idea.  This doesn't do that.

f.connections reflects what has been reported to us by our peer.
f.heals reflects what we believe we have in terms of connections.

When we receive an update, act on f.connections to have it reflect that update.

At the end, we go through the connections we think we have (f.heals) and 'heal' the ones that have not been reported to us in f.connections (note the !ok in the if statement)

If our peer reports an INITIAL_STATE transfer and its missing a connection we feel we should have, we heal that.
If our peer DELETEs a connection, we heal that.

Does that make sense?

haiodo: Ok, sounds good.

Issue #370, opened 2020-07-07
Interdomain: Add registry dnsresolve chain element to resolve passed domain to clienturl


denis-tingajkin:
## Context

#235

## Description

To add possible to connect proxy registry to remote registry we need to put into context URL to the remote target. 

## Solution

Add `dnsresolve` chain element that's simply will resolve passed domain to URL via DNS and pass it into context.

PR #371, opened 2020-07-07
Interdomain: Add registry dnsresolve chain element

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/370

### DIFF
--- a/pkg/registry/common/dnsresolve/common.go
+++ b/pkg/registry/common/dnsresolve/common.go
@@ -66,13 +66,4 @@ func resolveDomain(ctx context.Context, service, domain string, r Resolver) (*ur
 	return u, nil
 }
 
-func domain(ctx context.Context) (string, error) {
-	if v := ctx.Value(domainKey); v != nil {
-		if val, ok := v.(string); ok {
-			return val, nil
-		}
-	}
-	return "", errors.New("domain to resolve has not passed into context")
-}
-
 var _ Resolver = (*net.Resolver)(nil)
--- a/pkg/registry/common/dnsresolve/common.go
+++ b/pkg/registry/common/dnsresolve/common.go
@@ -1,0 +1,86 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dnsresolve
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+)
+
+// Domain represents wanted domain and service
+type Domain struct {
+	// Name is name of wanted domain
+	Name string
+	// Service is name of wanted service
+	Service string
+}
+
+// Resolver is DNS resolver
+type Resolver interface {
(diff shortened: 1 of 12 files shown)


pkg/registry/common/dnsresolve/common.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+)
+
+// Domain represents wanted domain and service
+type Domain struct {
```
edwarnicke, line 28: Why are we combining the domain and service here?  Also wouldn't that no longer be the *domain* ?  Why not just pass the network service name "service@domain" ?

denis-tingajkin: @edwarnicke Service here is not network service, it is service to lookup SRV record :)

edwarnicke: Could you say more, I don't quite follow.

denis-tingajkin: https://github.com/networkservicemesh/sdk/pull/371/files#diff-c9d8b841fcd143230a5ca28860c6e096R41

denis-tingajkin: Take a look at service parameter of SRV record format https://en.wikipedia.org/wiki/SRV_record

pkg/registry/common/dnsresolve/common.go
```diff
@@ -0,0 +1,86 @@
...
+	// Name is name of wanted domain
+	Name string
+	// Service is name of wanted service
+	Service string
+}
+
+// Resolver is DNS resolver
+type Resolver interface {
```
edwarnicke, line 36: Why do we have a new Resolver interface, why not just use the standard net.* ones?

edwarnicke: Ah, I see... testing shims...

denis-tingajkin: @edwarnicke Note: standard net.* provides structures

pkg/registry/common/dnsresolve/common.go
denis-tingajkin: In k8s Service is `Service.Spec.Ports.Name` field. For example in k8s example of service can be `svc.proxy.registry` and so on.

edwarnicke: OK... I'm not quite following why we are passing this through the context.  We can extract the domain from the Connection.NetworkServiceName directly... and we should probably pick a common SRV.Service name... what were your thoughts there?

denis-tingajkin: >nd we should probably pick a common SRV.Service name... what were your thoughts there?

I think yes, we can add some constant for SRV.Service name for registries. In this case, the struct Domain will be deleted.

denis-tingajkin: @edwarnicke should we put SRC.Service name for registries to API?

edwarnicke: Question... why couldn't it simply be a constant in the resolver package?

denis-tingajkin: I thought that it can be used by API users, finally fixed.

pkg/registry/common/dnsresolve/context.go
```diff
@@ -0,0 +1,30 @@
...
+import "context"
+
+type contextKey string
+
+const (
+	domainKey contextKey = "dnsresolve.Domain"
+)
+
+// WithDomain put wanted domain into context
+func WithDomain(ctx context.Context, domain string) context.Context {
```
edwarnicke, line 28: I still don't understand why we are putting this in the context instead of taking it from the request...

denis-tingajkin: I thought this approach is more scalable but ok, let's just get the domain of query from request. If dnsresolve will be used somewhere else we can return to WithDomain approach :)

denis-tingajkin: Finally fixed.

Issue #362, opened 2020-07-03
Interdomain: add registry connect chain element


denis-tingajkin:
## Context

https://github.com/networkservicemesh/sdk/issues/235

## Description

Proxy registry should be able to connect to the registry from another domain by the domain name. 

## Solution

Add analog of chain element [connect](https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/connect/server.go#L53) for registries.  

Example of usage:

1. Proxy registry receives  query with network service name = `ns-1@my.domain`.
2. Proxy registry resolves domain my.domain to IP.
3. Proxy registry connects to the next registry by IP from step2.

edwarnicke:
I’d recommend breaking it into two chain elements: one that resolves ns-1@my.domain to and IP, and another that does the connect behavior.  That way we can start with the simple ‘look up a SRV record in DNS approach’ ... and if folks have other needs they only have to replace the chain element that does the resolutions.

denis-tingajkin:
@edwarnicke it is a great idea! Will do, thanks

PR #374, opened 2020-07-08
Interdomain: Add registry connect chain element

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/362


pkg/registry/common/connect/nse_server.go
```diff
@@ -0,0 +1,132 @@
...
+	if v, ok := c.cache[key]; ok && v != nil {
+		v.expiryTime = time.Now().Add(c.connectExpiration)
+		return v.client
+	}
+	r := clienturl.NewNetworkServiceEndpointRegistryClient(ctx, c.clientFactory, c.dialOptions...)
+	c.cache[key] = &nseCacheEntry{client: r, expiryTime: time.Now().Add(c.connectExpiration)}
+	return r
+}
+
+func (c *connectNSEServer) setExpirationDuration(d time.Duration) {
```
edwarnicke, line 124: Why not just set these directly inline rather than making new functions that simply obscure what's being done?

pkg/registry/common/connect/nse_server.go
```diff
@@ -0,0 +1,132 @@
...
+	c.connectExpiration = d
+}
+
+func (c *connectNSEServer) setClientDialOptions(opts []grpc.DialOption) {
```
edwarnicke, line 128: Why not just set these directly inline rather than making new functions that simply obscure what's being done?

pkg/registry/common/connect/nse_server.go
```diff
@@ -0,0 +1,132 @@
...
+	c.dialOptions = opts
+}
+
+var _ registry.NetworkServiceEndpointRegistryServer = (*connectNSEServer)(nil)
```
edwarnicke, line 132: I'm super curious what this line is doing :)

pkg/registry/common/connect/nse_server.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type connectNSEServer struct {
+	dialOptions       []grpc.DialOption
+	clientFactory     func(ctx context.Context, cc grpc.ClientConnInterface) registry.NetworkServiceEndpointRegistryClient
+	cache             map[string]*nseCacheEntry
+	cacheLock         sync.Mutex
+	ctx               context.Context
+	once              sync.Once
+	connectExpiration time.Duration
```
edwarnicke, line 44: Why a connectionExperirationTimer?

pkg/registry/common/connect/nse_server.go
```diff
@@ -0,0 +1,132 @@
...
+type nseCacheEntry struct {
+	expiryTime time.Time
+	client     registry.NetworkServiceEndpointRegistryClient
+}
+
+type connectNSEServer struct {
+	dialOptions       []grpc.DialOption
+	clientFactory     func(ctx context.Context, cc grpc.ClientConnInterface) registry.NetworkServiceEndpointRegistryClient
+	cache             map[string]*nseCacheEntry
+	cacheLock         sync.Mutex
```
edwarnicke, line 41: Why global locks rather than a typed sync.Map?

pkg/tools/grpcutils/health.go
```diff
@@ -33,3 +36,22 @@ func RegisterHealthServices(s *grpc.Server, services ...interface{}) {
 		}
 	}
 }
+
+// WaitStatusServing wait status serving for each passed service
+func WaitStatusServing(ctx context.Context, client grpc_health_v1.HealthClient, services ...interface{}) error {
```
edwarnicke, line 41: Why not just use [grpc.WaitForReady](https://godoc.org/google.golang.org/grpc#WaitForReady) ?

pkg/registry/common/connect/nse_server.go
denis-tingajkin: It is interface compliance checker https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

pkg/registry/common/connect/nse_server.go
denis-tingajkin: It is needed to allow use the same options for ns/nse servers. In the case of inline, we'll need to create duplicated options and it will look something like that
```
NewNetworkServiceRegistryServer(..., connect.WithNetworkServiceClientDialOptions(..))
```
So to prevent this complication, I've added these functions.

pkg/registry/common/connect/nse_server.go
denis-tingajkin: https://github.com/networkservicemesh/sdk/pull/374#discussion_r451930392

pkg/registry/common/connect/nse_server.go
edwarnicke: OK... is there an event driven rather than timer driven option we could use to decide whether or not to retain connections?

pkg/tools/grpcutils/health.go
denis-tingajkin: his util method can be used for k8s probes, but actually, in this PR we can simply use grpc.WaitForReady. 
Fixed.

pkg/registry/common/connect/nse_server.go
denis-tingajkin: I've used executor for synchronization here because an executor can kill goroutine in the case when all events completed.

pkg/registry/common/connect/nse_server.go
denis-tingajkin: https://github.com/networkservicemesh/sdk/pull/374#discussion_r452108865

pkg/registry/common/connect/nse_server.go
denis-tingajkin: I like idea of simply store connection during some period while it is using.

edwarnicke: So here's the question... what triggers needing a client and what triggers no longer needing it?

For example... in NetworkService connect, that's driven by whether or not we have active NetworkService.Connections to a particular client.   In other words, do we have active interests with that client.  What is the 'active' interest here?

edwarnicke: > It is needed to allow use the same options for ns/nse servers. In the case of inline, we'll need to create duplicated options and it will look something like that
> 
> ```
> NewNetworkServiceRegistryServer(..., connect.WithNetworkServiceClientDialOptions(..))
> NewNetworkServiceEndpointRegistryServer(..., connect.WithNetworkServiceEndpointClientDialOptions(..))
> ```
> 
> So to prevent this complication, I've added these functions.

Also... I don't see why Options requires having these tiny one line functions that obscure what's happening attached to the connectNSEServer... perhaps I'm missing something.

pkg/registry/common/connect/nse_server.go
edwarnicke: But you still have a cacheLock mutex doing global locking... that's the part I don't understand.  Why should we have *any* global locks here... there are tools that allow us to avoid them.

pkg/registry/common/connect/nse_server.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (c *connectNSEServer) connect(ctx context.Context) registry.NetworkServiceEndpointRegistryClient {
+	defer c.executor.AsyncExec(c.cleanup)
+	key := ""
+	if url := clienturl.ClientURL(ctx); url != nil {
+		key = url.String()
+	}
+	var result registry.NetworkServiceEndpointRegistryClient
+	<-c.executor.AsyncExec(func() {
```
edwarnicke, line 96: OK... we've moved the global synchronicity to an executor... but its still present.  This could be done similarly to how NetworkService connect does it *without* hitting global locks on every call.

denis-tingajkin: The main difference between NetworkService connect and registry connect is that registry connect can not just know when the registry can delete a client to free resources. So that's why we need to use some strategy for finding and cleaning outdated resources. So, in this case, we need to use a `lock` because of finding and cleaning should be covered by a critical section to prevent some extrimal race conditions. In this case, executor looks good to me because it has smart goroutine spawning behavior.

pkg/registry/common/connect/nse_server.go
denis-tingajkin: > What is the 'active' interest here?

I think it is `active` while the client is interested to use it.

denis-tingajkin: >Also... I don't see why Options requires having these tiny one line functions that obscure what's happening attached to the connectNSEServer... perhaps I'm missing something.

Hm... Probably it is correct we can just pass dial options and expiration duration via function parameters :)

pkg/registry/common/connect/nse_server.go
denis-tingajkin: https://github.com/networkservicemesh/sdk/pull/374#discussion_r452300324

pkg/registry/common/connect/nse_server.go
edwarnicke: As a mechanical matter, that would be a better approach, definitely, but the question fundamentally remains: what constitutes functionally the meaningful lifetime of needing that connection?  My point is, I am pretty sure that's *event* not *time* driven.

pkg/registry/common/connect/nse_server.go
edwarnicke: Don't registry entries have an expireTime?

Does the critical section need to be scoped globally?

denis-tingajkin: > Don't registry entries have an expireTime?

Yes NSE has expiration_time but how you plan to use it here? 

For example, the user doing these steps:

```
1. Find nse1@domain1
2. Register nse2@floating.domain
3. Register nse3@floating.domain
4. Unregister nse4@domain1
```

So here we have two clients for _floating.domain_ and for _domain1_. Question: what the strategy you are seeing to use here?

denis-tingajkin: > Does the critical section need to be scoped globally?

Not sure that pattern from networkservice.connect is suitable here because we need to do two things in one critical section:
1. Iterate whole map of clients
2. Delete expired clients

edwarnicke: Or just use time.Timer :)  Reduces the scope dramatically :)

pkg/registry/common/connect/nse_server.go
denis-tingajkin: > My point is, I am pretty sure that's event not time driven.

But here is many things that can be not controlled for example in time _t1_ result of DNS query "domain1" can be url1 in time _t2_ result of DNS query "domain1" can be url2 and so on.

pkg/registry/common/connect/nse_server.go
denis-tingajkin: > Or just use time.Timer :) Reduces the scope dramatically :)

With timers, we have the same issue:
```
timer tick:
1. Iterate whole map of clients
<---race condition here for sync map
2. Delete expired clients
```
```
Register/Find/Unregister:
1. Lookup in cache
```
Also, timers will take some goroutine when in the case when connect is not used executor can kill goroutine and do not use machine resources to check empty map :)

pkg/registry/common/connect/nse_server.go
denis-tingajkin: Finally, options are removed and it looks more clear, thanks :)

pkg/registry/common/connect/nse_server.go
edwarnicke: Each Registry Entry we are tracking has its own expirationTimer... and could thus have its own timer.  The *event* being the expiration of the Registry entry.  If you track things similar to how we handle connections in Network Service connect, you should be able to avoid global locks, and global traversals.

denis-tingajkin: @edwarnicke ok, how you want to handle the case when user Finds NSE with expiration_time = 1 year? Do you suggest never a a close connection in this case?

denis-tingajkin: Finally should be fixed.

pkg/registry/common/connect/ns_server.go
```diff
@@ -0,0 +1,90 @@
...
+type nsCacheEntry struct {
+	expirationTimer *time.Timer
+	client          registry.NetworkServiceRegistryClient
+}
+
+type connectNSServer struct {
+	dialOptions       []grpc.DialOption
+	clientFactory     func(ctx context.Context, cc grpc.ClientConnInterface) registry.NetworkServiceRegistryClient
+	cache             nsClientMap
+	connectExpiration time.Duration
```
edwarnicke, line 40: Why are we setting a single arbitrary global timer here rather than taking the timing information from the events coming in.  When we proxy registration, that registration has an expiration timer, when we do a find, the responses have expiration timers... why introduce arbitrary new global ones into the system?

denis-tingajkin: > Why introduce arbitrary new global ones into the system?

Because if we got for example from domain1 NSE with expiration_time = 1 year, then we'll never close connection...

denis-tingajkin: @edwarnicke In short

### Problems of way to use expiration time of resource

## 1. Resources with long expiration time.

As I said above few times, if connect got NSE/NS with expiration_time > 1 year and so on then we'll never close the connection. The worst thing of this that the connection may be not even used...


## 2. The complex system of monitoring

By this point I mean the cases when we have two or more resources from one domain with different expiration time... Image the case:
User did these steps:
```
1. Register NSE `nse-2` with expiration_time 1 minute
2. Register NSE `nse-3` with expiration_time 20 minutes 
3. Register NSE `nse-1` with expiration_time 1 day // this will be selected as expiration time of connection
4. Unregister NSE `nse-1` 
```
In this case, we'll need to select the next expiration time  (20 minutes) and for this, we'll need to integrate the whole list of records... And so on cases.

## 3. Leak of connections

Let's imagine the next case: 

1. User requests `ns-1@domain1` in time t1.
2. DNS result of domain1 maybe url1 
3. Registry `connect` add to cache url1 and sets expiration time = ns-1.expiration_time(for example 1day)
4. User requests `ns-2@domain1`  in time t2.
5. DNS result of domain1 in this time maybe url2
6. Registry `connect` add to cache url2 and sets expiration time = ns-2.expiration_time(for example 1day) //!!! We have two connections when one of them may be not using!

**Actual**: We have 2 connections when one connection is not used more.
**Expected**: Do not need unused connections.

###  Solution

Use the map of connection where expiration time depends on user interested. (User doing many requests to domain1 OK store connection to domain1 in the cache. User did one time connect to domain2 and do not use this connection OK lets just remove it and so on)

@edwarnicke Please let me know If I missing something.

edwarnicke: Good point on long expire times on resources.  I like your proposed solution in principle (somehow detecting user interest to determine how long to hang onto a connection), what in concrete terms did you have in mind?  How do we detect user interest in a thing?

edwarnicke: Actually... wasn't part of the point of monitor to 'monitor' registered resources so we could quickly reregister if they disappear?

denis-tingajkin: Can we just use a current approach that if a connection is loaded from a cache then just update delete timer 
 https://github.com/networkservicemesh/sdk/pull/374/files#diff-c4e8f787c3f17d8f27d19a98904cf44bR76? :)

edwarnicke: Hmm... could we move the connExpiration timer to a constant rather than a parameter the user has to think about?

denis-tingajkin: In this case, we'll need to return to options because this parameter should be modifiable for testing goals.

edwarnicke: Options are fine :)

pkg/registry/common/connect/option.go
```diff
@@ -0,0 +1,53 @@
...
+
+// WithExpirationDuration sets expiration time for each connection
+func WithExpirationDuration(duration time.Duration) Option {
+	return applyOptionFunc(func(c configurable) {
+		c.setExpirationDuration(duration)
+	})
+}
+
+// WithClientDialOptions sets specific dial options for each created client
+func WithClientDialOptions(opts ...grpc.DialOption) Option {
```
denis-tingajkin, line 49: @edwarnicke Note: I've moved DialOption to option here to concatenate two options in New*Server() functions to prevent this function signature
```
fund New*Server(options []Option, dialOptions... grpc.DialOption)
```

pkg/registry/common/connect/ns_server.go
denis-tingajkin: Done :)

Issue #349, opened 2020-06-29
Interdomain: add registry chain element to proxying registration/find/unregistration to next registry


denis-tingajkin:
## Context

https://github.com/networkservicemesh/sdk/issues/235

## Description

We need to add possible to registries proxy all incoming **registrations**/**unregistratiom**/**find** to the next registry. 

### Use-case:

1. The registry receives registration with a special symbol in the name "@" or special label in labels.
2. The registry just proxying this traffic to the next registry.

## Questions:

Do we want to use special symbol "@" in ns/nse name to mark it as interdomain? As an alternative variant, we can add special labels for nse. For example:

```
nse:
  name:
  labels:
   icmp-responder:
     k:v
   interdomain:
     k:v
```

edwarnicke:
Simplistcally, yes, the @ means its coming from another registry, and so while we may *cache* information about it, we are never authoritative.  It is probably helpful however to think of *any* registry entry as having a network service mesh domain.

For example, the network service mesh domain of our local cluster may be named with a cluster name of some sort (cluster1.example.com for example).  Local users could skip the @cluster1.example.com, but the local registry is still for a domain, and if anyone outside the cluster is to be able to access it, it will need to have a name for that domain.

edwarnicke:
There are some other ideas which may be interesting to think about.  You could imagine, even if our cluster local domain was never accessible outside the cluster, that you could think of it as @local.  Its also possible that you might choose to run multiple network service mesh domains in a single cluster, scoped per K8s namespace, so you might @my-namespace.nsm.local or @my-namespace.nsm.${clustername}.

Please note, all of this is 'things to think about' not 'directions to go do a thing'.  Leave space for such things, but don't let them impede your progress :)

PR #381, opened 2020-07-13
Add proxy registry chain element

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/349

This is the last chain element for `proxy-registry-dns`

### DIFF
--- a/pkg/registry/common/connect/ns_server_test.go
+++ b/pkg/registry/common/connect/ns_server_test.go
@@ -20,11 +20,14 @@ import (
 	"context"
 	"net"
 	"net/url"
+	"runtime"
 	"testing"
+	"time"
+
+	"go.uber.org/goleak"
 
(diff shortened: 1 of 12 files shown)


pkg/registry/common/connect/ns_server_test.go
```diff
@@ -82,4 +83,12 @@ func TestConnect_NewNetworkServiceRegistryServer(t *testing.T) {
 	}}, findSrv)
 	require.Nil(t, err)
 	require.Equal(t, (<-ch).Name, "ns-1-1")
+
+	closeServer1()
+	closeServer2()
+
+	require.Eventually(t, func() bool {
+		runtime.GC()
```
denis-tingajkin, line 91: Note: registry/clienturl is using finalization. So we need to make sure that GC was called to avoid goroutine leaks in tests.

PR #387, opened 2020-07-13
Added: simple indexed round-robin algorithm

valeriy-simakov:
Signed-off-by: Valeriy Simakov <simakov.valeriy@gmail.com>
Simple round-robin algorithm and selector of any items via any algorithms

### DIFF
--- a/pkg/tools/algorithm/roundrobin/indexed_decider.go
+++ b/pkg/tools/algorithm/roundrobin/indexed_decider.go
@@ -1,0 +1,35 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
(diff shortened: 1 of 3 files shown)


denis-tingajkin:
Probably we need something more simple for example:

```
package roundrobin

import "sync"

type Roundrobin struct {
    index int32
    size int32
    mutex sync.Mutex
}

func New(size int) *Roundrobin {
	return &Roundrobin{
		size: int32(size),
    	}
}

func (r *Roundrobin) Next() int {
	r.mutex .Lock()
	defer r.mutex .Unlock()
	r.index = (r.index + 1) % r.size
	println(r.index)
	return int(r.index)
}
```

@edwarnicke , @haiodo  what do you think?

denis-tingajkin:
We've found that actually we do not need balancing algorithm for siov based nse

PR #390, opened 2020-07-15
Fix TestHealClient_Request

haiodo:
- remove Eventualy with a conext and channel wait for 500ms max.

### DIFF
--- a/pkg/networkservice/common/heal/client_test.go
+++ b/pkg/networkservice/common/heal/client_test.go
@@ -37,8 +37,9 @@ import (
 )
 
 const (
-	waitForTimeout = 50 * time.Millisecond
-	tickTimeout    = 10 * time.Millisecond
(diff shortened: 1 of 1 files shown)


pkg/networkservice/common/heal/client_test.go
```diff
@@ -104,16 +106,17 @@ func TestHealClient_Request(t *testing.T) {
...
+	select {
+	case <-ctx.Done():
+		require.FailNow(t, "timeout waiting for Heal event %v", time.Since(t1))
+		return
+	case <-onHealCh:
+		// All is fine, test is passed
+		break
 	}
-	require.Eventually(t, cond, waitForTimeout, tickTimeout)
+	logrus.Infof("passed with total time: %v", time.Since(t1))
```
edwarnicke, line 119: Interesting... this feels like it should be the same as require.Eventually... is there a known problem with require.Eventually?

PR #391, opened 2020-07-16
PoC: Do not use "Preemptive Interface Anti-Pattern" in chain elements

denis-tingaikin:

This PR shows the advantages of returning structs instead of interfaces for our chain elements.


pkg/networkservice/common/localbypass/server.go
```diff
@@ -35,48 +34,40 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 )
 
-type localBypassServer struct {
+// Server is  NetworkServiceServer chain element that tracks local Endpoints and substitutes their unix file socket as the clienturl.ClientURL(ctx) used to connect to them.
+type Server struct {
```
haiodo, line 38: I'd like idea actually to have it named Server.

edwarnicke:
The reasoning behind returning interfaces and not structs is:

1.  It precludes interface creep… structs can have more than just the NetworkServiceServer etc interfaces
2. Structs would allows things to depend on those special magic interfaces

Suddenly, instead of having something which can work with any NetworkServiceServer… its welded to a struct as a particular implementation of it.

Take the localbypass example you gave here.  Say I needed to implement a different localbypass for some reason.  Maybe its for Windows.  Maybe its for a non-K8s environment.  Every chain element in the system that depends on your localbypass struct is a barrier to doing so.  They make it over time impossible to actually swap in alternate chain elements (or eliminate them entirely).

denis-tingajkin:
> Take the localbypass example you gave here. Say I needed to implement a different localbypass for some reason. Maybe its for Windows. Maybe its for a non-K8s environment. Every chain element in the system that depends on your localbypass struct is a barrier to doing so. They make it over time impossible to actually swap in alternate chain elements (or eliminate them entirely).

Yes, here you are right. I think to avoid this problem we should use only interfaces for consumers. For example, for your example consumer of localbypass should use networkservice.NetworkServiceServer interface to avoid these kind problems.

denis-tingajkin:
In short, my point here is: chain elements can return struct, but all consumers of chain elements should use strict interface.

denis-tingaikin:
Finally, I think it can be closed

PR #394, opened 2020-07-17
Fix mechanismsServer next chain element call

lvfxx:
Implementation of #402 
Previously, in the one situation mechanismsServer called `next.Server(ctx).Request/Close` itself, but all of the each mechanism Servers called `next.Server...` too. Now mechanismsServer simply returns `conn` from selected mechanism Server request

### DIFF
--- a/pkg/networkservice/common/mechanisms/server.go
+++ b/pkg/networkservice/common/mechanisms/server.go
@@ -26,7 +26,6 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/chain"
-	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/trace"
 )
 
@@ -56,12 +55,7 @@ func (m *mechanismsServer) Request(ctx context.Context, request *networkservice.
 	if request.GetConnection().GetMechanism() != nil {
 		srv, ok := m.mechanisms[request.GetConnection().GetMechanism().GetType()]
 		if ok {
-			conn, err := srv.Request(ctx, request)
-			if err != nil {
-				return nil, err
-			}
-			request.Connection = conn
-			return next.Server(ctx).Request(ctx, request)
+			return srv.Request(ctx, request)
 		}
 		return nil, errors.Errorf("Unsupported Mechanism: %+v", request.GetConnection().GetMechanism())
 	}
--- a/pkg/networkservice/common/mechanisms/server_test.go
+++ b/pkg/networkservice/common/mechanisms/server_test.go
@@ -21,6 +21,10 @@ import (
 	"io/ioutil"
 	"testing"
 
+	"github.com/golang/protobuf/ptypes/empty"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 	"github.com/networkservicemesh/api/pkg/api/networkservice/mechanisms/cls"
 	"github.com/networkservicemesh/api/pkg/api/networkservice/mechanisms/kernel"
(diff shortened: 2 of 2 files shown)


pkg/networkservice/common/mechanisms/server.go
```diff
@@ -55,22 +54,25 @@ func NewServer(mechanisms map[string]networkservice.NetworkServiceServer) networ
...
+			return nil, errors.Errorf("Unsupported Mechanism: %+v", request.GetConnection().GetMechanism())
 		}
-		return nil, errors.Errorf("Unsupported Mechanism: %+v", request.GetConnection().GetMechanism())
+		conn, err := srv.Request(ctx, request)
+		if err == nil {
+			return conn, nil
+		}
+		// if err != nil try any available mechanism below
 	}
 	for _, mechanism := range request.GetMechanismPreferences() {
```
haiodo, line 66: Let's ignore already failed mechanism type.

denis-tingajkin:
@lvfxx Can you please also add an issue for this fix to add clarification about the bug for other committers

pkg/networkservice/common/mechanisms/server.go
denis-tingajkin: I think we also need to wrap errors and if no one mechanism server did not return positive answer then need to return this wrapped error :)

edwarnicke: @denis-tingajkin What are you hoping to get from wrapping here?  I ask because double or triple wrapping can lead to multiple stack traces making it hard to follow.  Say if the error occured at:

Frame1
Frame2
...
Frame20
and was wrapped or withstacked traced... and then at Frame13, we wrap it again, and at Frame2 we wrap it again, then you will get when you %+v it:

ErrMsgFrame2
Frame2
Frame1
ErrMsgFrame13
Frame13
Frame12
...
Frame2
Frame1
ErrMsgFrame20
Frame20
Frame19
...
Frame 13
...
Frame2
Frame1

it gets super super hard to see what's happening... generally wrapping once is usually sufficient, as close to origin of the error from a non-wrapping regime as possible.

edwarnicke:
@lvfxx I think at this time the mechanism servers at least in sdk-vppagent are doing the next.{Server,Client} calls... all this is doing is selecting *which* of them to use here.  I believe therefore that we don't need to call next.{Server,Client} here as its already been called by the mechanism chain elements themselves.  Happy to hear your thought though :)

pkg/networkservice/common/mechanisms/server.go
denis-tingajkin: @edwarnicke makes sense.

I think then we can simply log each negative answer via `log.Error(err.Error())`

lvfxx:
> @lvfxx I think at this time the mechanism servers at least in sdk-vppagent are doing the next.{Server,Client} calls... all this is doing is selecting _which_ of them to use here. I believe therefore that we don't need to call next.{Server,Client} here as its already been called by the mechanism chain elements themselves. Happy to hear your thought though :)

@edwarnicke Now this PR changes two things:
1) Removes `next.{Server,Client}` calls from the select mechanism chain element
2) If select mechanism chain element gets NetworkServiceRequest with filled Connection.Mechanism and attempt to call Request on that mechanism server returns error now element will try to connect via any other supported mechanism (previously it simply returned error)

pkg/networkservice/common/mechanisms/server.go
```diff
@@ -56,12 +55,7 @@ func (m *mechanismsServer) Request(ctx context.Context, request *networkservice.
 	if request.GetConnection().GetMechanism() != nil {
 		srv, ok := m.mechanisms[request.GetConnection().GetMechanism().GetType()]
 		if ok {
-			conn, err := srv.Request(ctx, request)
-			if err != nil {
-				return nil, err
-			}
-			request.Connection = conn
-			return next.Server(ctx).Request(ctx, request)
+			return srv.Request(ctx, request)
```
denis-tingajkin, line 58: Can we cover this line by unit test?

lvfxx: Yes, covered it

pkg/networkservice/common/mechanisms/server_test.go
```diff
@@ -38,6 +42,21 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/null"
 )
 
+type countRequestsNetworkServiceServer struct {
```
haiodo, line 45: Please use checkrequest.NewServer

PR #395, opened 2020-07-17
Rework for updatepath/updatetoken

haiodo:
Rework for updatepath/updatetoken

1. setid => updatepath - update connection.Id and add a new path segment if required. Move to next segment if current endpoint name is different.

2. updatepath => updatetoken - just update token value and expire time.


pkg/networkservice/common/updatepath/client.go
```diff
@@ -20,44 +20,35 @@ import (
...
-func NewClient(name string, tokenGenerator token.GeneratorFunc) networkservice.NetworkServiceClient {
-	return &updatePathClient{
-		commonUpdatePath: commonUpdatePath{
-			name:           name,
-			tokenGenerator: tokenGenerator,
-		},
-	}
+// NewClient - creates a new setId client.
+//             name - name of the client
+//             Iff the current pathSegment name != name && pathSegment.id != networkservice.Id, set a new uuid for
```
denis-tingajkin, line 36:
```suggestion
//             If the current pathSegment name != name && pathSegment.id != networkservice.Id, set a new uuid for
```

pkg/networkservice/common/updatetoken/server_test.go
```diff
@@ -0,0 +1,240 @@
...
+	ExpiresProto *timestamp.Timestamp
+}
+
+func (f *updatePathServerSuite) SetupSuite() {
+	f.Token, f.Expires, _ = TokenGenerator(nil)
+	f.ExpiresProto, _ = ptypes.TimestampProto(f.Expires)
+}
+
+func (f *updatePathServerSuite) TestNewServer_EmptyPathInRequest() {
+	t := f.T()
```
denis-tingajkin, line 61: Note: you can use f.Nil(err) and so on

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,77 @@
...
+	if conn.GetPath() == nil || len(conn.GetPath().GetPathSegments()) == 0 {
+		if conn.Path == nil {
+			conn.Path = &networkservice.Path{}
+		}
+		conn.Path.Index = 0
+		// Replace path for first segment.
+		conn.Path.PathSegments = []*networkservice.PathSegment{
+			{
+				Name: segmentName,
+				Id:   conn.Id,
```
edwarnicke, line 50: What if conn.Id hasn't been set elsewhere?  For example for a leaf client... this is taking conn.Id from the request... which is fine in cases like xconnect or nsmgr chains which chain a client from a server... but not correct for general clients that may not have one set at all.

edwarnicke: OK... just read down the server code and see it too is simply passing in the request.GetConnection() to update path... the question is... where are ids being generated?

pkg/networkservice/common/updatepath/client_test.go
```diff
@@ -14,147 +14,129 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package updatepath_test provides tests for updatePathClient and updatePathServer chain elements
+// Package setid_test provides a tests for package 'setid'
```
edwarnicke, line 17: Seems to be mismatched here to the package name.

pkg/networkservice/chains/client/client.go
```diff
@@ -55,11 +56,11 @@ func NewClient(ctx context.Context, name string, onHeal *networkservice.NetworkS
 		append(
 			append([]networkservice.NetworkServiceClient{
 				authorize.NewClient(),
-				setid.NewClient(name),
+				updatepath.NewClient(name),
```
edwarnicke, line 59: I like the rename here :)

pkg/networkservice/common/updatepath/client.go
```diff
@@ -20,44 +20,35 @@ import (
...
 
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
-	"github.com/networkservicemesh/sdk/pkg/tools/token"
 )
 
-type updatePathClient struct {
-	commonUpdatePath
+type idClient struct {
```
edwarnicke, line 30: I'd suggest renaming here to match the new package name.

pkg/networkservice/common/updatepath/client.go
```diff
@@ -20,44 +20,35 @@ import (
...
-// NewClient - creates a NetworkServiceClient chain element to update the Connection.Path
-//             - name - the name of the NetworkServiceClient of which the chain element is part
-func NewClient(name string, tokenGenerator token.GeneratorFunc) networkservice.NetworkServiceClient {
-	return &updatePathClient{
-		commonUpdatePath: commonUpdatePath{
-			name:           name,
-			tokenGenerator: tokenGenerator,
-		},
-	}
+// NewClient - creates a new setId client.
```
edwarnicke, line 34: Comment needs to also shift to reflect the new package name

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,77 @@
...
 	if conn == nil {
-		return errors.New("NetworkServiceRequest.Connection must not be nil")
+		conn = &networkservice.Connection{}
 	}
 
 	// If we don't have a Path, add one
-	if conn.GetPath() == nil {
-		conn.Path = &networkservice.Path{}
+	if conn.GetPath() == nil || len(conn.GetPath().GetPathSegments()) == 0 {
+		if conn.Path == nil {
```
edwarnicke, line 42: Why not just break this out into it's own top level if statement:

```go
if conn.GetPath() == nil {
  conn.Path == &networkservice.Path{}
}
if len(conn.GetPath().GetPathSevments()) == 0 {
...
}
```

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,77 @@
...
+			conn.Path = &networkservice.Path{}
+		}
+		conn.Path.Index = 0
```
edwarnicke, line 45: I believe we can count on conn.Path.Index to be zero as that's its 'zero value'.

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,77 @@
...
+		// Replace path for first segment.
```
edwarnicke, line 46: Are we replacing or setting?

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,77 @@
...
-		authInfo = p.AuthInfo
-	}
-
-	// Generate the tok
-	tok, expireTime, err := u.tokenGenerator(authInfo)
-	if err != nil {
-		return errors.WithStack(err)
+	if nextIndex < len(path.GetPathSegments()) && path.GetPathSegments()[nextIndex].Name != segmentName {
+		// We have next, but name is different
+		return nil, errors.Errorf("NetworkServiceRequest.Connection.Path.PathSegement[%d].Name should be %s, but it is %s",
```
edwarnicke, line 74: This should not be an error condition... this is a typical healing case, it means that someone in the chain has 'healed' to this server because it couldn't get whomever was previously occupying the path.  If I get a path like through pathSegment names like:

[A,B,C,D,E,F] index 2 and I'm G.  The right move is not to return an error to B, but to substitute myself for C in the path so it becomes [A,B,G,D,E,F]... D may or may not accept the substitution... but that's up to D.

pkg/networkservice/common/updatepath/common.go
edwarnicke: OK... just found id generation for conn.Id below for the nextIndex >= case... my underlying point here is: in the starting a new path case, if conn.Id == "" we should generate one.

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,77 @@
...
+	if nextIndex >= len(path.GetPathSegments()) {
+		// Generate new Id for Path segment and connection Id.
+		conn.Id = uuid.New().String()
+		path.PathSegments = append(path.PathSegments, &networkservice.PathSegment{
+			Name: segmentName,
+			Id:   conn.Id,
+		})
+	} else {
+		// Just update ID from next segment.
+		conn.Id = path.GetPathSegments()[nextIndex].Id
```
edwarnicke, line 87: I don't quite follow the logic here, could you explain?

pkg/networkservice/common/updatetoken/client.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+	"google.golang.org/grpc"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/token"
+)
+
+type updatePathClient struct {
```
edwarnicke, line 30: Could we rename these to match the new package name?

pkg/networkservice/common/updatetoken/client.go
```diff
@@ -0,0 +1,61 @@
...
+	tokenGenerator token.GeneratorFunc
+}
+
+// NewClient - creates a NetworkServiceClient chain element to update the Connection.Path
```
edwarnicke, line 34: Probably also want to update the comment.

pkg/networkservice/common/updatetoken/common.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/pkg/errors"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/token"
+)
+
+func updateToken(ctx context.Context, conn *networkservice.Connection, tokenGenerator token.GeneratorFunc) error {
+	// We need a real path if we don't have one
+	if conn == nil {
```
edwarnicke, line 35: Probably better to move this up into the Request case and simply create an empty connection if need be.

pkg/networkservice/common/updatepath/client.go
edwarnicke: I guess the question here is: did @haiodo mean 'if and only if' (often abbreviated iff) and do we think the reader will understand iff as that if he did :)

pkg/networkservice/common/updatetoken/server_test.go
edwarnicke: Or even better: f.NoError(err)

pkg/networkservice/common/updatepath/client.go
haiodo: Thanks. Fixed.

pkg/networkservice/common/updatepath/common.go
haiodo: ok, fixed.

pkg/networkservice/common/updatepath/client.go
haiodo: Fixed.

pkg/networkservice/common/updatepath/client.go
haiodo: Fixed.

pkg/networkservice/common/updatepath/common.go
haiodo: Thanks. fixed.

pkg/networkservice/common/updatepath/common.go
haiodo: Nope, since it could be case Path with wrong index is set and there is no path segments.

pkg/networkservice/common/updatepath/common.go
haiodo: Thanks. fixed.

pkg/networkservice/common/updatepath/common.go
haiodo: This is for case we came second time, in this case current Index are one below, so we need to increment it and set conenction.Id. I've updated code to be more clean.

pkg/networkservice/common/updatetoken/client.go
haiodo: Fixed.

pkg/networkservice/common/updatetoken/client.go
haiodo: Fixed.

pkg/networkservice/common/updatetoken/common.go
haiodo: Moved, but not sure it is better.

pkg/networkservice/common/updatepath/common.go
haiodo: Ok, Fixed.

pkg/networkservice/common/updatepath/common.go
edwarnicke: Good point.

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,76 @@
...
 	}
 
 	path := conn.GetPath()
 
-	// Make sure index isn't out of bound
-	if (len(path.GetPathSegments()) > 0) && int(path.GetIndex()) >= len(path.GetPathSegments()) {
-		return errors.Errorf("NetworkServiceRequest.Connection.Path.Index(%d) >= len(NetworkServiceRequest.Connection.Path.PathSegments)(%d)",
-			path.GetIndex(), len(path.GetPathSegments()))
+	if int(path.Index) < len(path.PathSegments) && path.PathSegments[path.Index].Name == segmentName {
+		// We already in current segment, just update connection Id, no need to increment
```
edwarnicke, line 57: This makes much much more sense now.

pkg/networkservice/common/updatepath/common.go
```diff
@@ -17,74 +17,76 @@
...
+			nextIndex, len(path.GetPathSegments()))
 	}
-
-	// Extract the authInfo:
-	var authInfo credentials.AuthInfo
-	if p, exists := peer.FromContext(ctx); exists {
-		authInfo = p.AuthInfo
+	if nextIndex < len(path.GetPathSegments()) && path.GetPathSegments()[nextIndex].Name != segmentName {
+		// We have next, but name is different, so let's update Path segment in this case.
+		// Id should be same, since next chain elements may depend on it.
```
edwarnicke, line 72: I don't quite follow why the id should be the same in this case.  As I understand it this is the case where we have:

Say we have a chain:

A(id:1) B(id:2) C(id:3) D(id:4)

B goes away, and A heals to G, G sees the entry for B(id:2) ... why should G reuse id:2 ?

haiodo: Because with id:2 it will go to C, and C could have a map of id:2 to some internal structures, and if it will be different, new connection will be established instead, and previous one will be closed by timeout.

edwarnicke: C shouldn't be mapping B's id:2 to an internal structure... it should be tracking any of it state data with its own id:3 for the connection.

haiodo: In this case it should be fine to generate new ID :-)

denis-tingajkin:
@edwarnicke Can we go with this PR?

PR #397, opened 2020-07-17
Round robin

Bolodya1997:
Restart of https://github.com/networkservicemesh/sdk/pull/387


pkg/tools/algorithm/roundrobin/roundrobin_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+package roundrobin
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+type rrtests struct {
```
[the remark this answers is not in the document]
Bolodya1997: Thanks, got it.

pkg/tools/algorithm/roundrobin/roundrobin_test.go
```diff
@@ -0,0 +1,59 @@
...
+	name  string
+	size  []int
+	index []int
+}
+
+var tests = []rrtests{
```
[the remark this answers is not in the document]
Bolodya1997: I've reworked test cases so the algorithm has been changed :)

pkg/networkservice/common/roundrobin/round_robin_selector.go
```diff
@@ -19,35 +19,28 @@
...
+
+	"github.com/networkservicemesh/sdk/pkg/tools/algorithm/roundrobin"
 )
 
 type roundRobinSelector struct {
-	// TODO - replace this with something like a sync.Map that doesn't require us to lock all requests for a simple
-	// selection
-	sync.Mutex
-	roundRobin map[string]int
+	roundRobin *roundRobinMap
```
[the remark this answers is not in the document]
Bolodya1997: Yep

pkg/networkservice/common/roundrobin/round_robin_selector.go
```diff
@@ -19,35 +19,28 @@
...
 }
 
 func newRoundRobinSelector() *roundRobinSelector {
 	return &roundRobinSelector{
-		roundRobin: make(map[string]int),
+		roundRobin: &roundRobinMap{},
```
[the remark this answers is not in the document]
Bolodya1997: Done

pkg/networkservice/common/roundrobin/round_robin_selector.go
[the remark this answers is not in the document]
Bolodya1997: I guess `nsRoundRobins` is going to be OK for this.

pkg/tools/algorithm/roundrobin/roundrobin.go
```diff
@@ -0,0 +1,32 @@
...
+
+// RoundRobin contains index
+type RoundRobin struct {
+	index uint32
+}
+
+// Index returns a new index for the given size
+func (r *RoundRobin) Index(size int) int {
+	return int((atomic.AddUint32(&r.index, 1) - 1) % uint32(size))
+}
```
edwarnicke, line 32: Question... why not just keep this in the roundrobin chain element... its small and elegant and easily localized there :)

Bolodya1997: We want to reuse it in SR-IOV to round robin across the PF/VF.

pkg/tools/algorithm/roundrobin/roundrobin.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// RoundRobin contains index
+type RoundRobin struct {
+	index uint32
+}
+
+// Index returns a new index for the given size
+func (r *RoundRobin) Index(size int) int {
+	return int((atomic.AddUint32(&r.index, 1) - 1) % uint32(size))
```
denis-tingajkin, line 31: What will happen when r.index has achieved uint32 max?

pkg/tools/algorithm/roundrobin/roundrobin.go
```diff
@@ -0,0 +1,32 @@
...
+// Package roundrobin provides a selection by round robins algorithm
+package roundrobin
+
+import (
+	"sync/atomic"
+)
+
+// RoundRobin contains index
+type RoundRobin struct {
+	index uint32
```
denis-tingajkin, line 26: Can we use atomic and store size inside RoundRobin instead of passing it via parameter for function `Index`?
```suggestion
	index uint32
	size uint32
```

pkg/tools/algorithm/roundrobin/roundrobin.go
Bolodya1997: It will be zeroed, so some options can be skipped on the 100500+ iteration.

pkg/tools/algorithm/roundrobin/roundrobin.go
Bolodya1997: If we do it like that, there will be a such a problem to use `RoundRobin` with mutable options list.

pkg/tools/algorithm/roundrobin/roundrobin.go
denis-tingajkin: I'm not sure that it is a good approach.

haiodo:
I think do we really require a round-robin selector, or it should be balance chain element to select a working one with some passed selection alghorithm.

pkg/tools/algorithm/roundrobin/roundrobin.go
Bolodya1997: @edwarnicke 
Was this chain element designed as a load balancer for all new incoming connection requests? Or should it be context-local and contain a simple for-loop to discover an accessible endpoint for a single new incoming request?

Issue #408, opened 2020-07-26
Rewrite registry expire to not use mutexes


edwarnicke:
The expire registry chain element uses mutex, resulting in purely unnecessary contention:

https://github.com/networkservicemesh/sdk/blob/master/pkg/registry/common/expire/nse_server.go#L30

this will result in only one register/unregister being processable at a time, and will also block *all* register/unregister calls when handling expiration:

https://github.com/networkservicemesh/sdk/blob/master/pkg/registry/common/expire/nse_server.go#L70

Better to use a serialize.Executor which would not require blocking by the Register or Unregister calls.

edwarnicke:
I'm also curious why expire is a wrapper and not a chain element :)

denis-tingajkin:
Nice catch, will do! Thanks :)

PR #410, opened 2020-08-01
Rewrite registry expire to not use mutexes

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/408


pkg/registry/common/expire/nse_server.go
```diff
@@ -21,31 +21,37 @@ import (
...
 
 func (n *nseServer) setPeriod(d time.Duration) {
 	n.period = d
 }
 
 func (n *nseServer) Register(ctx context.Context, nse *registry.NetworkServiceEndpoint) (*registry.NetworkServiceEndpoint, error) {
-	n.once.Do(n.monitor)
-	r, err := n.server.Register(ctx, nse)
+	n.once.Do(func() {
+		n.server = next.NetworkServiceEndpointRegistryServer(ctx)
```
edwarnicke, line 45: Super curious why this is in the once.Do ... I think its probably harmless... but curious :)

denis-tingajkin: It is needed to call unregister for the next chain elements when NSE/NS has achieved expiration time.

denis-tingajkin:
I found that many things can be reworked with timers.
TODO: replaces maps to timers.

pkg/registry/common/expire/nse_server.go
edwarnicke: Why not just call Unregister against expire itself?

edwarnicke:
> I found that many things can be reworked with timers.
> TODO: replaces maps to timers.

Two things to keep in mind:

1.  time.Timers are *super* cheap in Go.
2.  time.Timers will not be garbage collected until they either fire or someone calls Stop() ... so if someone unregisters... you need to make sure you Stop() their expire timer :)

pkg/registry/common/expire/nse_server.go
denis-tingajkin: Should be fixed.

PR #420, opened 2020-08-10
Add common FS watcher tool

Bolodya1997:
Add `fswatcher.WatchOn(paths...)`.

### DIFF
--- a/pkg/networkservice/common/excludedprefixes/server.go
+++ b/pkg/networkservice/common/excludedprefixes/server.go
@@ -47,30 +47,15 @@ type excludedPrefixesServer struct {
 
 func (eps *excludedPrefixesServer) init() {
 	logger := trace.Log(eps.ctx)
-	onUpdate := func(_ fsnotify.Event) error {
-		bytes, err := ioutil.ReadFile(filepath.Clean(eps.configPath))
+	onUpdate := func(event fsnotify.Event) error {
+		configPath, err := filepath.Abs(event.Name)
 		if err != nil {
-			logger.Errorf("An error during read file %v, error: %v", eps.configPath, err.Error())
-			return err
+			logger.Warnf("error getting config path from the event, try to use the stored path")
+			return eps.updatePrefixPool(eps.configPath)
 		}
-		source := struct {
-			Prefixes []string
-		}{}
-		err = yaml.Unmarshal(bytes, &source)
-		if err != nil {
-			logger.Errorf("Can not create unmarshal prefixes, err: %v", err.Error())
-			return err
-		}
-		pool, err := prefixpool.New(source.Prefixes...)
-		if err != nil {
-			logger.Errorf("Can not create prefixpool with prefixes: %+v, err: %v", pool.GetPrefixes(), err.Error())
-			return err
-		}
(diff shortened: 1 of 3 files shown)


pkg/tools/fswatcher/watcher.go
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package fswatcher provide common tools for FS watching
+package fswatcher
+
+import "github.com/fsnotify/fsnotify"
+
+// WatchOn creates a FS watcher on given ...paths
+func WatchOn(paths ...string) (*fsnotify.Watcher, error) {
```
denis-tingajkin, line 23: Can we also move [goroutine](https://github.com/networkservicemesh/cmd-forwarder-sriov/blob/5aee0083a1c36299ec55d9d9f887abcbe5b333ae/local/sdk-sriov/pkg/k8s/manager.go#L128) spawn?
```suggestion
func WatchOn(paths ...string) (*fsnotify.Watcher, error) {
```

edwarnicke:
Why not just use fswatcher directly... we are shaving a super super small number of lines at the expense of obscuring what we are doing (using a fairly standard API).

pkg/tools/fswatcher/watcher.go
Bolodya1997: I've moved select logic. Not sure if we want to create goroutine inside of it.

pkg/networkservice/common/excludedprefixes/server.go
```diff
@@ -44,26 +47,35 @@ type excludedPrefixesServer struct {
 
 func (eps *excludedPrefixesServer) init() {
 	logger := trace.Log(eps.ctx)
-	updatePrefixes := func(bytes []byte) {
+	onUpdate := func(_ fsnotify.Event) error {
+		bytes, err := ioutil.ReadFile(filepath.Clean(eps.configPath))
```
denis-tingajkin, line 51: Can we here extract path from the event?

Bolodya1997: Yes, we can get `event.Name` and it should be the path for the file triggered the event.

denis-tingajkin, line 51: Please use `event.Name` because in future excludedPrefixesServer can listen to a few files.

Bolodya1997: Woop! I've tried in and double checked - `event.Name` is a relative path to the file :(
So we really can't do nothing with this...

denis-tingajkin: Do not think so. 
1. You can look at os.Executable...
2. You can look at https://golang.org/src/path/filepath/path.go?s=7016:7053#L232

Bolodya1997: FIxed

denis-tingajkin:
@edwarnicke Do you mind this approach?

denis-tingajkin:
https://github.com/networkservicemesh/sdk/blob/master/pkg/tools/fs/utils.go#L35

PR #423, opened 2020-08-11
First step to including Upstream PR info in downstream commit.

edwarnicke:


edwarnicke:
This PR will create a Commit message that in a followup PR we can start using for pushing downstream

PR #437, opened 2020-08-15
Add workload api spire helper function for fetching sources

denis-tingajkin:

## Motivation

This PR adds helper function for spire workloadapi. The function uses default spire parameters which can be used in each cmd application such as spire socket path, logging from context, selecting svid based on executable name.

### DIFF
--- a/pkg/tools/spire/workloadapi.go
+++ b/pkg/tools/spire/workloadapi.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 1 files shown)


edwarnicke:
I'm a little leery of obscuring the fact we are using the spire API... plus I'm not convinced this is an improvement over the stock:

```golang
	// Get a X509Source
	source, err := workloadapi.NewX509Source(ctx)
	if err != nil {
		logrus.Fatalf("error getting x509 source: %+v", err)
	}
	svid, err := source.GetX509SVID()
	if err != nil {
		logrus.Fatalf("error getting x509 svid: %+v", err)
	}
	logrus.Infof("SVID: %q", svid.ID)
```
which makes it *very* clear what we are doing, and really *wants* to live in main() (where it properly *should*... not buried layers deep in things where it becomes a fixed point in the system).

edwarnicke:
Also... if a decision is going to be made about which SVID to pick... that's a decision that should be made *at main level* ... not deep in library calls.

denis-tingajkin:
@edwarnicke ok, patch moved on cmd level. https://github.com/networkservicemesh/cmd-registry-memory/pull/56

PR #443, opened 2020-08-19
fix: grpcutils.ListenAndServe add a check if previous socket file was deleted successfully

denis-tingajkin:

### DIFF
--- a/pkg/tools/grpcutils/listen_and_serve.go
+++ b/pkg/tools/grpcutils/listen_and_serve.go
@@ -47,7 +47,7 @@ func ListenAndServe(ctx context.Context, address *url.URL, server *grpc.Server) 
 
 	if network == unixScheme {
 		err := os.Remove(target)
-		if !errors.Is(err, os.ErrNotExist) {
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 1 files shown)


haiodo (approved):

pkg/tools/grpcutils/listen_and_serve.go
```diff
@@ -47,7 +47,7 @@ func ListenAndServe(ctx context.Context, address *url.URL, server *grpc.Server)
 
 	if network == unixScheme {
 		err := os.Remove(target)
-		if !errors.Is(err, os.ErrNotExist) {
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
```
edwarnicke, line 50: Have we hit this issue in practice in the multi-repo code?

denis-tingajkin: Yes, the socket file can exist on the host file system, but actually is not using. This patch related to another thing related to case if the file has been deleted.

edwarnicke: I know it *could* exist... I was curious if we'd hit a case like that yet :)  The reason was I think the only part of the system that is potentially prone to that problem is in the NSMgr ...

PR #445, opened 2020-08-20
fix: clienturl should work correctly in case of using unix sockets in k8s env

denis-tingajkin:

### DIFF
--- a/pkg/networkservice/common/clienturl/client.go
+++ b/pkg/networkservice/common/clienturl/client.go
@@ -71,7 +71,7 @@ func (u *clientURLClient) init() error {
 			return
 		}
 		var cc *grpc.ClientConn
-		cc, u.dialErr = grpc.DialContext(u.ctx, grpcutils.URLToTarget(clientURL), u.dialOptions...)
+		cc, u.dialErr = grpcutils.DialContext(u.ctx, grpcutils.URLToTarget(clientURL), u.dialOptions...)
 		if u.dialErr != nil {
 			return
 		}
--- a/pkg/registry/common/clienturl/ns_client.go
+++ b/pkg/registry/common/clienturl/ns_client.go
@@ -95,7 +95,7 @@ func (u *nsRegistryURLClient) init() error {
 			return
 		}
 		var cc *grpc.ClientConn
-		cc, u.dialErr = grpc.DialContext(u.ctx, grpcutils.URLToTarget(clientURL), u.dialOptions...)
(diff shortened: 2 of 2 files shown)


pkg/networkservice/common/clienturl/client.go
```diff
@@ -71,7 +71,7 @@ func (u *clientURLClient) init() error {
 			return
 		}
 		var cc *grpc.ClientConn
-		cc, u.dialErr = grpc.DialContext(u.ctx, grpcutils.URLToTarget(clientURL), u.dialOptions...)
+		cc, u.dialErr = grpcutils.DialContext(u.ctx, grpcutils.URLToTarget(clientURL), u.dialOptions...)
```
edwarnicke, line 74: Why are we having to shift to our own DialContext here?  grpcutils.URLToTarget(clientURL) should do it... what am I missing?

denis-tingajkin: > Why are we having to shift to our own DialContext here? 

This func has been added in https://github.com/networkservicemesh/sdk/pull/442

Motivation is that we should able to correctly work with the environment where file API is not fast. For example in k8s mounted unix socket file can be not available for some time and so on (when actually server has started and listened to the socket).

edwarnicke: OK... no.  Lets *not* do this.  All this does is obscure what's going on my burying an arbitrary choice of timeout in a library and *force* immediate dialing (both really bad).  If the main function wants a dial timeout.. that belongs in a *context*.... not in a library function.

edwarnicke:
Pushed #446 to remove the temptation to the anti-pattern of encapsulating this sort of thing deep down in libraries.

pkg/networkservice/common/clienturl/client.go
edwarnicke: As to motivation... totally agree on working correctly in environments where for example the API socket may not be available for a while... but that should be handled by Dial/CallOptions from main... not in a library function.

Please also note: great care has been taken to *generally* do late dial, not early dial (some exceptions exist in connect where dial has to succeed without the context of a call to make sense).  Wait spinning on dial here is the opposite of that.

If we want to be resilient in K8s environments, then we should set a proper context timeout *in main* for our call... and if need be the call option WaitForReady(true).

denis-tingajkin: @edwarnicke As I know WaitForReady(true) is a call option that is using on-call some API, not on dialing.

PR #446, opened 2020-08-20
Encapsulating choices of context and dial options in a library is an anti-pattern. That should be driven from top level.  Lets not do that here.

edwarnicke:


pkg/tools/grpcutils/dial.go
```diff
@@ -1,51 +0,0 @@
...
-// This function may be useful in the environments when not guarantee that the server socket is immediately provided.
-func DialContext(ctx context.Context, target string, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
-	return DialContextTimeout(ctx, target, time.Minute, opts...)
-}
-
-// DialContextTimeout tries to create a client connection to the given target during passed timeout.
-// This function may be useful in the environments when not guarantee that the server socket is immediately provided.
-func DialContextTimeout(ctx context.Context, target string, duration time.Duration, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error) {
-	timeout := time.After(duration)
-	for {
```
edwarnicke, line 37: Among other issues... this forces any DialContext to *block* until success.  Since *most* uses of DialContext should *not* do that... this is inappropriate in a drop in replacement for DialContext.

pkg/tools/grpcutils/dial.go
```diff
@@ -1,51 +0,0 @@
...
-	"time"
-
-	"github.com/pkg/errors"
-	"google.golang.org/grpc"
-)
-
-// DialContext tries to create a client connection to the given target during 1 minute.
-// This function may be useful in the environments when not guarantee that the server socket is immediately provided.
-func DialContext(ctx context.Context, target string, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
-	return DialContextTimeout(ctx, target, time.Minute, opts...)
```
edwarnicke, line 30: Timeout choices should come from outside... from the main or from the context of a call... not from a library.

denis-tingajkin:
@edwarnicke ok, How then we should handle a case when we are using unix based client/server + k8s slow file mount API?

PR #451, opened 2020-08-27
Excluded prefixes file constants now are exported

tigrulya-exe:
Constants are needed for networkservicemesh/cmd-exclude-prefixes-k8s#7

### DIFF
--- a/pkg/networkservice/common/excludedprefixes/const.go
+++ b/pkg/networkservice/common/excludedprefixes/const.go
@@ -18,10 +18,10 @@ package excludedprefixes
 
 /* These variables set default path to the config file */
 const (
-	// PrefixesFile - excluded prefixes file name
-	PrefixesFile = "excluded_prefixes.yaml"
-	// NsmConfigDir - excluded prefixes file directory name
-	NsmConfigDir = "/var/lib/networkservicemesh/config"
+	// prefixesFile - excluded prefixes file name
+	prefixesFile = "excluded_prefixes.yaml"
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/excludedprefixes/const.go
```diff
@@ -18,7 +18,10 @@ package excludedprefixes
...
 const (
-	prefixesFile            = "excluded_prefixes.yaml"
-	nsmConfigDir            = "/var/lib/networkservicemesh/config"
-	prefixesFilePathDefault = nsmConfigDir + "/" + prefixesFile
+	// PrefixesFile - excluded prefixes file name
+	PrefixesFile = "excluded_prefixes.yaml"
+	// NsmConfigDir - excluded prefixes file directory name
+	NsmConfigDir = "/var/lib/networkservicemesh/config"
+	// PrefixesFilePathDefault - excluded prefixes file absolute path
+	PrefixesFilePathDefault = NsmConfigDir + "/" + PrefixesFile
```
denis-tingajkin, line 26: Do we need to export only this constant?

tigrulya-exe: And `NsmConfigDir`, because we need to create directory if it doesn't exist.

denis-tingajkin: I guess we can don't care about NsmConfigDir because of k8s provides DirectoryOrCreate option for volumes :)

PR #453, opened 2020-09-01
We should return host not path for unspecified IPs

haiodo:
We should return host not path for unspecified IPs

### DIFF
--- a/pkg/tools/grpcutils/url.go
+++ b/pkg/tools/grpcutils/url.go
@@ -39,7 +39,7 @@ func URLToTarget(u *url.URL) (target string) {
 func AddressToURL(addr net.Addr) *url.URL {
 	if tcpAddr, ok := addr.(*net.TCPAddr); ok {
 		if tcpAddr.IP.IsUnspecified() {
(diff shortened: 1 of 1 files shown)


pkg/tools/grpcutils/url.go
```diff
@@ -39,7 +39,7 @@ func URLToTarget(u *url.URL) (target string) {
 func AddressToURL(addr net.Addr) *url.URL {
 	if tcpAddr, ok := addr.(*net.TCPAddr); ok {
 		if tcpAddr.IP.IsUnspecified() {
-			return &url.URL{Scheme: addr.Network(), Path: fmt.Sprintf("127.0.0.1:%v", tcpAddr.Port)}
+			return &url.URL{Scheme: addr.Network(), Host: fmt.Sprintf("127.0.0.1:%v", tcpAddr.Port)}
```
denis-tingajkin, line 42: Good catch, actually when URL parses TCP address it put IP and pot into host :)

denis-tingajkin (approved):

PR #454, opened 2020-09-09
Return error in case no network service is defined

haiodo:
Return error in case no network service is defined

### DIFF
--- a/pkg/networkservice/common/discover/server.go
+++ b/pkg/networkservice/common/discover/server.go
@@ -87,7 +87,15 @@ func (d *discoverCandidatesServer) Request(ctx context.Context, request *network
 	}
 
 	nsList := registry.ReadNetworkServiceList(nsStream)
+	if len(nsList) == 0 {
+		return nil, errors.Errorf("network service %s is not found", request.GetConnection().GetNetworkService())
+	}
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/discover/server.go
```diff
@@ -87,6 +87,9 @@ func (d *discoverCandidatesServer) Request(ctx context.Context, request *network
 	}
 
 	nsList := registry.ReadNetworkServiceList(nsStream)
+	if len(nseList) == 0 {
+		return nil, errors.Errorf("network service %s is not found", request.GetConnection().GetNetworkService())
+	}
```
denis-tingajkin, line 92:
```suggestion
	if len(nsList) == 0 {
		return nil, errors.Errorf("network service %s is not found", request.GetConnection().GetNetworkService())
	}
	if len(nseList) == 0 {
		return nil, errors.Errorf("network service endpoint for service %s is not found", request.GetConnection().GetNetworkService())
	}
```

denis-tingajkin (changes requested):

pkg/networkservice/common/discover/server.go
haiodo: Fixed, Thanks.

denis-tingajkin (approved):

PR #464, opened 2020-09-18
Interpose should request forwarder on rerequest

Bolodya1997:
Fix issue when interpose chain element doesn't request forwarder on double request


denis-tingajkin (commented):
Could you also add a unit test to cover the issue?

PR #469, opened 2020-09-22
Client chain elements

Bolodya1997:
1. Add 'interfaceName' to kernel mechanism client
2. ~Add memif mechanism client~


pkg/networkservice/common/mechanisms/memif/client.go
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2020 Doc.ai and/or its affiliates.
```
edwarnicke, line 1: I'm a bit confused why this is here... what is it attaching to the memif?  You need something (vpp or something else) to attach to memif...

Bolodya1997: OK, guess it is not the right place/time for the memif client :)

pkg/networkservice/common/mechanisms/kernel/client.go
```diff
@@ -21,19 +21,24 @@ import (
...
+type kernelMechanismClient struct {
+	interfaceName string
+}
 
 // NewClient - returns client that sets kernel preferred mechanism
-func NewClient() networkservice.NetworkServiceClient {
-	return &kernelMechanismClient{}
+func NewClient(interfaceName string) networkservice.NetworkServiceClient {
+	return &kernelMechanismClient{
+		interfaceName: interfaceName,
```
denis-tingajkin, line 40: As I can see we can work with `interfaceName== ""` https://github.com/networkservicemesh/api/blob/e5aee3ba99efe11afe73263b9d2e4396da3da1f7/pkg/api/networkservice/mechanisms/kernel/helpers.go#L92
So I would recommend adding an option for this chain element to set some custom interface name.

denis-tingajkin (changes requested):

pkg/networkservice/common/mechanisms/kernel/client.go
Bolodya1997: It is actually an option, user can set `interfaceName== ""` if there is no need for some custom interface name.

denis-tingajkin: I think better if the user do not need to pass anything into the constructor.

Bolodya1997: Yep, done

pkg/networkservice/common/mechanisms/kernel/option.go
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+package kernel
+
+type option struct {
+	interfaceName string
+}
+
+// Option for kernel mechanism
+type Option func(*option)
```
denis-tingajkin, line 24:
```suggestion
type Option func(*kernelMechanismClient )
```

denis-tingajkin (approved):

PR #470, opened 2020-09-23
Rework heal chain element to support updatetoken restoring Connection.Id to the Path.Index

edwarnicke:
The updatetoken chain element should restore the Connection.Id to match Path.PathSegment[Index].

Additionally, updatetoken should be the last thing that happens before
sending out the RPC call, so as to insure the Connection.Id and Path.Index
are always for the Connection.Id within the client when in client chain
elements.

This is important because otherwise the Connect.Id on the return through the
client does not match the Request.Connection.Id, and various indexing by identifiers
fails to work properly (including things like heal etc).

Because injectpeer is simply present to enable updatetoken, it is also moved
to the end of the chain just before updatetoken.

Unfortunately, this revealed a huge number of race conditions in the heal chain element,
which has been completely reworked to dispel them.

In doing so, it was discovered that eventchannel//monitor_client.go and monitor_connection_client.go
has some issues around respecting context.Context.

Which resulted in some small changes to adapters/monitor_server_to_client.go

### DIFF
--- a/pkg/networkservice/common/heal/client.go
+++ b/pkg/networkservice/common/heal/client.go
@@ -19,31 +19,28 @@ package heal
 
 import (
 	"context"
-	"sync"
-	"time"
+	"runtime"
 
+	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 
-	"github.com/networkservicemesh/sdk/pkg/networkservice/core/trace"
-	"github.com/networkservicemesh/sdk/pkg/tools/addressof"
-	"github.com/networkservicemesh/sdk/pkg/tools/extend"
 	"github.com/networkservicemesh/sdk/pkg/tools/serialize"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/addressof"
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 )
 
 type healClient struct {
-	ctx            context.Context
-	client         networkservice.MonitorConnectionClient
-	onHeal         *networkservice.NetworkServiceClient
-	heals          map[string]func()
-	connections    map[string]*networkservice.Connection
-	updateExecutor serialize.Executor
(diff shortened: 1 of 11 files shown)


pkg/networkservice/chains/chainstest/builder.go
```diff
@@ -227,8 +227,8 @@ func supplyMemoryRegistry() registry.Registry {
 }
 
 func supplyDummyForwarder(ctx context.Context, name string, generateToken token.GeneratorFunc, connectTo *url.URL, dialOptions ...grpc.DialOption) endpoint.Endpoint {
-	var result endpoint.Endpoint
-	result = endpoint.NewServer(ctx,
+	result := &struct{ endpoint.Endpoint }{}
+	result.Endpoint = endpoint.NewServer(ctx,
```
edwarnicke, line 231: This is a simple fix to a buggy test.  Net-net ... sometimes heal is triggered by one of the tests, and if it is, the previous version of this hits a panic because 

addressof.NetworkServiceClient(adapters.NewServerToClient(result))

Was basically sticking a nil chain element in the chain... this fixes that.

pkg/networkservice/common/heal/client.go
```diff
@@ -121,12 +121,21 @@ func (f *healClient) eventloop(recv networkservice.MonitorConnection_MonitorConn
 
 			case networkservice.ConnectionEventType_UPDATE:
 				for _, conn := range event.GetConnections() {
-					f.connections[conn.GetId()] = conn
+					path := conn.GetPath()
+					index := int(path.GetIndex() - 1)
+					if index >= 0 && len(path.GetPathSegments()) > index {
+						// TODO - do we really need the conn here, and even if we do, is this the *right* conn?
+						f.connections[path.GetPathSegments()[index].Id] = conn.Clone()
+					}
```
edwarnicke, line 129: Heal needs to correctly translate the changes reported by downstream to upstream

edwarnicke:
heal also needed to be fixed to correctly handle shifting from ConnectionEvents (which use the downstream server conn.Id) to the client's conn.Id

edwarnicke:
And hit a nil issue in chainstest.supplyDummyForwarder that needed to be fixed.

fkautz (approved):

PR #474, opened 2020-09-23
Small refactoring of interpose/localbypass/clienturl 

denis-tingajkin:

## Motivation

1. Moved clienturl to tools, that reduced lines of code and entities count. 
2. Added clienturl.Map and reused it in interpose and localbypass chain elements.


edwarnicke:
Overall good idea.  I'd suggest though that:

1. The context bit go into pkg/tools/clienturlctx
2. The Map bit go into pkg/tools/clienturlmap

This way you wind up with clean pkg naming that doesn't collide with clienturl chain elements.

denis-tingajkin:
@edwarnicke awesome idea, fixed

edwarnicke:
> @edwarnicke awesome idea, fixed

Ah.. I updated my comment to suggest stringurl not clienturlmap as package names there ... makes things a little more generic, in case there's a need for stringurl outside the clienturl context :)

denis-tingajkin:
@edwarnicke Indeed `string url` looks better and potentially can be reused

PR #481, opened 2020-09-24
Increase spire-agent healthcheck wait time

Bolodya1997:
Spire agent fails to start in 10 attempts, need to add extra tumeout.

### DIFF
--- a/pkg/tools/spire/start.go
+++ b/pkg/tools/spire/start.go
@@ -27,6 +27,7 @@ import (
 	"path"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/spiffe/go-spiffe/v2/workloadapi"
 
@@ -34,6 +35,10 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/networkservicemesh/sdk/pkg/tools/log"
+)
+
+const (
+	healthCheckTimeout = 10 * time.Second
 )
 
 // AddEntry - adds an entry to the spire server for parentID, spiffeID, and selector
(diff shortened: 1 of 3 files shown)


denis-tingajkin (commented):
I would also recommend adding a check that if the context has no deadline or context cannot be canceld (for example context.Background) then we need to use some default timeout.

pkg/tools/spire/start.go
```diff
@@ -220,3 +209,23 @@ func writeDefaultConfigFiles(ctx context.Context, spireRoot string) (string, err
...
 }
+
+func execHealthCheck(ctx context.Context, cmdStr string, options ...*exechelper.Option) error {
+	if _, ok := ctx.Deadline(); !ok {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, healthCheckTimeout)
+		defer cancel()
+	}
+
+healthCheck:
```
denis-tingajkin, line 220: Can we do not use labels? I would suggest just use a for loop
```suggestion:
for {
```

pkg/tools/spire/start.go
```diff
@@ -220,3 +209,22 @@ func writeDefaultConfigFiles(ctx context.Context, spireRoot string) (string, err
...
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, healthCheckTimeout)
+		defer cancel()
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
```
denis-tingajkin, line 224: It looks like this branch could load CPU. Should we add wait in case if err != nil?

Bolodya1997: `exechelper.Run(...)` starts process with `os/exec` and waits for its completion, so there is actually a wait, it is not a "for true" case.

denis-tingajkin: yes, but the process may fail in 50 ms

Bolodya1997: We can, so why not :)

edwarnicke:
This PR appears to be breaking downstream: 

https://github.com/networkservicemesh/cmd-forwarder-vppagent/pull/156/checks?check_run_id=1179132829

Also... I'm not clear on quite what the issue being solved here was...

Issue #483, opened 2020-09-24
Add networkservice/querycache chain element to reduce count of registry calls


denis-tingajkin:
## Problem

During testing, I found that in case of remote NSM connection we doing **5 extra calls into the registry**. It does not look good in general for interomain usecase... The root cause is https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/discover/server.go#L50

## Solution

Add `querycache` chain element that will be used on client-side and store all resolved via Find request NSEs.

edwarnicke:
Why a querycache rather than simply using memory as a bidrectional write through cache?

PR #487, opened 2020-09-26
feat: add nse querycache chain element for nsmgr

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/483


pkg/registry/memory/nse_server.go
```diff
@@ -34,7 +36,7 @@ import (
 type networkServiceEndpointRegistryServer struct {
 	networkServiceEndpoints NetworkServiceEndpointSyncMap
 	executor                serialize.Executor
-	eventChannels           []chan *registry.NetworkServiceEndpoint
+	eventChannels           map[string]chan *registry.NetworkServiceEndpoint
```
edwarnicke, line 39: How is this being protected against concurrency issues?

denis-tingajkin: This map is using only with executor.

edwarnicke: And I apparently was tired when I reviewed this... because I totally didn't see the executor just above it :(

edwarnicke:
Why a querycache rather than simply using memory as a bidrectional write through cache?

denis-tingajkin:
@edwarnicke `querycache` allows to store NSEs from the remote repository and watch them. Note: NSEs resolving via Find request when memory just keeps registered NSEs

pkg/registry/memory/nse_server.go
denis-tingajkin: That's ok, sometimes better to ask than look :)

edwarnicke:
Ah... got it... so you opted to separate the functionality... which is probably wise now that I think about it.

denis-tingajkin:
@edwarnicke also `qurycache` is using on client-side :) And this chain element fixes this issue https://github.com/networkservicemesh/sdk/issues/483

pkg/registry/common/querycache/nse_server.go
```diff
@@ -0,0 +1,95 @@
...
+	return streamchannel.NewNetworkServiceEndpointFindClient(ctx, resultCh), nil
+}
+
+func (q *queryCacheNSEClient) Unregister(ctx context.Context, in *registry.NetworkServiceEndpoint, opts ...grpc.CallOption) (*empty.Empty, error) {
+	return next.NetworkServiceEndpointRegistryClient(ctx).Unregister(ctx, in, opts...)
+}
+
+// NewClient creates new querycache registry.NetworkServiceEndpointRegistryClient that caches all resolved NSEs
+// All cached NSE is
+func NewClient() registry.NetworkServiceEndpointRegistryClient {
```
edwarnicke, line 93: Why is this file called nse_server.go when it contains a client?

denis-tingajkin: Oops! Will fix!

denis-tingajkin: thanks! Fixed.

PR #493, opened 2020-09-29
Liberalize heal test timers for non-linux OS's (macos and windows workers are slow in CI)

edwarnicke:


edwarnicke:
Don't merge just yet, this has turned into a probe PR to figure out why the TestHealClient_Request is wobbly in windows (and sometimes macos) CI.  I think its revealing a timing difference that is tripping over something racy in there... so want to get to the bottom of it.

PR #511, opened 2020-10-07
fix: zero size chain should not panic

denis-tingajkin:

### DIFF
--- a/pkg/networkservice/core/next/tests/client_test.go
+++ b/pkg/networkservice/core/next/tests/client_test.go
@@ -31,6 +31,9 @@ import (
 func TestNewNetworkServiceClientShouldNotPanic(t *testing.T) {
 	assert.NotPanics(t, func() {
 		_, _ = next.NewNetworkServiceClient().Request(context.Context(nil), nil)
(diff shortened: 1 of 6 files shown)


pkg/networkservice/core/next/tests/client_test.go
```diff
@@ -31,6 +31,9 @@ import (
 func TestNewNetworkServiceClientShouldNotPanic(t *testing.T) {
 	assert.NotPanics(t, func() {
 		_, _ = next.NewNetworkServiceClient().Request(context.Context(nil), nil)
+		_, _ = next.NewWrappedNetworkServiceClient(func(client networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
```
Bolodya1997, line 34: Probably it would be better to create another `TestNewWrappedNetworkServiceClientShouldNotPanic` test scenario instead of replacing the existing one.

denis-tingajkin: Hmm, actually I guess this line is not needed because NewWrappedNetworkServiceClient uses inside NewNetworkServiceClient :)

denis-tingajkin: It is not replacing just expanding.

Bolodya1997 (approved):

haiodo (approved):

Issue #512, opened 2020-10-08
Nested serialize.Executor.AsyncExec calls may deadlock when overwhelmed


xzfc:
Example of code that deadlocks:
```go
func TestDeadlock(t *testing.T) {
        var exec serialize.Executor

        exec.AsyncExec(func() {
                time.Sleep(time.Second)
                exec.AsyncExec(func() {})
        })

        for i := 0; i < 105; i++ { // 105 is a number that slightly more than channelSize
                exec.AsyncExec(func() {})
        }

        <-exec.AsyncExec(func() {})
}
```
The issue is that `AsyncExec()` may block on [channel write](https://github.com/networkservicemesh/sdk/blob/cf9c2259e946f48d65c4e4f44e8a283e239ebb50/pkg/tools/serialize/serialize.go#L63).

We have code that uses nested `AsyncExec()` and thus may potentially deadlock:
* https://github.com/networkservicemesh/sdk/blob/cf9c2259e946f48d65c4e4f44e8a283e239ebb50/pkg/networkservice/connectioncontext/dnscontext/client.go#L107

We should either document this limitation and rewrite the code that uses nested `AsyncExec()` OR rework `serialize.Executor` to support unlimited nested calls.

denis-tingajkin:
Tested locally it is reproducible.
Note: This dead local can be achieved in the real world only by **Server** chain elements which can receive in one moment 100(channelSize) or more requests.

denis-tingajkin:
To fix the issue we can use a concurrent queue instead of channel.

denis-tingajkin:
@edwarnicke , @haiodo 

I've found additional deadlock with executor:
```
func TestDeadlock2(t *testing.T)  {
	e := serialize.Executor{}
	e.AsyncExec(func() {
		time.Sleep(time.Millisecond * 100)
		<-e.AsyncExec(func() {
		})
	})
	<-e.AsyncExec(func() {
		logrus.Info("Deadlock")
	})
}
```

It looks like we need to fully rework ` serialize.Executor` to support this kind scenario. Thoughts?

edwarnicke:
I don't think the issue in this one is calling AsyncExec inside AsyncExec ... its *blocking* on the nested AsyncExec call.  And yes, that's one probably best addressed by documentation.

PR #519, opened 2020-10-13
fix: Nested serialize.Executor.AsyncExec calls may deadlock when overwhelmed 

denis-tingajkin:

## Motivation

https://github.com/networkservicemesh/sdk/issues/512

### DIFF
--- a/pkg/tools/serialize/serialize.go
+++ b/pkg/tools/serialize/serialize.go
@@ -20,19 +20,16 @@
 package serialize
 
 import (
+	"container/list"
 	"sync"
 	"sync/atomic"
 )
 
-const (
-	channelSize = 100
-)
-
 // Executor - a struct that can be used to guarantee exclusive, in order execution of functions.
 type Executor struct {
-	execCh chan func()
-	init   sync.Once
-	count  int32
+	mutex sync.Mutex
+	queue list.List
+	count int32
 }
 
 // NewExecutor - provides a new Executor
@@ -44,25 +41,26 @@ func NewExecutor() Executor {
 // AsyncExec - guarantees f() will be executed Exclusively and in the Order submitted.
 //        It immediately returns a channel that will be closed when f() has completed execution.
 func (e *Executor) AsyncExec(f func()) <-chan struct{} {
-	// Initialize *once*
-	e.init.Do(func() {
-		e.execCh = make(chan func(), channelSize)
+	done := make(chan struct{})
+	e.queue.PushBack(func() {
+		f()
+		close(done)
 	})
 	// Start go routine if we don't have one
 	if atomic.AddInt32(&e.count, 1) == 1 {
 		go func() {
-			for g := range e.execCh {
-				g()
+			for {
+				e.mutex.Lock()
+				first := e.queue.Front()
+				e.queue.Remove(first)
+				e.mutex.Unlock()
+				first.Value.(func())()
 				if atomic.AddInt32(&e.count, -1) == 0 {
 					return
 				}
 			}
 		}()
 	}
-	done := make(chan struct{})
-	e.execCh <- func() {
-		f()
-		close(done)
-	}
+
 	return done
 }
--- a/pkg/tools/serialize/serialize.go
+++ b/pkg/tools/serialize/serialize.go
@@ -20,7 +20,6 @@
 package serialize
 
 import (
-	"container/list"
 	"sync"
 	"sync/atomic"
 )
@@ -28,7 +27,7 @@ import (
 // Executor - a struct that can be used to guarantee exclusive, in order execution of functions.
 type Executor struct {
 	mutex sync.Mutex
-	queue list.List
+	queue []func()
 	count int32
 }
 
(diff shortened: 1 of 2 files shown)


pkg/tools/serialize/serialize.go
```diff
@@ -20,19 +20,16 @@
...
-const (
-	channelSize = 100
-)
-
 // Executor - a struct that can be used to guarantee exclusive, in order execution of functions.
 type Executor struct {
-	execCh chan func()
-	init   sync.Once
-	count  int32
+	mutex sync.Mutex
```
haiodo, line 30: @denis-tingajkin could we have some benchmark tests to compare both variants?

denis-tingajkin:
Bench test results

## Before

[root]goos: darwin
goarch: amd64
pkg: github.com/networkservicemesh/sdk/pkg/tools/serialize
BenchmarkExecutorSync
BenchmarkExecutorSync-8   	 1908020	       616 ns/op
PASS


goos: darwin
goarch: amd64
pkg: github.com/networkservicemesh/sdk/pkg/tools/serialize
BenchmarkExecutorSync
BenchmarkExecutorSync-8   	 2145922	       544 ns/op
PASS

## After

```
goos: darwin
goarch: amd64
pkg: github.com/networkservicemesh/sdk/pkg/tools/serialize
BenchmarkExecutorAsync
BenchmarkExecutorAsync-8   	 4798452	       236 ns/op
PASS

goos: darwin
goarch: amd64
pkg: github.com/networkservicemesh/sdk/pkg/tools/serialize
BenchmarkExecutorSync
BenchmarkExecutorSync-8   	 1908020	       616 ns/op
PASS
```

pkg/tools/serialize/serialize.go
```diff
@@ -44,25 +40,28 @@ func NewExecutor() Executor {
 // AsyncExec - guarantees f() will be executed Exclusively and in the Order submitted.
 //        It immediately returns a channel that will be closed when f() has completed execution.
 func (e *Executor) AsyncExec(f func()) <-chan struct{} {
-	// Initialize *once*
-	e.init.Do(func() {
-		e.execCh = make(chan func(), channelSize)
+	done := make(chan struct{})
+	e.mutex.Lock()
+	e.queue = append(e.queue, func() {
```
edwarnicke, line 45: Apropos of @haiodo 's comment about benchmarking... append here is a rather expensive operation.

denis-tingajkin: @edwarnicke please take a look at the benchmark result https://github.com/networkservicemesh/sdk/pull/519#issuecomment-707750806

denis-tingajkin: slices in go are optimized

pkg/tools/serialize/serialize.go
```diff
@@ -24,15 +24,11 @@ import (
...
-const (
-	channelSize = 100
-)
-
 // Executor - a struct that can be used to guarantee exclusive, in order execution of functions.
 type Executor struct {
-	execCh chan func()
-	init   sync.Once
-	count  int32
+	mutex sync.Mutex
```
edwarnicke, line 29: Mutex's do not guarantee ordering:

https://golang.org/src/sync/mutex.go?s=765:813#L42

So it's entirely possible for events to be queued out of order due to the out of order acquisition of a mutex.  Use of a mutex here basically yields the ordering guarantee.

denis-tingajkin: Yes, you're correct and it is a real problem. It seems to me we need to create our own analog of channel or FIFO mutex. Any ideas?

denis-tingajkin: Also, we can just increase the capacity of the channel and add notes

denis-tingajkin: @edwarnicke  Also, we can just increase the capacity of the channel and add notes

denis-tingajkin: @edwarnicke Lets consider use cases

In what follows, by the **Executor**, we mean `seriazlie.Executor` based on the mutex.
 
Usecases:

## 1. One goroutine using the Executor.
In this case, tasks will be executed FIFO because mutex will be used consistently. 

## 2. Two or more goroutine using the Executor
In this case, tasks will be executed FIFO for each goroutine but the order of goroutine maybe is not FIFO. This thing is not guaranteeing even by channel.


@edwarnicke Can you provide an example of what are you meaning?

edwarnicke:
I'm not quite ready on this one... exploring another option currently.

denis-tingajkin:
@edwarnicke I also did unboundedchannel but it wors x3 slow than this version and currently, I did not see the problems with mutex variant

edwarnicke:
Checkout this: https://github.com/edwarnicke/serialize/pull/3

denis-tingajkin:
@edwarnicke Lets move SDK to use executor from the separate repo :)

denis-tingajkin:
Will be fixed by https://github.com/edwarnicke/serialize/pull/3#pullrequestreview-510447947

PR #533, opened 2020-10-15
Add opentracing grpc interceptor options

tigrulya-exe:
Added opentracing grpc interceptor options to endpoint grpc server, sandbox client and server to enable correct tracing through grpc calls.


haiodo (approved):

pkg/networkservice/chains/endpoint/server.go
```diff
@@ -80,7 +83,7 @@ func (e *endpoint) Register(s *grpc.Server) {
 
 // Serve  - serves passed Endpoint on grpc
 func Serve(ctx context.Context, listenOn *url.URL, endpoint Endpoint, opt ...grpc.ServerOption) <-chan error {
-	server := grpc.NewServer(opt...)
+	server := grpc.NewServer(append(spanhelper.WithTracing(), opt...)...)
```
denis-tingajkin, line 86: Why do we need this here?

pkg/tools/sandbox/utils.go
```diff
@@ -80,7 +83,8 @@ func NewEndpoint(ctx context.Context, nse *registry.NetworkServiceEndpoint, gene
 
 // NewClient creates new networkservice.NetworkServiceClient configured to connect to the passed URL.
 func NewClient(ctx context.Context, generatorFunc token.GeneratorFunc, connectTo *url.URL, additionalFunctionality ...networkservice.NetworkServiceClient) (networkservice.NetworkServiceClient, error) {
-	cc, err := grpc.DialContext(ctx, grpcutils.URLToTarget(connectTo), grpc.WithBlock(), grpc.WithInsecure())
+	cc, err := grpc.DialContext(ctx, grpcutils.URLToTarget(connectTo),
```
denis-tingajkin, line 86: Why do we need this here?

PR #537, opened 2020-10-16
Revert adding opentracing grpc options to endpoint server

tigrulya-exe:
Signed-off-by: Tigran Manasyan <tigran.manasyan@xored.com>

### DIFF
--- a/pkg/networkservice/chains/endpoint/server.go
+++ b/pkg/networkservice/chains/endpoint/server.go
@@ -23,8 +23,6 @@ import (
 	"context"
 	"net/url"
 
-	"github.com/networkservicemesh/sdk/pkg/tools/spanhelper"
-
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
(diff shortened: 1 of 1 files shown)


pkg/networkservice/chains/endpoint/server.go
```diff
@@ -83,7 +81,7 @@ func (e *endpoint) Register(s *grpc.Server) {
 
 // Serve  - serves passed Endpoint on grpc
 func Serve(ctx context.Context, listenOn *url.URL, endpoint Endpoint, opt ...grpc.ServerOption) <-chan error {
-	server := grpc.NewServer(append(spanhelper.WithTracing(), opt...)...)
+	server := grpc.NewServer(opt...)
```
denis-tingajkin, line 84: Do we need also do the same change for the client?

tigrulya-exe: In #533 I've added options to `sandbox.NewClient()`, which we use only for tests. Did you meant it?

denis-tingajkin: yes, you're correct

Issue #521, opened 2020-10-13
Pass-through NSEs cannot work because interpose incorrectly looking in Path


xzfc:
This is an attempt to write a pass-through test for NSMgr: https://github.com/xzfc/networkservicemesh-sdk/commit/70425f85bc669a0710b2d10f51aea73c9de27798.
It fails, probably because the same NSMgr appears twice in the path. Need to investigate and fix NSMgr to handle this scenario. (the test itself could be wrong too)

<details><summary>Request diagram</summary><p>

![graphviz](https://user-images.githubusercontent.com/5121426/95838032-d484db00-0d30-11eb-8630-c391baa9f515.png)

</p></details>

<details><summary>Test output</summary><p>

```
$ go test -v -run TestNSMGR_Chain
=== RUN   TestNSMGR_Chain
INFO[0000] ExcludedPrefixesService: adding excluded prefixes to connection: [] 
ERRO[0000] An error during watch file: no such file or directory 
INFO[0000] ExcludedPrefixesService: adding excluded prefixes to connection: [] 
Forwarding to -> my-service-remote-2
INFO[0000] ExcludedPrefixesService: adding excluded prefixes to connection: [] 
INFO[0000] ExcludedPrefixesService: adding excluded prefixes to connection: [] 
INFO[0000] ExcludedPrefixesService: adding excluded prefixes to connection: [] 
Forwarding to -> my-service-remote-2
INFO[0000] ExcludedPrefixesService: adding excluded prefixes to connection: [] 
    server_test.go:203: 
                Error Trace:    server_test.go:203
                Error:          Received unexpected error:
                                rpc error: code = Unknown desc = all cross NSE failed to connect to endpoint tcp://127.0.0.1:41529 connection: id:"244fbcf0-7310-4ae2-bf16-bbe0c2be7cdd" network_service:"my-service-remote-3" context:<ip_context:<> > path:<index:1 path_segments:<name:"nsc-5e6b6297-bfd3-4aee-a17d-24195f66224d" id:"1" token:"TestToken" expires:<seconds:32503683661 nanos:1 > > path_segments:<name:"nsmgr-2b4f8de3-f4a2-4163-adae-8dae0ec8f2f3" id:"244fbcf0-7310-4ae2-bf16-bbe0c2be7cdd" token:"TestToken" expires:<seconds:32503683661 nanos:1 > > > network_service_endpoint_name:"endpoint-3" 
[...]

PR #544, opened 2020-10-19
Pass-through test

glazychev-art:
#521

### DIFF
--- a/pkg/networkservice/chains/nsmgr/server_test.go
+++ b/pkg/networkservice/chains/nsmgr/server_test.go
@@ -19,9 +19,17 @@ package nsmgr_test
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
+	"net/url"
 	"testing"
 	"time"
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"google.golang.org/grpc"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/adapters"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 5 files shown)


pkg/networkservice/chains/nsmgr/server_test.go
```diff
@@ -140,3 +148,114 @@ func TestNSMGR_LocalUsecase(t *testing.T) {
 	require.NotNil(t, conn2)
 	require.Equal(t, 5, len(conn2.Path.PathSegments))
 }
+
+func TestNSMGR_PassThrough(t *testing.T) {
```
denis-tingajkin, line 152: Can you also add a test that covers a case when all nses are registered in one nsmgr?

glazychev-art: Done

pkg/registry/common/interpose/server.go
```diff
@@ -38,6 +38,11 @@ import (
 // interposeNSEName - a common prefix for all registered cross NSEs
 const interposeNSEName = "interpose-nse#"
 
+// GetCrossNSEPrefix - returns interposeNSEName
+func GetCrossNSEPrefix() string {
```
denis-tingajkin, line 42: Can we just make `interposeNSEName` as public?

denis-tingajkin (commented):
In general looks good to me

denis-tingajkin (approved):

pkg/networkservice/common/interpose/server.go
```diff
@@ -135,10 +131,23 @@ func (l *interposeServer) Request(ctx context.Context, request *networkservice.N
...
 	id := ""
-	for i := conn.GetPath().GetIndex(); i > 0; i-- {
-		if conn.GetPath().GetPathSegments()[i].Name == l.name {
-			id = conn.GetPath().GetPathSegments()[i].Id
+	segmentIdx := conn.GetPath().GetIndex()
+	isCrossNSEPrev := false
+
+	l.endpoints.Range(func(key string, crossNSEURL *url.URL) bool {
+		isCrossNSEPrev =
+			conn.Path.PathSegments[segmentIdx-1].Name == strings.TrimPrefix(key, interpose.InterposeNSEName)
```
haiodo, line 139: You should just check for Path segment Name to match endpoint name, you should not use interpose.InterposeNSENam it is just for registry.

haiodo (approved):

PR #550, opened 2020-10-20
Reduce logs

tigrulya-exe:
Added tracehelper chain element for logs reducing


pkg/networkservice/common/tracehelper/context.go
```diff
@@ -0,0 +1,49 @@
...
+const (
+	connectionInfoKey contextKeyType = "ConnectionInfo"
+)
+
+type contextKeyType string
+
+// ConnectionInfo - struct, containing string representations of request and response, used for tracing.
+type ConnectionInfo struct {
+	// Request is string representation of networkservice.NetworkServiceRequest
+	Request string
```
haiodo, line 32: I would prefer to see a protobuf message here.

edwarnicke:
Would it perhaps simply make more sense to set the  log level of 'Trace' logs in the trace chain element to 'Trace' and then we can use logrus log level controls to manage them?

edwarnicke:
(the same means might be used to manage whether to do internal spans as well)

edwarnicke:
This might be simpler:

https://github.com/networkservicemesh/sdk/pull/556

It's benefits are:
1.  Super simple
2. Also removes the trace chain element when not in use (which makes it easier to read call chains in debugging)

Drawbacks:
1.  It ties Jaeger internal tracing to logrus.TraceLevel
2. Its static (set at startup time) and not dynamic.

So consider it as an idea for log reduction, but perhaps an imperfect one.

pkg/networkservice/common/tracehelper/context.go
tigrulya-exe: Replaced with `proto.Message`

pkg/networkservice/core/chain/client.go
```diff
@@ -20,14 +20,16 @@ import (
...
+	"github.com/networkservicemesh/sdk/pkg/networkservice/common/tracehelper"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/trace"
 )
 
 // NewNetworkServiceClient - chains together a list of networkservice.NetworkServiceClient with tracing
 func NewNetworkServiceClient(clients ...networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
 	if logrus.GetLevel() == logrus.TraceLevel {
-		return next.NewWrappedNetworkServiceClient(trace.NewNetworkServiceClient, clients...)
+		tracedClients := next.NewWrappedNetworkServiceClient(trace.NewNetworkServiceClient, clients...)
```
haiodo, line 31: Please handle all logic inside trace.NewNetworkServiceClient

haiodo:
@tigrulya-exe It would be interesting how it is shown, with existing and new approach. We could put example into readme file. Also Some benchmarking with and without diff and tracing could be interesting. +Option to enable diff?

pkg/networkservice/core/chain/client.go
tigrulya-exe: Removed additional next.{Client, Server} wrapping

tigrulya-exe:
@haiodo Added README.md with example and benchmarks.

PR #562, opened 2020-10-28
Recvfd Registry should expect Endpoint name on return

haiodo:
Recvfd Registry should expect Endpoint name on return

### DIFF
--- a/pkg/registry/common/recvfd/server.go
+++ b/pkg/registry/common/recvfd/server.go
@@ -54,10 +54,15 @@ func (r *recvfdNseServer) Register(ctx context.Context, endpoint *registry.Netwo
 	}
 
 	// Get the fileMap
-	fileMap, _ := r.fileMaps.LoadOrStore(endpoint.GetName(), &perEndpointFileMap{
+	fileMap := &perEndpointFileMap{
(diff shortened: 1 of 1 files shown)


pkg/registry/common/recvfd/server.go
```diff
@@ -73,6 +78,10 @@ func (r *recvfdNseServer) Register(ctx context.Context, endpoint *registry.Netwo
 	}
 	returnedEndpoint = returnedEndpoint.Clone()
 
+	if endpointName != returnedEndpoint.Name {
```
edwarnicke, line 81: Curious, when would endpointName != returnedEndpoint.Name?

denis-tingajkin: I feel because remote side could set unique identity for incoming endpoint

edwarnicke: Got it.  I was curious if there was a specific use case in mind, or just an abundance of caution :)

haiodo: If they are same we are during heal Register, to update we alive. And if not we expect remote side to update with a proper name.

PR #563, opened 2020-10-28
Support of RecvFD with Nsmgr chain

haiodo:
Support of RecvFD with Nsmgr chain


edwarnicke:
recvfd and sendfd don't (and can't) support windows, so you probably want to add a

```
// +build !windows
```

To your nsmgr chain.

denis-tingajkin (approved):

haiodo:
> recvfd and sendfd don't (and can't) support windows, so you probably want to add a
> 
> ```
> // +build !windows
> ```
> 
> To your nsmgr chain.

Not sure it is a good idea actually, so I think about adding dummy chains to include sendfd/recvfd into nsmgr.

PR #566, opened 2020-10-30
Add memif client

Bolodya1997:


pkg/networkservice/common/mechanisms/memif/client.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+type memifMechanismClient struct {
+	socketFileName string
+}
+
+// NewClient - returns client that sets memif preferred mechanism
+func NewClient(socketFileName string) networkservice.NetworkServiceClient {
```
edwarnicke, line 38: Generally speaking, clients are *always* consumers of socket files in memif, *never* providers.  Do we have a use case in which we need it to be otherwise?

PR #571, opened 2020-11-06
discoverServer: handle Close

### LINKED ISSUE #570: NSMgr should pass Close() calls
Currently, NSMgr ignores incoming NetworkService `Close()` calls so that the corresponding NSE never receives them. More precisely, it is ignored by [discover] which is a part of NSMgr chain.

[discover]: https://github.com/networkservicemesh/sdk/blob/3b08d97e38650d6e85de383b395431a08c99a9eb/pkg/networkservice/common/discover/server.go#L103-L105

Related PR that introduced this behavior: https://github.com/networkservicemesh/sdk/pull/263.

To-do:
* [x] Fix `discoverServer` #571
* [ ] Refactor NSMgr sandbox tests, see https://github.com/networkservicemesh/sdk/pull/571#discussion_r518904845. Refactored tests should check that `Close()` works.
* [ ] Ensure that `Close()` works in a pass-trough scenario. Current status: `healClient` crashes. https://github.com/xzfc/networkservicemesh-sdk/commit/bb1e979bb0adff007a5c4e2d28f67fac1220446a

xzfc:
This PR fixes #570 for Local/Remote one-hop NSMgr use-cases.

### DIFF
--- a/pkg/networkservice/chains/nsmgr/server_test.go
+++ b/pkg/networkservice/chains/nsmgr/server_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/url"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -62,7 +63,8 @@ func TestNSMGR_RemoteUsecase(t *testing.T) {
 		NetworkServiceNames: []string{"my-service-remote"},
 	}
 
-	_, err := sandbox.NewEndpoint(ctx, nseReg, sandbox.GenerateTestToken, domain.Nodes[0].NSMgr)
+	counter := &counterServer{}
+	_, err := sandbox.NewEndpoint(ctx, nseReg, sandbox.GenerateTestToken, domain.Nodes[0].NSMgr, counter)
 	require.NoError(t, err)
 
 	request := &networkservice.NetworkServiceRequest{
@@ -79,7 +81,7 @@ func TestNSMGR_RemoteUsecase(t *testing.T) {
 	nsc, err := sandbox.NewClient(ctx, sandbox.GenerateTestToken, domain.Nodes[1].NSMgr.URL)
 	require.NoError(t, err)
 
-	conn, err := nsc.Request(ctx, request)
+	conn, err := nsc.Request(ctx, request.Clone())
 	require.NoError(t, err)
 	require.NotNil(t, conn)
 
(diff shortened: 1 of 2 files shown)


pkg/networkservice/chains/nsmgr/server_test.go
```diff
@@ -45,66 +45,30 @@ import (
...
+	var nodesCount, nscNode, expectedPathSegments int
+	if remote {
+		nodesCount = 2
+		nscNode = 1
+		expectedPathSegments = 8
+	} else {
+		nodesCount = 1
+		nscNode = 0
+		expectedPathSegments = 5
 	}
```
denis-tingajkin, line 68: Please consider extracting testSample struct  and using _for_ loop over samples

```go
type testSample struct {
  nodeCount uint
  nscNode uint
  nseNode uint
  reRequestCount unit
  expectedPathSegments unit
  expectedreRequestCount unit 
}
```

denis-tingajkin (changes requested):

pkg/networkservice/chains/nsmgr/server_test.go
xzfc: I don't feel that the only two cases worth generalizing.

denis-tingajkin: 1. Actually you already generalized these two cases
2. With `testSample` we can cover many cases such as 
    2.1. nsc - node1, nse - node 2
    2.2. nsc - node2, nse - node1
    2.3. nsc - node1, nse -node1 (nodesCount=1),
    2.4. nsc  - node2, nse -node2 (nodesCount=2)
    2.6. all above scenarious with reRequestsCount = 2
    2.7. all above scenarious with reRequestsCount = 5
    2.7. all above scenarious with reRequestsCount = 10
3. test pattern with testSample looks more better than adding method with bolean logic.

pkg/networkservice/chains/nsmgr/server_test.go
```diff
@@ -318,3 +328,17 @@ func (p *passThroughClient) Close(ctx context.Context, conn *networkservice.Conn
 	conn = conn.Clone()
 	return next.Client(ctx).Close(ctx, conn, opts...)
 }
+
+type counterServer struct {
+	Requests, Closes int
```
denis-tingajkin, line 333: Please use atomic int32 to prevent race condition in case of a testing parallel request from a few clients

denis-tingajkin (commented):
Looks good, please add using of atomic to prevent race conditions

denis-tingajkin (approved):

Issue #573, opened 2020-11-06
Filter mechanism chain element works incorrectly in remote case


Bolodya1997:
# Issue
In the remote case filter mechanism chain element removes all mechanism preferences from the request:
```
Client (adds LOCAL) -> NSMgr                           |  request:{ LOCAL }
NSMgr (removes all REMOTE) -> Forwarder                |  request:{ LOCAL }
Forwarder (adds LOCAL, REMOTE) -> NSMgr                |  request:{ LOCAL, REMOTE }
NSMgr (removes all LOCAL) -> remote NSMgr              |  request:{ REMOTE }
remote NSMgr (removes all REMOTE) -> remote Forwarder  |  request:{}
```
# Possible solution
Don't filter mechanisms if target URL is interpose NSE.

denis-tingajkin:
@Bolodya1997 Actually the issue is invalid. The problem was related to the interpose chain element... WIll revert #575.

PR #576, opened 2020-11-08
fix: Filter mechanism chain element works incorrectly in remote case

denis-tingajkin:
## Motivation

1. Fixes root cause of https://github.com/networkservicemesh/sdk/issues/573
2. Reverts https://github.com/networkservicemesh/sdk/pull/575 because the real problem is related to `interpose`.


pkg/networkservice/chains/nsmgr/server.go
```diff
@@ -112,24 +122,16 @@ func NewServer(ctx context.Context, nsmRegistration *registryapi.NetworkServiceE
 		localbypass.NewServer(&localbypassRegistryServer),
 		excludedprefixes.NewServer(ctx),
 		newRecvFD(), // Receive any files passed
-		interpose.NewServer(nsmRegistration.Name, &interposeRegistry),
+		interpose.NewServer(nsmRegistration.Name, &interposeRegistry, connectServer),
```
haiodo, line 125: @denis-tingajkin We already discussed before, and I think we should simply had a context variable describing we are going into cross connect NSE, and just modify filter mechanism to check for it. Instead of making a chain jumpings.

pkg/networkservice/common/interpose/server.go
```diff
@@ -93,15 +97,15 @@ func (l *interposeServer) Request(ctx context.Context, request *networkservice.N
 		// Iterate over all cross connect NSEs to check one with passed state.
 
 		l.endpoints.Range(func(key string, crossNSEURL *url.URL) bool {
-			crossCTX := clienturlctx.WithClientURL(ctx, crossNSEURL)
+			crossCTX := extend.WithValuesFromContext(ctx, clienturlctx.WithClientURL(context.Background(), crossNSEURL))
```
haiodo, line 100: Did we loose a timeout here?

pkg/networkservice/common/interpose/server.go
```diff
@@ -93,15 +97,15 @@ func (l *interposeServer) Request(ctx context.Context, request *networkservice.N
...
 
 			// Store client connection and selected cross connection URL.
 			connInfo, _ = l.activeConnection.LoadOrStore(clientConnID, connectionInfo{
 				endpointURL:     clientURL,
 				interposeNSEURL: crossNSEURL,
 				requestingNSE:   true,
 			})
-			result, err = next.Server(crossCTX).Request(crossCTX, request)
+			result, err = l.connectServer.Request(crossCTX, request)
```
haiodo, line 108: I will prefer to set a context variable about direction, or extend client url to had information about direction was overridden.

haiodo (changes requested):
Please see comments.

pkg/networkservice/common/interpose/server.go
denis-tingajkin: Please look at `extend`, we only lose values here.

pkg/networkservice/chains/nsmgr/server.go
denis-tingajkin: @haiodo I think context variable is a more complex solution. Because in this case all elements after interpose should know about two interpose branches...

pkg/networkservice/common/interpose/server.go
denis-tingajkin: Let's continue to discuss here https://github.com/networkservicemesh/sdk/pull/576#discussion_r519550687

pkg/networkservice/common/interpose/server.go
haiodo: ok

Bolodya1997 (approved):
Looks great.

pkg/registry/common/interpose/server.go
```diff
@@ -42,8 +42,13 @@ type interposeRegistry struct {
 	endpoints *stringurl.Map
 }
 
+// Is returns true if passed nama contains interpose identity
```
haiodo, line 45:
```suggestion
// Is returns true if passed name contains interpose identity
```

haiodo (approved):
Looks good, @denis-tingajkin just please fix typo

pkg/registry/common/interpose/server.go
denis-tingajkin: @haiodo thanks, fixed

denis-tingajkin:
@haiodo done

haiodo (approved):

PR #580, opened 2020-11-10
refactor: Simplify setlogoption chain element usage + fix few issues

denis-tingajkin:

## Motivation

1. Simplified usage of setlogoption chain element
2. Fixed issue with using a traced chain for nse/ns chain registries in case if the log level is not tracing
3. setlogoption is using only if log level is tracing.


pkg/registry/common/setlogoption/ns_server.go
```diff
@@ -30,42 +30,40 @@ import (
...
+	return s.NetworkServiceRegistry_FindServer.Send(ns)
 }
 
-func (s *setNSLogOptionFindServer) Context() context.Context {
+func (s *setLogOptionFindServer) Context() context.Context {
 	return s.ctx
 }
 
-func (s *setNSLogOption) Register(ctx context.Context, service *registry.NetworkService) (*registry.NetworkService, error) {
+func (s *setNSLogOption) Register(ctx context.Context, endpoint *registry.NetworkService) (*registry.NetworkService, error) {
```
haiodo, line 48: why it is endpoint if it is service?

denis-tingajkin: oops! Fixed, thanks

haiodo (approved):

Issue #552, opened 2020-10-20
Extract connectClient mechanism translation logic


xzfc:
Currently, `connectClient` is used to translate `Connection.Mechanism` between incoming server connection and outgoing client connection:
https://github.com/networkservicemesh/sdk/blob/fbf9869cd4413a8be38f89b93220f10835a8fcf1/pkg/networkservice/common/connect/client.go#L47-L51

This behavior is required by `nsmgr` which uses `connect`: https://github.com/networkservicemesh/sdk/blob/fbf9869cd4413a8be38f89b93220f10835a8fcf1/pkg/networkservice/chains/nsmgr/server.go#L113

However, it is not required by `nsmgrproxy` which uses `connect` too:
https://github.com/networkservicemesh/sdk/blob/fbf9869cd4413a8be38f89b93220f10835a8fcf1/pkg/networkservice/chains/nsmgrproxy/server.go#L44

We need to extract mechanism translation logic from `connectClient` into a new chain element and embed it into `nsmgr` chain. Also, we may hide `connect.NewClient` as it becomes an implementation detail.

edwarnicke:
> We need to extract mechanism translation logic from connectClient into a new chain element and embed it into nsmgr chain.

Question, would it might also be useful to allow for an option (preferably a With style option) to provide a function for the connection translation required by particular use?

xzfc:
Ok, let's list translation use-cases:
* `Connection.Mechanism`: this issue.
* `Connection.Id`: required in the past, but no longer needed, as you said in https://github.com/networkservicemesh/sdk/pull/525#discussion_r505650097.

Do we have any other use-cases for this?

edwarnicke:
> Ok, let's list translation use-cases:
> 
> * `Connection.Mechanism`: this issue.
> * `Connection.Id`: required in the past, but no longer needed, as you said in [#525 (comment)](https://github.com/networkservicemesh/sdk/pull/525#discussion_r505650097).
> 
> Do we have any other use-cases for this?

Those capture the two currently on the table yes.  I could imagine other aspects of the clientConnection (labels for example) being selectively copied to the serverConnection (thus my suggestion for a relatively genericizable With* Option for this).

xzfc:
I think we may make it configurable later if such a requirement arises.

But currently, it is intended to be used in only one place (`nsmgr`) and only to translate Mechanisms, so we may make it unconfigurable, for simplicity.

edwarnicke:
[...]

PR #585, opened 2020-11-12
Fix connect

Bolodya1997:
# Issue
Existing `connect.NewServer()` chain elements provides some request, connection translation which should be different for the different cases (ex: NSMgr, interpose NSE, pass through NSE) https://github.com/networkservicemesh/sdk/issues/552.
# Solution
~Add `translationClientFactory` to extract all translation logic from the `connect.NewServer()`, add `translation` package with common translation clients `translation.NewNSMgrClient()`..., and`translation.Builder` to create custom ones.~
Add `NewCrossConnectClientFactory` for the cross connect (interpose NSE, pass through NSE) cases, rework connect.


pkg/tools/clientmap/refcount.go
```diff
@@ -31,54 +31,75 @@ type refcountClient struct {
...
+	client := raw.(*refcountClient)
+
+	atomic.AddInt32(&client.count, 1)
+
+	return client.client, loaded
+}
+
+// LoadAndDelete returns the value stored in the map for a key, and decrements the refcount. If no value found or the
+// refcount < 0, returns (nil, false). If refcount == 0, deletes the value.
+func (r *RefcountMap) LoadAndDelete(key string) (networkservice.NetworkServiceClient, bool) {
```
denis-tingajkin, line 84: Note: LoadAndDelete must be done under one _lock section_. Please try to use `r.m.LoadAndDelete`

denis-tingajkin (changes requested):

pkg/tools/clientmap/refcount.go
Bolodya1997: Yep, there was an issue. `r.m.LoadAndDelete` can't be used here because refcount is stored in the map entry. Added `sync.RWMutex` to fix issue.

pkg/networkservice/common/connect/translation/clients.go
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package translation
+
+import "github.com/networkservicemesh/api/pkg/api/networkservice"
+
+// NewNSMgrClient returns a new translation client for the NSMgr
+func NewNSMgrClient() networkservice.NetworkServiceClient {
```
edwarnicke, line 22: Shouldn't these live with their appropriate chains or chain elements rather than centralizing them here?

edwarnicke:
One other potentially thought here... rather than having translation factories as arguments to connect ... we could have connect store its client connection in [metadata](https://github.com/networkservicemesh/sdk/pull/590) and then appropriate subsequent chain elements could handle the translation without having to complicate the connect logic with translation.

Bolodya1997:
> One other potentially thought here... rather than having translation factories as arguments to connect ... we could have connect store its client connection in [metadata](https://github.com/networkservicemesh/sdk/pull/590) and then appropriate subsequent chain elements could handle the translation without having to complicate the connect logic with translation.

Yep, it can be very helpful to store some additional info about the server+client connections in s single place, but we need to perform translation logic before the client chain:
```go
connect.NewServer(
	ctx,
	client.NewClientFactory(
		name,
		// What to call onHeal
		addressof.NetworkServiceClient(adapters.NewServerToClient(rv)),
		tokenGenerator,
		// <-- all these chain elements comes after the updatepath, refresh, updatetoken...
		connectioncontextkernel.NewClient(),
		tag.NewClient(ctx, vppConn),
		// mechanisms
		memif.NewClient(vppConn, &lastSocketID),
		kernel.NewClient(vppConn),
		recvfd.NewClient(),
	),
	clientDialOptions...,
),
```
so if we want to perform some translation, we should do it before the client.NewClientFactory() chain. And it is actually the reason why we need to pass `translationFactory` to the connection server.

Bolodya1997:
> One other potentially thought here... rather than having translation factories as arguments to connect ... we could have connect store its client connection in [metadata](https://github.com/networkservicemesh/sdk/pull/590) and then appropriate subsequent chain elements could handle the translation without having to complicate the connect logic with translation.

There are actually 2 different approaches we can use to implement server request -> client request and client connection -> server connection translations.

### Use metadata and stored `serverRequest`, `clientConnection`

```go
endpoint := chain.NewNetworkServiceServer(
    ...
    customtranslator.NewRequestServer(), // <-- creates `clientRequest` from request/metadata, stores `request` 
                                         //     as `serverRequest` in metadata, replaces `request` with
                                         //     `clientRequest`
    connection.NewServer( // <-- already uses `clientRequest`, no need for translation
        client.NewFactory(clientChain), // <-- same
    ),
    customtranslator.NewConnectionServer(), // <-- updates `clientRequest` with `conn` (`clientConn`), stores
                                            //     it in metadata, replaces `request` with `serverRequest`
    ...
)
```
We can't add translation as a part of `clientChain` because it is processed after all default `client` chain elements:
https://github.com/networkservicemesh/sdk/blob/332daa2594bdcfb961112d032b5a8c100d5e61d4/pkg/networkservice/chains/client/client.go#L59-L70
so it should be performed in server chain and so we need 2 translation elements:
1. to translate server request before starting processing client chain
2. to translate client connection after the client chain
Another one thing I really don't like is that `NewRequestServer`, `connection.NewServer` and `NewConnectionServer` are the part of the server chain **but** they are processing client request.

### Use `translationFactory`

```go
endpoint := chain.NewNetworkServiceServer(
    ...
    connection.NewServer( // <-- uses only `request`
        translation.NewClient, // <-- is inserted before the `clientChain`, replaces `request` with
                               //     `clientRequest`, returns `serverConn` instead of `clientConn`
        client.NewFactory(clientChain), // <-- already uses `clientRequest`, no need for translation
    ),
    ...
)
```
Here we have only 1 chain element incapsulating all translation logic and so it is only one who is processing both server and client request. No one in the server chain knows about the client request/connection, same for the client chain and the server request/connection.

pkg/networkservice/common/connect/translation/clients.go
Bolodya1997: Yep, fixed.

pkg/networkservice/common/connect/translation/builder.go
```diff
@@ -0,0 +1,105 @@
...
+
+// ReplaceMechanism replaces request.Connection.Mechanism with clientConn.Mechanism
+func ReplaceMechanism() RequestOption {
+	return func(request *networkservice.NetworkServiceRequest, clientConn *networkservice.Connection) {
+		request.GetConnection().Mechanism = clientConn.GetMechanism()
+	}
+}
+
+// ReplacePath replaces request.Connection.Path with clientConn.Path
+func ReplacePath() RequestOption {
```
denis-tingajkin, line 68: This option is not using.

denis-tingajkin (commented):
Other things look for me over desined. 
But I tihnk this PR can be merged as bug fixing.

I would suggest to solve this with something like: Create two separte client factories:
1. For nsmgr and so on.
2. For forwarders.

Something like
```
nsmgr.NewClientFactory
forwarder.NewClientFactory
```

edwarnicke:
> ### Use `translationFactory`
> ```go
> endpoint := chain.NewNetworkServiceServer(
>     ...
>     connection.NewServer( // <-- uses only `request`
>         translation.NewClient, // <-- is inserted before the `clientChain`, replaces `request` with
>                                //     `clientRequest`, returns `serverConn` instead of `clientConn`
>         client.NewFactory(clientChain), // <-- already uses `clientRequest`, no need for translation
>     ),
>     ...
> )
> ```
> 
> Here we have only 1 chain element incapsulating all translation logic and so it is only one who is processing both server and client request. No one in the server chain knows about the client request/connection, same for the client chain and the server request/connection.

I like this approach very much :)

pkg/networkservice/common/connect/translation/builder.go
Bolodya1997: Now removed all options, translation client translates only mechanisms.

pkg/tools/clientmap/refcount.go
```diff
@@ -17,68 +17,105 @@
...
+	count int32
+	networkservice.NetworkServiceClient
 }
 
 // RefcountMap - clientmap.Map wrapped with refcounting.  Each Store,Load,LoadOrStore increments the count,
 //               each Delete decrements the count.  When the count is zero, Delete deletes.
 type RefcountMap struct {
-	Map
+	m    Map
+	lock sync.RWMutex
```
denis-tingajkin, line 35: Why do we need a lock here?

denis-tingajkin (commented):
Looks good

pkg/networkservice/common/translation/client.go
```diff
@@ -0,0 +1,72 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package translation provides client chain element to translate server request mechanisms to client request mechanisms
+// and client connection mechanisms to server connection mechanisms
+package translation
```
denis-tingajkin, line 19: Can we rename this to something meaning that it is related to mechanisms?

For example: `translationmechanism` or something similar.

pkg/tools/clientmap/refcount.go
Bolodya1997: For `Delete`, `LoadAndDelete` we need to be sure that no one should update `entry.refcount` when we are going to delete an entry.

pkg/networkservice/common/translation/client.go
Bolodya1997: Renamed to `mechanismtranslation`.

pkg/tools/clientmap/refcount.go
denis-tingajkin: @Bolodya1997 Then I guess we should not to use sync.Map it looks like an over synchronization

Bolodya1997: Yep, it was too slow in comparison with simple `map + sync.Mutex`. But if we put `sync.RWMutex` in every `refcountClient` instead of keeping single lock for all map, it becomes significantly faster.
```
BenchmarkRefcountMap
BenchmarkRefcountMap-4       	 6499250	       212 ns/op
BenchmarkSyncRefcountMap
BenchmarkSyncRefcountMap-4   	 4078995	       307 ns/op
```
Reworked `RefcountMap` + added bench tests.

pkg/tools/clientmap/refcount.go
```diff
@@ -52,12 +49,21 @@ func (r *RefcountMap) Store(key string, value networkservice.NetworkServiceClien
...
+	})
 	client := raw.(*refcountClient)
+	if !loaded {
+		return client.NetworkServiceClient, false
+	}
 
+	client.lock.RLock()
+	defer client.lock.RUnlock()
+
+	if atomic.LoadInt32(&client.count) <= 0 {
```
denis-tingajkin, line 64: Do we still need to use the atomic under the locker?

Bolodya1997: Yep, it is `RLock()` not `Lock()`.

denis-tingajkin: RLock means that you could safely _read_. So I see that it is still over synchronized.

denis-tingajkin (changes requested):

pkg/tools/clientmap/refcount.go
Bolodya1997: `RLock`, `Lock` here are not for reading and writing `client.count`. `RLock` is for updating `client.count`. `Lock` is for deleting `client`.
`RLock` doesn't guarantee that only 1 goroutine updates `client.count` at the moment and it is actually what we need for the updates to increase performance.
https://github.com/networkservicemesh/sdk/blob/fe5ced750db9551ac39a3780e494e4b3fed76154/pkg/tools/clientmap/refcount.go#L61-L67
`Lock` guarantees that no one would restore `client` after it as been deleted and so it does some guarantees:
https://github.com/networkservicemesh/sdk/blob/fe5ced750db9551ac39a3780e494e4b3fed76154/pkg/tools/clientmap/refcount.go#L113-L125

Bolodya1997: Reworked with simple `sync.Mutex` in another PR: https://github.com/networkservicemesh/sdk/pull/610.

denis-tingajkin (approved):
@Bolodya1997 Could you also update vpp-forwarder and vppagent-forwarder?

denis-tingajkin:
@edwarnicke Let us know if you have something in mind related to this changes.

PR #587, opened 2020-11-13
Fix interpose

Bolodya1997:
# Issue
`interpose.NewServer()` works incorrect for refresh, close scenarios. Existing test doesn't cover it.
# Solution
Remove bool switches, check _if interpose_ case with `connectionInfo.clientConnID`. Add valid test.

### DIFF
--- a/pkg/networkservice/common/interpose/server_test.go
+++ b/pkg/networkservice/common/interpose/server_test.go
@@ -14,131 +14,109 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package crossnse_test define a full nsmgr + cross connect NSE + Endpoint GRPC based test
 package interpose_test
 
 import (
 	"context"
+	"net/url"
+	"testing"
 
-	"github.com/networkservicemesh/sdk/pkg/tools/clienturlctx"
-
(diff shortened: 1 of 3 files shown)


haiodo (approved):
Looks good.

pkg/networkservice/common/interpose/server_test.go
```diff
@@ -14,131 +14,109 @@
...
-	})
-	require.Nil(t, err)
-	require.NotNil(t, reg)
+	// 3. Close
 
-	interposeRegName := reg.Name
-	interposeNSE := endpoint.NewServer(ctx, interposeRegName,
-		authorize.NewServer(),
-		TokenGenerator)
+	conn = conn.Clone()
```
denis-tingajkin, line 102: Can we add a chain element into the end to check that Close has been called on the last chain element(NSE)

denis-tingajkin (commented):
In general looks good

pkg/networkservice/common/interpose/server_test.go
Bolodya1997: Added `touchServer` to verify that every `Request`, `Close` is invoked on the last chain element.

denis-tingajkin (approved):

PR #590, opened 2020-11-18
Add metadata chain element

edwarnicke:


fkautz (approved):

denis-tingajkin:
@edwarnicke For what scenarios we need to use the metadata chain element?

edwarnicke:
metadata allows us to have information on a per Connection basis that is both accessible *within* the call from other elements in the chain and also persists *between* calls.

Example from [sdk-vpp's memif mechanism chain element](https://github.com/networkservicemesh/sdk-vpp/blob/main/pkg/networkservice/mechanisms/memif/common.go#L167) which checks to see if we *have* a swIfIndex already for this connection.

Corresponding example from [sdk-vpp's memif mechanism chain element](https://github.com/networkservicemesh/sdk-vpp/blob/main/pkg/networkservice/mechanisms/memif/common.go#L110) where we store it.

This turns out to be *insanely* useful as it both allows a fairly clean type safe collaboration amongst chain elements, but also allows a very simple clean way of managing our own state between calls for a class of information.

denis-tingajkin:
> metadata allows us to have information on a per Connection basis that is both accessible within the call from other elements in the chain and also persists between calls.

Why we can use mechanism parameters for this?

edwarnicke:
> > metadata allows us to have information on a per Connection basis that is both accessible within the call from other elements in the chain and also persists between calls.
> 
> Why we can't use mechanism parameters for this?

Hmm... that's an interesting idea :)

I see two issues with using mechanism parameters here:

1.  Mechanism parameters are semantically about communicating information between client and endpoint wrt what you need to know to construct the mechanism, not the internal implementation details (which things like swIfIndex are).
2.  We shouldn't be leaking implementation details out in Mechanism, which tends to be passed to other entities (client to server for example).
3.  Mechanism parameters are not persistent for a connection between calls.  If I get a Close call (or a reRequest) ... I need to know the swIfIndex I already created for the connection (in this example).

Issue #507, opened 2020-10-05
IPAM should be idempotent


xzfc:
## Current Behavior
IPAM allocates new IPs even if the request already has assigned IPs. If used together with `refresh`, this means that the endpoint would allocate a new pair of IPs on every `refresh` update.

## Expected Behavior
IPAM shouldn't allocate IPs if they are already present in the request connection. In practice, this means, IPAM should be idempotent.

## Steps to Reproduce

Add the following test to [`point2pointipam/server_test.go`](https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/ipam/point2pointipam/server_test.go).

```go
func TestRepeat(t *testing.T) {
	_, ipnet, err := net.ParseCIDR("192.168.3.4/16")
	require.NoError(t, err)
	srv := point2pointipam.NewServer(ipnet)

	req := newRequest()

	conn1, err := srv.Request(context.Background(), req.Clone())
	assert.NoError(t, err)
	assert.Equal(t, "192.168.0.0/32", conn1.Context.IpContext.DstIpAddr)

	req.Connection = conn1

	conn2, err := srv.Request(context.Background(), req.Clone())
	assert.NoError(t, err)
	assert.Equal(t, "192.168.0.0/32", conn2.Context.IpContext.DstIpAddr) // FAIL: actual value is "192.168.0.2/32"
}
```

```
--- FAIL: TestRepeat (0.00s)
    server_test.go:60: 
        	Error Trace:	server_test.go:60
        	Error:      	Not equal: 
        	            	expected: "192.168.0.0/32"
        	            	actual  : "192.168.0.2/32"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-192.168.0.0/32
        	            	+192.168.0.2/32
        	Test:       	TestRepeat
FAIL
exit status 1
FAIL	github.com/networkservicemesh/sdk/pkg/networkservice/ipam/point2pointipam	0.003s
```

PR #595, opened 2020-11-19
Rework ipam

Bolodya1997:
# Issue
1. Existing ipam works incorrectly - it returns IP addresses with `/32` subnet instead of `/31` and even more can return IP addresses not even from the one `/31` subnet.
2. Supported only p2p scenario, when we sometimes want to use ipam to generate IP addresses for some given subnet.
3. Refresh requests to the ipam would allocate new IP addresses instead of retrieving already allocated ones.
# Solution
1. Generate 2 subsequent IP addresses with `dstIP ^ srcIP == 1`.
2. If `request.Connection.Context.IpContext.DstIpRequired == false` work as IP net ipam.
3. Store connection info and check for updates. If there is no need to update, return previously selected IP addresses.


xzfc:
> 3. Store connection info and check for updates. If there is no need to update, return previously selected IP addresses.

Closes #507.

pkg/networkservice/ipam/ip_pool.go
```diff
@@ -0,0 +1,131 @@
...
+	"sync"
+
+	"github.com/RoaringBitmap/roaring"
+	"github.com/pkg/errors"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/cidr"
+)
+
+var (
+	p2pMask net.IPMask = []byte{0xFF, 0xFF, 0xFF, 0xFE}
```
denis-tingajkin, line 31: Can we make this a private function to prevent static data invalidation?

pkg/networkservice/ipam/server.go
```diff
@@ -0,0 +1,160 @@
...
+				ipPool:  ipPool,
+				srcAddr: srcAddr,
+				dstAddr: dstAddr,
+			}
+			s.connInfos.Store(conn.GetId(), connInfo)
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
```
denis-tingajkin, line 125: Consider moving this logic to a separate private method to reduce nesting complexity.

denis-tingajkin (commented):
In general, I like this direction.

@Bolodya1997 Could you also add a README.md file with some clarification about p2p, subnet cases? I guess this can be useful for others.

Bolodya1997:
Depends on https://github.com/networkservicemesh/sdk/pull/612.

pkg/networkservice/ipam/ip_pool.go
Bolodya1997: Yep, fixed.

pkg/networkservice/ipam/server.go
Bolodya1997: Yep, fixed.

Bolodya1997:
> In general, I like this direction.
> 
> @Bolodya1997 Could you also add a README.md file with some clarification about p2p, subnet cases? I guess this can be useful for others.

Yep, README.md added.

pkg/networkservice/ipam/README.md
```diff
@@ -0,0 +1,47 @@
...
+It is a server chain element implementing IPAM service. It works in 2 modes:
+
+### point to point
+
+```go
+conn, _ := ipam.NewServer("10.0.0.0/24").Request(ctx, &networkservice.NetworkServiceRequest{
+    Connection: &networkservice.Connection{
+        Context: &networkservice.ConnectionContext{
+            IpContext: &networkservice.IPContext{
+                DstIpRequired: true, // <-- p2p mode
```
denis-tingajkin, line 25: I'm not sure that it is the correct way to use `DstIpRequired` flag to select the ipam mode. Probably we can create a separate ipam chain element for subnets. 
@Bolodya1997 , @edwarnicke , @haiodo  Thoughts?

haiodo: I've had more generic question here, do we have an IPAM scenario without SrcIpRequired? Probable switch API to had enum values of required IPs? Thoughts?

denis-tingajkin: > I've had more generic question here, do we have an IPAM scenario without SrcIpRequired?

I think no, https://github.com/networkservicemesh/networkservicemesh/search?q=SrcIpRequired in monorepo as I can see no one handles these values.

> Probable switch API to had enum values of required IPs? 

I like idea with enum values  of required IPs

Issue #597, opened 2020-11-20
roundrobin should handle a case if selected endpoint returns error


denis-tingajkin:
## Actual behavior

Currently, `roundrobin` fails if the selected endpoint is busy or returns an error https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/roundrobin/server.go#L63

OR The same problem with other scerio:

1. bad NSE registered with ns-1
2. NSC requested NSE with ns-1
3. working NSE registered with ns-1 (in parallel with step 2)

**Actual**: NSC fails due to roundrobin selects bad NSE and not tries to select working NSE.

## Expected behaviour

`roundrobin` should try to select next endpoint if the previous one fails.

## Potential solution

1. `rounrobin` should iterate over all Candidates in round-robin sequence.
2. `discovery` should try to call next with new candidates if they've registered during request AND previous candidates fail AND ctx is not done.

PR #598, opened 2020-11-20
fix: common/roundrobin: should handle a case if selected endpoint returns error

denis-tingajkin:

## Motivation
https://github.com/networkservicemesh/sdk/issues/597

### DIFF
--- a/pkg/networkservice/common/roundrobin/server.go
+++ b/pkg/networkservice/common/roundrobin/server.go
@@ -72,5 +72,5 @@ func (s *selectEndpointServer) Close(ctx context.Context, conn *networkservice.C
 	if clienturlctx.ClientURL(ctx) != nil {
 		return next.Server(ctx).Close(ctx, conn)
 	}
-	return nil, errors.Errorf("inccorect connection: %+v", conn)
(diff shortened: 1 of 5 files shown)


pkg/networkservice/common/roundrobin/server.go
```diff
@@ -44,37 +44,33 @@ func NewServer() networkservice.NetworkServiceServer {
...
-		return nil, err
-	}
-	return next.Server(ctx).Close(ctx, conn)
-}
+	candidates := discover.Candidates(ctx)
 
-func (s *selectEndpointServer) withClientURL(ctx context.Context, conn *networkservice.Connection) (context.Context, error) {
-	if clienturlctx.ClientURL(ctx) == nil {
-		candidates := discover.Candidates(ctx)
+	for i := 0; i < len(candidates.Endpoints); i++ {
```
haiodo, line 52: Why no iterate till end of context deadline?

haiodo: ok, Denis answered. We had round robin per id, so we will iterate over all endpoints.

haiodo (approved):

PR #601, opened 2020-11-24
Replacing comparison of proto messages

### LINKED ISSUE #565: Do not compare protomessages via reflect.DeepEqual
## Problem

Since v1.25 proto messages contain private states. Example: https://github.com/networkservicemesh/api/blob/master/pkg/api/networkservice/connection.pb.go#L289-L291

So we should not use reflect.DeepEqual for proto messages as we do it in https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/core/trace/common.go#L59

## Steps to reproduce
```
func TestDeepEqual_PrivateProtoStates(t *testing.T) {
	p1 := new(networkservice.Path)
	p1.String()
	p2 := new(networkservice.Path)
	require.True(t, reflect.DeepEqual(p1, p2))
}
```

glazychev-art:

Fixes #565

### DIFF
--- a/pkg/tools/matchutils/utils.go
+++ b/pkg/tools/matchutils/utils.go
@@ -18,8 +18,10 @@
 package matchutils
 
 import (
-	"reflect"
 	"strings"
+
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/networkservicemesh/api/pkg/api/registry"
 )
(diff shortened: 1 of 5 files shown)


pkg/tools/matchutils/utils.go
```diff
@@ -18,29 +18,47 @@
...
 	return (left.Name == "" || strings.Contains(right.Name, left.Name)) &&
-		(left.NetworkServiceLabels == nil || reflect.DeepEqual(left.NetworkServiceLabels, right.NetworkServiceLabels)) &&
+		(left.NetworkServiceLabels == nil || labelsContains(right.NetworkServiceLabels, left.NetworkServiceLabels)) &&
 		(left.ExpirationTime == nil || left.ExpirationTime.Seconds == right.ExpirationTime.Seconds) &&
 		(left.NetworkServiceNames == nil || contains(right.NetworkServiceNames, left.NetworkServiceNames)) &&
 		(left.Url == "" || strings.Contains(right.Url, left.Url))
 }
 
-func contains(what, where []string) bool {
+func labelsContains(where, what map[string]*registry.NetworkServiceLabels) bool {
```
denis-tingajkin, line 45: Could you cover this by tests?

glazychev-art: Yes, new commit was added

pkg/tools/matchutils/utils_test.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+		NetworkServiceLabels: labels,
+		ExpirationTime: &timestamp.Timestamp{
+			Seconds: 500,
+			Nanos:   0,
+		},
+	}
+}
+
+func TestOneNSELabelsMatching(t *testing.T) {
```
denis-tingajkin, line 57: Could you use memory registry server and cover this via Find API?

denis-tingajkin (approved):

PR #605, opened 2020-11-25
Set back nse@domain name on return from interdomainURLServer

Bolodya1997:
# Issue
interdomainURLServer updates `Request.Connection.NetworkServiceEndpointName` from `nse@domain.url` to `nse` and doesn't set it back to the returned `Connection`. If no one before in the chain saves full name, `@domain.url` will be lost and refresh request will be trying to find `nse` locally until timeout.
# Solution
Set `Connection.NetworkServiceEndpointName = interDomainNSEName` on return.

### DIFF
--- a/pkg/networkservice/chains/nsmgrproxy/server_test.go
+++ b/pkg/networkservice/chains/nsmgrproxy/server_test.go
@@ -86,9 +86,7 @@ func TestNSMGR_InterdomainUseCase(t *testing.T) {
 	// Simulate refresh from client.
 
 	refreshRequest := request.Clone()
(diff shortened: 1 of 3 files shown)


denis-tingajkin:
>  refresh request will be trying to find nse locally until timeout.

Not sure abot that. It is already covered by test: https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/chains/nsmgrproxy/server_test.go#L86

denis-tingajkin (approved):
OK, looks like it is needed for new connect approach.

PR #616, opened 2020-12-04
Switch from GODIRECT to GOPRIVATE to improve reliability of updates

edwarnicke:


edwarnicke:
GOPRIVATE=github.com/networkservicemesh instructs go get to not use the proxy for the actual networkservicemesh bits (which we have updated) but to use the proxy for the rest of dependencies (which we are getting failures on, see:
https://github.com/networkservicemesh/sdk/actions/runs/400924623 )

denis-tingajkin (approved):

PR #627, opened 2020-12-11
Rework client factory using Generator

Bolodya1997:
# Issue
`client.NewClientFactory` receives a list of `networkservice.NetworkServiceClient` that should be added to the new client chain and this `additionalFunctionality` list is the same chain element instances for all generated clients. If `clientURL` changes we create a new client, but we presume the old `Connection.Id`, so reusing chain elements from the old client can provide some tricky bugs.
# Solution
Instead of passing list of `networkservice.NetworkServiceClient` we can pass list of type:
```go
// Generator is a type for networkservice.NetworkServiceClient generator function
type Generator func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
```
so on every new client creation by the factory we will create new chain elements for that client. To make usage of such 
`Generator` easier we can add some common wrappers:
```go
// FromClient is a common networkservice.NetworkServiceClient wrapper to Generator
func FromClient(client networkservice.NetworkServiceClient) Generator {
	return func(_ context.Context, _ grpc.ClientConnInterface) networkservice.NetworkServiceClient {
		return client
	}
}

// FromConstructor is a common networkservice.NetworkServiceClient constructor wrapper to Generator
func FromConstructor(constructor func() networkservice.NetworkServiceClient) Generator {
	return func(_ context.Context, _ grpc.ClientConnInterface) networkservice.NetworkServiceClient {
		return constructor()
	}
}
```
If we create a new client that is supposed to be used as a part of connect client chain and it doesn't match common wrappers it should define its own `NewClientGenerator` method in addition to (or replacing) `NewClient`.
```go
func NewClientGenerator(t ...params) client.Generator {
	return func(_ context.Context, _ grpc.ClientConnInterface) networkservice.NetworkServiceClient {
		// ...
	}
}
```
# No more need for NewCrossConnectClientFactory
`client.NewCrossConnectClientFactory` can now be replaced with:
```go
client.NewClientFactory(
	name,
	onHeal,
	tokenGenerator,
	client.FromConstructor(mechanismtranslation.NewClient), // <-- now can be used in such way
	// additionalFunctionalityGenerators...
),
```

### DIFF
--- a/pkg/networkservice/chains/client/client.go
+++ b/pkg/networkservice/chains/client/client.go
@@ -80,34 +80,34 @@ func NewClient(
 	return rv
 }
 
-// Generator is a type for networkservice.NetworkServiceClient generator function
-type Generator func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
+// Supplier is a type for networkservice.NetworkServiceClient supplier function
+type Supplier func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
 
 // NewClientFactory - returns a (2.), (3.) cases func(cc grpc.ClientConnInterface) NSM client factory.
 func NewClientFactory(
 	name string,
 	onHeal *networkservice.NetworkServiceClient,
 	tokenGenerator token.GeneratorFunc,
-	additionalFunctionalityGenerators ...Generator,
+	additionalFunctionalitySuppliers ...Supplier,
 ) func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
 	return func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
 		var additionalFunctionality []networkservice.NetworkServiceClient
-		for _, additionalFunctionalityFactory := range additionalFunctionalityGenerators {
+		for _, additionalFunctionalityFactory := range additionalFunctionalitySuppliers {
 			additionalFunctionality = append(additionalFunctionality, additionalFunctionalityFactory(ctx, cc))
 		}
 		return NewClient(ctx, name, onHeal, tokenGenerator, cc, additionalFunctionality...)
 	}
 }
 
-// FromClient is a common networkservice.NetworkServiceClient wrapper to Generator
-func FromClient(client networkservice.NetworkServiceClient) Generator {
+// FromClient is a common networkservice.NetworkServiceClient wrapper to Supplier
+func FromClient(client networkservice.NetworkServiceClient) Supplier {
 	return func(_ context.Context, _ grpc.ClientConnInterface) networkservice.NetworkServiceClient {
 		return client
 	}
 }
 
-// FromConstructor is a common networkservice.NetworkServiceClient constructor wrapper to Generator
-func FromConstructor(constructor func() networkservice.NetworkServiceClient) Generator {
+// FromConstructor is a common networkservice.NetworkServiceClient constructor wrapper to Supplier
+func FromConstructor(constructor func() networkservice.NetworkServiceClient) Supplier {
 	return func(_ context.Context, _ grpc.ClientConnInterface) networkservice.NetworkServiceClient {
 		return constructor()
 	}
--- a/pkg/networkservice/chains/client/client.go
+++ b/pkg/networkservice/chains/client/client.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/authorize"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/heal"
-	"github.com/networkservicemesh/sdk/pkg/networkservice/common/mechanismtranslation"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/refresh"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/serialize"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/updatepath"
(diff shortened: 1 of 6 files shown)


pkg/networkservice/chains/client/client.go
```diff
@@ -74,19 +80,35 @@ func NewClient(ctx context.Context, name string, onHeal *networkservice.NetworkS
 	return rv
 }
 
-// NewCrossConnectClientFactory - returns a (2.) case func(cc grpc.ClientConnInterface) NSM client factory.
-func NewCrossConnectClientFactory(name string, onHeal *networkservice.NetworkServiceClient, tokenGenerator token.GeneratorFunc, additionalFunctionality ...networkservice.NetworkServiceClient) func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
+// Generator is a type for networkservice.NetworkServiceClient generator function
+type Generator func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
```
denis-tingajkin, line 84:
```suggestion
type Supplier func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
```

pkg/networkservice/chains/client/client.go
```diff
@@ -74,19 +80,35 @@ func NewClient(ctx context.Context, name string, onHeal *networkservice.NetworkS
...
+		return NewClient(ctx, name, onHeal, tokenGenerator, cc, additionalFunctionality...)
 	}
 }
 
-// NewClientFactory - returns a (3.) case func(cc grpc.ClientConnInterface) NSM client factory.
-func NewClientFactory(name string, onHeal *networkservice.NetworkServiceClient, tokenGenerator token.GeneratorFunc, additionalFunctionality ...networkservice.NetworkServiceClient) func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
-	return func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
-		return NewClient(ctx, name, onHeal, tokenGenerator, cc, additionalFunctionality...)
+// FromClient is a common networkservice.NetworkServiceClient wrapper to Generator
+func FromClient(client networkservice.NetworkServiceClient) Generator {
```
denis-tingajkin, line 103: Do we really need this? I guess we should not use already created clients.

pkg/networkservice/chains/nsmgr/server_test.go
```diff
@@ -457,15 +464,6 @@ func (c *counterServer) Close(ctx context.Context, connection *networkservice.Co
...
-func (c *busyEndpoint) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
-	return nil, errors.New("sorry, endpoint is busy, try again later")
-}
-
-func (c *busyEndpoint) Close(ctx context.Context, connection *networkservice.Connection) (*empty.Empty, error) {
-	return nil, errors.New("sorry, endpoint is busy, try again later")
-}
 func newBusyEndpoint() networkservice.NetworkServiceServer {
-	return new(busyEndpoint)
+	return injecterror.NewServer(errors.New("sorry, endpoint is busy, try again later"))
```
denis-tingajkin, line 468: Nice

pkg/networkservice/chains/client/client.go
Bolodya1997: We can have some stateless client chain elements, so there is no need to create a new instance for them for each client.
E.g.: 
https://github.com/networkservicemesh/sdk/blob/1f553739f0c327c154994322a4f9f5e09dce1763/pkg/networkservice/chains/nsmgr/server_test.go#L328

denis-tingajkin: you're correct, but I think it is a complex way. I feel we should not consider is a chain element stateless or not.

pkg/networkservice/chains/client/client.go
Bolodya1997: Renamed.

pkg/networkservice/chains/client/client.go
Bolodya1997: OK, fixed.

edwarnicke:
I'd suggest we simply eliminate client.NewClientFactory in favor of 
https://github.com/networkservicemesh/sdk-vpp/blob/main/pkg/networkservice/chains/xconnectns/server.go#L74-L74

denis-tingaikin:
@edwarnicke I like this suggestion, and I also have created an issue to prove it :) 
https://github.com/networkservicemesh/sdk/issues/727

@Bolodya1997 Please let's do https://github.com/networkservicemesh/sdk/issues/727 to prove that these changes are required or let's follow Ed's suggestion.

denis-tingaikin:
@Bolodya1997 Can we close this?

denis-tingaikin:
Blocked by https://github.com/networkservicemesh/sdk/issues/727

pkg/networkservice/chains/client/client.go
```diff
@@ -74,19 +80,28 @@ func NewClient(ctx context.Context, name string, onHeal *networkservice.NetworkS
...
 }
 
-// NewClientFactory - returns a (3.) case func(cc grpc.ClientConnInterface) NSM client factory.
-func NewClientFactory(name string, onHeal *networkservice.NetworkServiceClient, tokenGenerator token.GeneratorFunc, additionalFunctionality ...networkservice.NetworkServiceClient) func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
-	return func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
-		return NewClient(ctx, name, onHeal, tokenGenerator, cc, additionalFunctionality...)
+// FromConstructor is a common networkservice.NetworkServiceClient constructor wrapper to Supplier
+func FromConstructor(constructor func() networkservice.NetworkServiceClient) Supplier {
+	return func(_ context.Context, _ grpc.ClientConnInterface) networkservice.NetworkServiceClient {
+		return constructor()
```
edwarnicke, line 105: So why all this additional baggage when we can simply use the func(_ context.Context, _ grpc.ClientConnInterface) networkservice.NetworkServiceClient

PR #641, opened 2021-01-12
Loggers unification

### LINKED ISSUE #614: Unify logging and tracing API
## Problem statement

Currently, we have three ways to logging staff into the project.

1. Logging via  `log` pkg. https://github.com/networkservicemesh/sdk/tree/master/pkg/tools/log
2. Logging via `trace` pkg. https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/core/trace/context.go#L51
3. Logging via `spanhelper` pkg. https://github.com/networkservicemesh/sdk/tree/master/pkg/tools/spanhelper

The count of logging APIs could confuse the developers. It is also could be not clear what the package should be used in the concrete case. Besides our logging could be not dependent on a concrete logging provider(logurs).

## Solution

Implement and merge the logging logic of these packages into one.

The result of pkg should export the next API:

```go
package TODO

type Logger interface{
    Info(v ...interface{})
    Infof(format string, v ...interface{})
    Warn(v ...interface{})
    Warnf(format string, v ...interface{})
    Error(v ...interface{})
    Errorf(format string, v ...interface{})
    Fatal(v ...interface{})
    Fatalf(format string, v ...interface{})
  
    WithField(k, v interface{}) Logger
}

func Log(ctx context.Context) Logger {
    //TODO: implement the code here
}
```

glazychev-art:
Closes #614 
Logging is available using `logger` package.


pkg/networkservice/common/excludedprefixes/server.go
```diff
@@ -44,7 +45,7 @@ type excludedPrefixesServer struct {
 }
 
 func (eps *excludedPrefixesServer) init() {
-	logger := trace.Log(eps.ctx)
```
haiodo, line 47: For initialisation purpose please use direct logger from tools instead of taking if from context.

pkg/networkservice/common/timeout/server_test.go
```diff
@@ -182,7 +183,7 @@ func newConnectionsServer(t *testing.T) *connectionsServer {
 	}
 }
 
-func (s *connectionsServer) validator(open, closed int) func() bool {
+func (s *connectionsServer) validator(ctx context.Context, open, closed int) func() bool {
```
haiodo, line 186: I do not think ctx and logging is required here.

pkg/networkservice/core/chain/client.go
```diff
@@ -18,15 +18,15 @@ package chain
...
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/trace"
+	"github.com/networkservicemesh/sdk/pkg/tools/logger"
 )
 
 // NewNetworkServiceClient - chains together a list of networkservice.NetworkServiceClient with tracing
 func NewNetworkServiceClient(clients ...networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
-	if logrus.GetLevel() == logrus.TraceLevel {
+	if logger.IsTracingEnabled() {
```
haiodo, line 29: I think we need to create wrapped every time and just pass null logger, so it will be more easy to achieve turn on/off logging per individual request.

pkg/networkservice/core/chain/server.go
```diff
@@ -20,24 +20,24 @@ package chain
...
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/setlogoption"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/trace"
+	"github.com/networkservicemesh/sdk/pkg/tools/logger"
 )
 
 // NewNetworkServiceServer - chains together a list of networkservice.Servers with tracing
 func NewNetworkServiceServer(servers ...networkservice.NetworkServiceServer) networkservice.NetworkServiceServer {
-	if logrus.GetLevel() == logrus.TraceLevel {
+	if logger.IsTracingEnabled() {
```
haiodo, line 32: Same here.

pkg/networkservice/core/chain/server.go
```diff
@@ -20,24 +20,24 @@ package chain
...
 		return next.NewWrappedNetworkServiceServer(trace.NewNetworkServiceServer, servers...)
 	}
 	return next.NewNetworkServiceServer(servers...)
 }
 
 // NewNamedNetworkServiceServer - chains together a list of networkservice.Servers with tracing and name log option
 func NewNamedNetworkServiceServer(name string, servers ...networkservice.NetworkServiceServer) networkservice.NetworkServiceServer {
-	if logrus.GetLevel() == logrus.TraceLevel {
+	if logger.IsTracingEnabled() {
```
haiodo, line 40: And here.

pkg/networkservice/core/next/client.go
```diff
@@ -64,6 +66,7 @@ func (n *nextClient) Request(ctx context.Context, request *networkservice.Networ
 		if np := Client(ctx); np != nil {
 			nextParent = np
 		}
+		ctx = logger.WithLog(ctx)
```
haiodo, line 69: Why we override with null logger every time?

pkg/networkservice/core/next/server.go
```diff
@@ -64,6 +66,7 @@ func (n *nextServer) Request(ctx context.Context, request *networkservice.Networ
 		if np := Server(ctx); np != nil {
 			nextParent = np
 		}
+		ctx = logger.WithLog(ctx)
```
haiodo, line 69: Same why here.

pkg/networkservice/core/next/server.go
```diff
@@ -77,6 +80,7 @@ func (n *nextServer) Close(ctx context.Context, conn *networkservice.Connection)
...
```
haiodo, line 83: And here.

pkg/networkservice/core/trace/client.go
```diff
@@ -45,48 +47,47 @@ func (t *traceClient) Request(ctx context.Context, request *networkservice.Netwo
 	ctx = withTrace(ctx)
```
haiodo, line 47: Trace chain element should be aware of logger.IsTracingEnabled() and in future it could take value from metadata incoming context to turn off/on logging per request.

pkg/networkservice/core/trace/common.go
```diff
@@ -17,40 +17,41 @@
...
-func logRequest(span spanhelper.SpanHelper, request proto.Message) {
-	connInfo, ok := trace(span.Context())
+func logRequest(ctx context.Context, request proto.Message) {
+	connInfo, ok := trace(ctx)
 	if ok && !proto.Equal(connInfo.Request, request) {
 		if connInfo.Request != nil && connInfo.Request.ProtoReflect().Descriptor().FullName() == request.ProtoReflect().Descriptor().FullName() {
 			requestDiff, hadChanges := Diff(connInfo.Request.ProtoReflect(), request.ProtoReflect())
 			if hadChanges {
-				span.LogObject("request-diff", requestDiff)
+				logger.Log(ctx).Infof("request-diff=%v", requestDiff)
```
haiodo, line 36: We loose semantic here, opentracing and jaeger support working with object structure if object is passed as is to it, but here we convert it to string representation and will loose this fancy functionality on jaeger UI.

pkg/registry/core/chain/ns_registry.go
```diff
@@ -20,24 +20,24 @@ package chain
...
 	"github.com/networkservicemesh/sdk/pkg/registry/common/setlogoption"
 	"github.com/networkservicemesh/sdk/pkg/registry/core/next"
 	"github.com/networkservicemesh/sdk/pkg/registry/core/trace"
+	"github.com/networkservicemesh/sdk/pkg/tools/logger"
 )
 
 // NewNetworkServiceRegistryServer - creates a chain of servers
 func NewNetworkServiceRegistryServer(servers ...registry.NetworkServiceRegistryServer) registry.NetworkServiceRegistryServer {
-	if logrus.GetLevel() == logrus.TraceLevel {
+	if logger.IsTracingEnabled() {
```
haiodo, line 32: Same here, please move check into trace chain element.

pkg/tools/callback/server.go
```diff
@@ -65,7 +67,7 @@ func IdentityByAuthority(ctx context.Context) (string, error) {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
 		err := errors.New("Not metadata provided")
-		logrus.Error(err)
+		logger.Log(logruslogger.WithLog(context.Background())).Error(err)
```
haiodo, line 70: I think this is overusage, please leave logrus usage as Is.

pkg/tools/prefixpool/prefixpool_test.go
```diff
@@ -76,7 +78,7 @@ func TestExtractPrefixes_1_ipv4(t *testing.T) {
 	require.Nil(t, err)
 	require.Equal(t, 20, len(newPrefixes))
 	require.Equal(t, 4, len(prefixes))
-	logrus.Printf("%v", newPrefixes)
+	logger.Log(logger.WithLog(context.Background())).Infof("%v", newPrefixes)
```
haiodo, line 81: Over usage. Please use logrus directly or remove usage of logging.

pkg/tools/spire/start.go
```diff
@@ -32,22 +32,24 @@ import (
...
 
+func getLogrusEntry() *logrus.Entry {
+	return logrus.WithTime(time.Now())
+}
+
 // AddEntry - adds an entry to the spire server for parentID, spiffeID, and selector
 //            parentID is usually the same as the agentID provided to Start()
 func AddEntry(ctx context.Context, parentID, spiffeID, selector string) error {
 	cmdStr := "spire-server entry create -parentID %s -spiffeID %s -selector %s -registrationUDSPath %s/spire-registration.sock"
 	cmdStr = fmt.Sprintf(cmdStr, parentID, spiffeID, selector, spireRoot)
```
haiodo, line 49: Please construct just one entry instead of multiple.

haiodo (changes requested):

edwarnicke:
Why not rather than 'Log' and 'WithField' simply have 'Entry' to return the current entry.  From there you can log or add fields easily.  Simplifies the API.

pkg/tools/callback/connection.go
```diff
@@ -24,7 +24,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/sirupsen/logrus"
```
haiodo, line 27: Callback is a separate library, so it should use logrus I think.

pkg/tools/callback/server.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Doc.ai and/or its affiliates.
+// Copyright (c) 2020-2021 Doc.ai and/or its affiliates.
```
haiodo, line 1: I do not think is is required to change callback APIs.

pkg/tools/callback/server_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Doc.ai and/or its affiliates.
+// Copyright (c) 2020-2021 Doc.ai and/or its affiliates.
```
haiodo, line 1: Same here

pkg/networkservice/common/excludedprefixes/server.go
```diff
@@ -44,7 +47,7 @@ type excludedPrefixesServer struct {
...
+	_, log := logruslogger.New(context.Background())
```
denis-tingaikin, line 50: here can be passed context from first the request.

pkg/networkservice/common/excludedprefixes/server.go
```diff
@@ -44,7 +47,7 @@ type excludedPrefixesServer struct {
 }
 
 func (eps *excludedPrefixesServer) init() {
```
denis-tingaikin, line 49:
```suggestion
func (eps *excludedPrefixesServer) init(ctx context.Context) {
```

pkg/networkservice/common/externalips/server.go
```diff
@@ -64,20 +67,21 @@ func NewServer(chainCtx context.Context, options ...Option) networkservice.Netwo
 	for _, o := range options {
 		o(result)
 	}
+	chainCtxWithLog, log := logruslogger.New(chainCtx)
```
denis-tingaikin, line 70: Why we cannot put it into chain context in the place where chain context decorated?

pkg/networkservice/common/heal/client_test.go
```diff
@@ -110,12 +110,12 @@ func TestHealClient_Request(t *testing.T) {
 	}
 	_, err = client.Close(requestCtx, conn)
 	require.NoError(t, err)
-	logrus.Infof("passed with total time: %v", time.Since(t1))
+	ctx = logger.WithLog(ctx)
+	logger.Log(ctx).Infof("passed with total time: %v", time.Since(t1))
```
denis-tingaikin, line 114: Actually, it is not needed we can look at test time...
```suggestion
```

pkg/networkservice/common/excludedprefixes/server_test.go
```diff
@@ -56,7 +58,7 @@ func TestNewExcludedPrefixesService(t *testing.T) {
 	}))
 	req := request()
 
-	_, err := chain.Request(context.Background(), req)
+	_, err := chain.Request(logger.WithLog(context.Background()), req)
```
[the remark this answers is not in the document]
denis-tingaikin: Also, apply this for the same places.

denis-tingaikin (changes requested):

haiodo (approved):

pkg/networkservice/chains/nsmgr/server_test.go
```diff
@@ -44,15 +44,16 @@ import (
...
 	"github.com/networkservicemesh/sdk/pkg/tools/sandbox"
-	"github.com/networkservicemesh/sdk/pkg/tools/spanhelper"
 )
 
 func TestNSMGR_RemoteUsecase_Parallel(t *testing.T) {
 	defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
-	logrus.SetOutput(ioutil.Discard)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
+	ctx = logger.WithLog(ctx)
```
denis-tingaikin, line 55: Should we put a logger inside the sandbox to reduce changes?

pkg/networkservice/chains/nsmgr/server_test.go
```diff
@@ -400,12 +403,12 @@ func TestNSMGR_PassThroughRemote(t *testing.T) {
...
 func TestNSMGR_PassThroughLocal(t *testing.T) {
-	nsesCount := 7
-
 	defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
-	logrus.SetOutput(ioutil.Discard)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
+	ctx = logger.WithLog(ctx)
+
+	nsesCount := 7
```
denis-tingaikin, line 411:
```suggestion
	const nsesCount = 7
```

denis-tingaikin (commented):
Other things look good to me. Please look at the comments above.

pkg/networkservice/chains/nsmgr/server_test.go
glazychev-art: Yes, we can, but in this case we also need to create a logger inside sandbox/utils/NewEndpoint. 
You can take a look, PR is updated.

denis-tingaikin (approved):

PR #661, opened 2021-01-22
Interpose should create unique name for all new interpose NSEs

Bolodya1997:
# Issue
If multiple interpose NSEs will try to register with same names, we will have a collision and lose some of them.
# Solution
Generate new unique name for interpose NSEs if we haven't registered them yet before.

### DIFF
--- a/pkg/networkservice/chains/nsmgr/server_test.go
+++ b/pkg/networkservice/chains/nsmgr/server_test.go
@@ -513,6 +513,63 @@ func TestNSMGR_PassThroughLocal(t *testing.T) {
 	require.Equal(t, 5*(nsesCount-1)+5, len(conn.Path.PathSegments))
 }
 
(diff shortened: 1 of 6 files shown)


denis-tingaikin (changes requested):
Please cover this by sandbox tests to make sure that it is working as expected with other applications.

Bolodya1997:
Needs updated sandbox by #668.

Bolodya1997:
> Please cover this by sandbox tests to make sure that it is working as expected with other applications.

Done.

denis-tingaikin (commented):
Should we just add setid into interpose chain?

Bolodya1997:
> Should we just add setid into interpose chain?

We cannot use `setid` because it would overwrite `interpose-nse#` prefix.

PR #662, opened 2021-01-22
Fix setid

Bolodya1997:
# Issue
`setid` registry chain element doesn't generate new name if NSE already has name.
# Solution
Create unique names for all newly registrated endpoints.


denis-tingaikin (changes requested):
Please cover this by sandbox tests to make sure that it is working as expected with other applications.

pkg/registry/common/setid/gen.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Doc.ai and/or its affiliates.
+// Copyright (c) 2021 Doc.ai and/or its affiliates.
```
haiodo, line 1:
```suggestion
// Copyright (c) 2020-2021 Doc.ai and/or its affiliates.
```

Bolodya1997: Fixed.

Bolodya1997:
Needs updated sandbox by #668.

pkg/registry/common/setid/gen.go
Bolodya1997: Yep, fixed.

Bolodya1997:
> Please cover this by sandbox tests to make sure that it is working as expected with other applications.

Done.

pkg/registry/common/setid/server.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// NewNetworkServiceEndpointRegistryServer creates new instance of NetworkServiceRegistryServer which set the unique
+// name for the endpoint on registration
+func NewNetworkServiceEndpointRegistryServer() registry.NetworkServiceEndpointRegistryServer {
+	return new(setIDServer)
+}
+
+func (s *setIDServer) Register(ctx context.Context, nse *registry.NetworkServiceEndpoint) (*registry.NetworkServiceEndpoint, error) {
+	name, suffix := interDomainName(nse.Name)
```
denis-tingaikin, line 42: Why setid knows about interdomain?

Bolodya1997: Because if we want to register NSE in some other domain, we shouldn't override domain suffix:
https://github.com/networkservicemesh/sdk/blob/536f38f8986817e62259be5f258203006ef96a7e/pkg/registry/core/interdomain/interdomain_nse_test.go#L239-L246

denis-tingaikin: Yes, but this chain element shouldn't know about the interdomain.

Bolodya1997: Fixed with id prefix.

denis-tingaikin (approved):

Issue #657, opened 2021-01-18
Reorganize nsmgr unit tests to use testify suite and fix all found issues


denis-tingaikin:
## Motivation

Our integration tests reusing nsmgr and forwarder instances for each test in the suite. Most of the issues that we are finding can be detected and fixed in a more easy way: we need to reorganize nsmgr unit tests in sdk. https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/chains/nsmgr/server_test.go

## Solution
1. Move tests from https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/chains/nsmgr/server_test.go  to another pkg for use cases
2. Make each test in https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/chains/nsmgr/server_test.go  as part of a [testify suite](https://godoc.org/github.com/stretchr/testify/suite).
3. Fix found issues in separate PRs.

**Benefits**:

1. Unit tests will work faster because each test will reuse nsmgr/forwarder
2. We cover and fix the most popular bugs for the integration repo.

PR #663, opened 2021-01-22
Use suites for nsmgr tests

glazychev-art:
Fixes: #657 

### DIFF
--- a/pkg/networkservice/chains/nsmgr/nsmgr_test.go
+++ b/pkg/networkservice/chains/nsmgr/nsmgr_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
 	"go.uber.org/goleak"
 	"google.golang.org/grpc"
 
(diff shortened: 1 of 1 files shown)


denis-tingaikin:
Why is this draft?

edwarnicke:
@glazychev-art Can this be closed?

denis-tingaikin:
@glazychev-art Could you rebase this?

pkg/networkservice/chains/nsmgr/server_heal_test.go
```diff
@@ -201,7 +201,7 @@ func testNSMGRHealForwarder(t *testing.T, nodeNum int, restored bool, customConf
 		NetworkServiceNames: []string{"my-service-remote"},
 	}
 
-	counter := &counterServer{}
+	counter := sandbox.NewCounterServer()
```
denis-tingaikin, line 204: Probably better move this into networkservice/tools pkg

pkg/networkservice/chains/usecase/usecase_test.go
```diff
@@ -234,50 +228,42 @@ func TestNSMGR_RemoteUsecase_BusyEndpoints(t *testing.T) {
...
+	// Endpoint unregister
+	for i := 0; i < len(nsesReg); i++ {
+		_, err = s.domain.Nodes[1].EndpointRegistryClient.Unregister(ctx, nsesReg[i])
+		require.NoError(t, err)
+	}
 }
 
-func TestNSMGR_RemoteUsecase(t *testing.T) {
-	t.Cleanup(func() { goleak.VerifyNone(t) })
+func (s *nsmgrSuite) TestRemote() {
```
denis-tingaikin, line 245: I would be suggesting to focus on use cases. So each test for nsmgr suite should have the suffix Usecase and it should test a real use-case :)
```suggestion
func (s *nsmgrSuite) TestRemoteUsecase() {
```
Also apply this for other tests.

denis-tingaikin (commented):
In general, looks good to me. Added few comments.

pkg/networkservice/chains/usecase/usecase_test.go
```diff
@@ -46,294 +47,264 @@ import (
...
-		SetNodesCount(2).
+	domain *sandbox.Domain
+}
+
+func TestNsmgr(t *testing.T) {
+	suite.Run(t, new(nsmgrSuite))
+}
+
+func (s *nsmgrSuite) SetupSuite() {
+	// Provide 5 seconds per test
```
Bolodya1997, line 71: I suppose here we can use simple:
```go
ctx, cancel := context.WithCancel(context.Background())
defer cancel()
```
because we have a timeout for each test in suite.

pkg/networkservice/chains/usecase/usecase_test.go
```diff
@@ -46,294 +47,264 @@ import (
...
 	require.NoError(t, err)
 	require.NotNil(t, conn)
 	require.Equal(t, 8, len(conn.Path.PathSegments))
-	require.Equal(t, int32(2), atomic.LoadInt32(&counter.Requests))
-
-	// Close.
+	require.Equal(t, int32(2), atomic.LoadInt32(&counterServer.Requests))
 
+	// Close
 	e, err := nsc.Close(ctx, conn)
```
Bolodya1997, line 123: Probably that would be better:
```go
_, err = nsc.Close(ctx, conn) // <-- it is incorrect to check returned *empty.Empty value
```

pkg/networkservice/chains/usecase/usecase_test.go
```diff
@@ -46,294 +47,264 @@ import (
...
+	require.NoError(t, err)
 }
 
-func TestNSMGR_ConnectToDeadNSE(t *testing.T) {
-	t.Cleanup(func() { goleak.VerifyNone(t) })
+func (s *nsmgrSuite) Test_ConnectToDeadNSEUsecase() {
+	t := s.T()
+
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	nseCtx, killNse := context.WithTimeout(context.Background(), time.Second*5)
```
Bolodya1997, line 292: Probably it would be better to base all new contexts in test on `ctx`:
```go
nseCtx, killNse := context.WithCancel(ctx)
```

pkg/networkservice/chains/usecase/usecase_test.go
```diff
@@ -520,9 +467,44 @@ func TestNSMGR_PassThroughLocal(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+
+	// We have lazy initialization in some chain elements in both networkservice, registry chains. So registering an
+	// endpoint and requesting it from client can result in new endless NSMgr goroutines.
+	testNSEAndClient(ctx, t, s.domain, &registry.NetworkServiceEndpoint{
+		Name:                "endpoint-init",
+		NetworkServiceNames: []string{"service-init"},
+	})
+
```
Bolodya1997, line 495: This test was broken by https://github.com/networkservicemesh/sdk/pull/806/commits/725e8dbed932349dcf0d4566b67b4f52f5bde83d.
Please add here back check for the goroutines leaks:
```go
defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
```

denis-tingaikin: This is not related to this PR.

Bolodya1997 (approved):

denis-tingaikin (commented):
This PR moving nsmgr tests into suits but as I can see in code all nsmgr suits going to usecase pkg. 

I think if we have only nsmgr suite after refactoring then we can keep it in `chains/nsmgr`. 

What do you think?

denis-tingaikin (approved):

Issue #622, opened 2020-12-07
Traced chain element prints the wrong diff


denis-tingaikin:
## Steps to reproduce

1. Create a chain of `discover` chain element https://github.com/networkservicemesh/sdk/tree/master/pkg/networkservice/common/discover.
2. Call request.
3. Check diff in logs.

Actual: diff is path and context fields. Example
```
Dec  7 20:42:07.395 [TRAC] [name:nsmgr-k9gk5] (6)      ⎆ sdk/pkg/networkservice/common/discover/discoverCandidatesServer.Request() span=472a05d52d72bd1d:1a4823734812f42a:7b6bdf0f0531219e:1
Dec  7 20:42:07.395 [TRAC] [name:nsmgr-k9gk5] (6.1)        request-diff={"connection":{"path":{"path_segments":{"3":{"expires":{"seconds":1607377140},"token":"eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJzcGlmZmU6Ly9leGFtcGxlLm9yZy9ucy9uc20tc3lzdGVtL3NhL2RlZmF1bHQiLCJleHAiOjE2MDczNzcxNDAsInN1YiI6InNwaWZmZTovL2V4YW1wbGUub3JnL25zL25zbS1zeXN0ZW0vc2EvZGVmYXVsdCJ9.NS1zk8d7TDAptczHopx0j0IR2oTbeMDp1I-G3Ue4yEE8SQvjm51FmUCNX_svJekCHIQy5OJCRW45h9b7Hy9R8A"}}}}} span=472a05d52d72bd1d:1a4823734812f42a:7b6bdf0f0531219e:1
```

Expected: dicover changes only NS, NSE name, diff should be in format "{ns=ns1, nse=nse1}"

PR #667, opened 2021-01-25
Fix trace diffs

glazychev-art:
Fixes: #622 

### DIFF
--- a/pkg/networkservice/core/trace/client.go
+++ b/pkg/networkservice/core/trace/client.go
@@ -27,24 +27,26 @@ import (
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 	"github.com/networkservicemesh/sdk/pkg/tools/logger"
-
 	"github.com/networkservicemesh/sdk/pkg/tools/typeutils"
 )
 
-type traceClient struct {
+type beginTraceClient struct {
 	traced networkservice.NetworkServiceClient
 }
 
+type endTraceClient struct{}
+
 // NewNetworkServiceClient - wraps tracing around the supplied networkservice.NetworkServiceClient
 func NewNetworkServiceClient(traced networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
-	if traced == nil {
-		return &traceClient{traced: traced}
-	}
-	return &traceClient{traced: traced}
+	return next.NewNetworkServiceClient(
+		&beginTraceClient{traced: traced},
+		&endTraceClient{},
+	)
 }
 
-func (t *traceClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+func (t *beginTraceClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
 	// Create a new logger
 	operation := typeutils.GetFuncName(t.traced, "Request")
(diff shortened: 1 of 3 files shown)


pkg/networkservice/core/next/tail_server.go
```diff
@@ -28,6 +30,10 @@ import (
 // to insure that we never call a method on a nil object
 type tailServer struct{}
 
+func newWrappedTailServer(wrapper ServerWrapper) networkservice.NetworkServiceServer {
```
denis-tingaikin, line 33: We should not modify next to fix this issue. 
`next` should not know anything about logging.

denis-tingaikin (changes requested):
Can we inject into the tail of the servers/clients in `chain` some element to detect that we achieved the tail?

pkg/networkservice/core/trace/client.go
```diff
@@ -27,21 +27,35 @@ import (
...
+func (t *printerClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+	logRequest(ctx, request)
+	conn, err := next.Client(ctx).Request(ctx, request, opts...)
+	logResponse(ctx, conn)
+	return conn, err
+}
+
+func (t *printerClient) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
+	logRequest(ctx, conn)
+	return next.Client(ctx).Close(ctx, conn, opts...)
```
denis-tingaikin, line 50: Why we do not log response here?

glazychev-art: I think because `Close(...)` returns `protobuf.Empty`. There is nothing for logging.

denis-tingaikin: Yes, but can we try to log changes in conn?

glazychev-art: Probably you are right.
Do you think there might be the next implementation of Close?
```
func Close(...) {
	// some code
	r, err := next.Server(ctx).Close(ctx, connection)
	// changes inside connection here
	return r, err
}
```

pkg/networkservice/core/trace/client.go
```diff
@@ -50,8 +64,6 @@ func (t *traceClient) Request(ctx context.Context, request *networkservice.Netwo
 	ctx, finish := withLog(ctx, operation)
 	defer finish()
 
-	logRequest(ctx, request)
```
haiodo, line 53: You still need to remember current request before it will be modified with traced.
```connInfo.Request = proto.Clone(request)```

**Review by haiodo — APPROVED** [review]

File: pkg/networkservice/core/trace/client.go [hunk 4]
```diff
@@ -27,21 +27,37 @@ import (
...
+
 // NewNetworkServiceClient - wraps tracing around the supplied networkservice.NetworkServiceClient
 func NewNetworkServiceClient(traced networkservice.NetworkServiceClient) networkservice.NetworkServiceClient {
-	if traced == nil {
-		return &traceClient{traced: traced}
-	}
-	return &traceClient{traced: traced}
+	return next.NewNetworkServiceClient(
+		&traceClient{traced: traced},
+		&printerClient{},
```
denis-tingaikin, line 59:
```suggestion
		&beginTraceClient{traced: traced},
		&endTraceClient{},
```

denis-tingaikin (approved):
In general looks good

pkg/networkservice/core/trace/client.go
glazychev-art: Done

PR #677, opened 2021-01-27
Don't call resolver if domain is already resolved

Bolodya1997:
# Motivation
There is no need to call resolver if domain is already in `tcp://${addr}:${port}` format.

### DIFF
--- a/pkg/registry/common/dnsresolve/common.go
+++ b/pkg/registry/common/dnsresolve/common.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Doc.ai and/or its affiliates.
+// Copyright (c) 2020-2021 Doc.ai and/or its affiliates.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"regexp"
 )
 
 // Resolver is DNS resolver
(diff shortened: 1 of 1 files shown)


pkg/registry/common/dnsresolve/common.go
```diff
@@ -37,34 +38,48 @@ type Resolver interface {
...
 	}
 
-	if len(ips) == 0 {
-		return nil, errors.New("resolver.LookupIPAddr return empty result")
-	}
+	return u, nil
+}
 
-	u, err := url.Parse(fmt.Sprintf("tcp://%v:%v", ips[0].IP, records[0].Port))
+var resolvedPattern = regexp.MustCompile("tcp://(?P<addr>[0-9.:]+):(?P<port>[0-9]+)?")
```
denis-tingaikin, line 73: Can we just try to parse it as a URL?

denis-tingaikin (changes requested):

pkg/registry/common/dnsresolve/common.go
Bolodya1997: Nope, any domain is most possible a valid URL. We need a domain in `tcp://ip:port` format.

pkg/registry/common/dnsresolve/common.go
```diff
@@ -37,34 +38,48 @@ type Resolver interface {
 }
 
 func resolveDomain(ctx context.Context, service, domain string, r Resolver) (*url.URL, error) {
-	serviceDomain := fmt.Sprintf("%v.%v", service, domain)
-	_, records, err := r.LookupSRV(ctx, service, "tcp", serviceDomain)
+	var resolved string
+	if isResolved(domain) {
```
denis-tingaikin, line 42: please move this check to top-level before call resolve domain.

pkg/registry/common/dnsresolve/common.go
denis-tingaikin: Lets then check that part in URL after the scheme is a valid IP and PORT.

You could split Path into two strings entry to parse parts with default parsers for IP and int

denis-tingaikin (changes requested):

denis-tingaikin (approved):

PR #682, opened 2021-01-28
Fix NS watch

### LINKED ISSUE #619: `TestNSMGR_RemoteUsecase_BusyEndpoints` has goroutine leaks sometimes
https://github.com/networkservicemesh/sdk/pull/612/checks?check_run_id=1508552180
```
--- FAIL: TestNSMGR_RemoteUsecase_BusyEndpoints (1.33s)
Error:     leaks.go:78: found unexpected goroutines:
        [Goroutine 638 in state chan send, with github.com/networkservicemesh/api/pkg/api/registry.ReadNetworkServiceChannel.func1 on top of the stack:
        goroutine 638 [chan send]:
        github.com/networkservicemesh/api/pkg/api/registry.ReadNetworkServiceChannel.func1(0x11b1760, 0xc0003ff440, 0xc00067e5a0)
        	/home/runner/go/pkg/mod/github.com/networkservicemesh/api@v0.0.0-20201204203731-4294f67deaa4/pkg/api/registry/helpers.go:54 +0x7a
        created by github.com/networkservicemesh/api/pkg/api/registry.ReadNetworkServiceChannel
        	/home/runner/go/pkg/mod/github.com/networkservicemesh/api@v0.0.0-20201204203731-4294f67deaa4/pkg/api/registry/helpers.go:52 +0x85
        ]
```

Bolodya1997:
# Issue
`discover` chain element don't read fully channel from `registry.ReadNetworkServiceEndpointChannel` - it results in #619.
# Solution
Use GRPC Stream instead of channels.

Closes #619.

### DIFF
--- a/pkg/networkservice/chains/nsmgr/server_test.go
+++ b/pkg/networkservice/chains/nsmgr/server_test.go
@@ -149,8 +149,6 @@ func TestNSMGR_SelectsRestartingEndpoint(t *testing.T) {
 }
 
 func TestNSMGR_RemoteUsecase_BusyEndpoints(t *testing.T) {
-	t.Skip("https://github.com/networkservicemesh/sdk/issues/619")
-
 	defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
 	defer cancel()
--- a/pkg/networkservice/common/discover/server.go
+++ b/pkg/networkservice/common/discover/server.go
@@ -102,70 +102,62 @@ func (d *discoverCandidatesServer) discoverNetworkServiceEndpoint(ctx context.Co
 			Name: nseName,
 		},
 	}
+
 	nseStream, err := d.nseClient.Find(ctx, query)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 	nseList := registry.ReadNetworkServiceEndpointList(nseStream)
+
 	if len(nseList) != 0 {
 		return nseList[0], nil
 	}
+
 	query.Watch = true
+
 	nseStream, err = d.nseClient.Find(ctx, query)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-	nseCh := registry.ReadNetworkServiceEndpointChannel(nseStream)
-	select {
-	case <-ctx.Done():
-		return nil, errors.Wrapf(ctx.Err(), "nse: %+v is not found", query.NetworkServiceEndpoint)
-	case nse, ok := <-nseCh:
-		if ok {
-			return nse, nil
-		}
-		return nil, errors.New("nse stream is closed")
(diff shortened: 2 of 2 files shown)


edwarnicke:
If you don't send updates on re-regsister... how do the things receiving notification know that the Registry entry has not expired?

edwarnicke:
I also don't see how this is causing goroutine leaks...

Bolodya1997:
> I also don't see how this is causing goroutine leaks...

`discover` receives only first updates and the others are getting stuck in channel:
https://github.com/networkservicemesh/sdk/blob/1759d4b9155be7921850f6d8fbe42c8e407aba77/pkg/networkservice/common/discover/server.go#L195-L203

Bolodya1997:
> If you don't send updates on re-regsister... how do the things receiving notification know that the Registry entry has not expired?

I have definitely missed this case :)

Bolodya1997:
> If you don't send updates on re-regsister... how do the things receiving notification know that the Registry entry has not expired?

I believe this is OK - expire is watching only for NSE updates.
https://github.com/networkservicemesh/sdk/blob/1759d4b9155be7921850f6d8fbe42c8e407aba77/pkg/registry/common/expire/ns_server.go#L62-L69

edwarnicke:
So I guess the real question is: why are we passing this all via channels to discover instead of just having discover use the existing Registry API, which would let it simply use the Find without Watch semantics...

Bolodya1997:
> So I guess the real question is: why are we passing this all via channels to discover instead of just having discover use the existing Registry API, which would let it simply use the Find without Watch semantics...

If there is no NS in registry, discover blocks until the required NS appears. So I guess it is why we need Watch semantics.

denis-tingaikin:
@Bolodya1997 
I'm not quite following this. I feel the root cause can be related to memory.Find

edwarnicke:
> > So I guess the real question is: why are we passing this all via channels to discover instead of just having discover use the existing Registry API, which would let it simply use the Find without Watch semantics...
> 
> If there is no NS in registry, discover blocks until the required NS appears. So I guess it is why we need Watch semantics.

So you'd use Find with Watch==true and when you got a response, close the FindClient so as to stop getting future updates.

It feels like we are making problems for ourselves by introducing new channel semantics when we have really *go* semantics for this purpose in the existing API.

denis-tingaikin:
@edwarnicke , @Bolodya1997 

The problem can not be related to API helper functions. The real issue is wrong _using of these methods_.

denis-tingaikin:
@edwarnicke , @Bolodya1997  I feel this may fix the issue https://github.com/networkservicemesh/api/pull/78

denis-tingaikin (approved):
This looks good. 

For this case, I agreed with that stream.Recv looks more simple.
For other cases, networkservicemesh/api#78  will fix the same issue.

So merging.

PR #684, opened 2021-01-29
Fix: expire NS server incorrectly handles updates from the NSE server

denis-tingaikin:

## Motivation

1. ns `expire` server should correctly handle a case when remote registry sent an update with deleted nse.
2. ns `expire` server should correctly count refs of network services in case of nse updates.

Also added/improved unit tests to cover both scenarios.


pkg/registry/common/expire/ns_server.go
```diff
@@ -30,88 +29,135 @@ import (
...
+			timer, ok := state.Timers[nse.Name]
+			expirationDuration := time.Until(nse.ExpirationTime.AsTime().Local())
+			if !ok {
+				if expirationDuration > 0 {
+					state.Timers[nse.Name] = time.AfterFunc(expirationDuration, func() {
+						state.Lock()
+						ctx := state.Context
+						delete(state.Timers, nse.Name)
+						state.Unlock()
 						_, _ = n.Unregister(ctx, &registry.NetworkService{Name: ns})
```
Bolodya1997, line 69: I believe there should be check for `len(state.Timers) == 0`.

pkg/registry/common/expire/ns_server.go
```diff
@@ -30,88 +29,135 @@ import (
...
+					})
+					state.Unlock()
+				}
+				continue
+			}
+
+			if expirationDuration < 0 {
+				delete(state.Timers, nse.Name)
+				if timer.Stop() {
+					if len(state.Timers) == 0 {
```
Bolodya1997, line 79: I suppose we don't need such check here because we already have it in `Unregister`.

pkg/registry/common/expire/ns_server.go
```diff
@@ -30,88 +29,135 @@ import (
...
+	}
+
+	valuesCtx := extend.WithValuesFromContext(n.chainCtx, ctx)
+
+	v, _ := n.nsStates.LoadOrStore(request.Name, &nsState{
+		Timers: make(map[string]*time.Timer),
+	})
+
+	v.Lock()
+	v.Context = valuesCtx
```
Bolodya1997, line 127: Do we need to overwrite context on every new NS registered?

denis-tingaikin: I think yes, the chain can be changed.

pkg/registry/common/expire/ns_server.go
denis-tingaikin: It is free to check because we're underlock here.

Bolodya1997: So then we should possibly add the same check to the `time.AfterFunc(...)` to make it look consistent?

Bolodya1997 (approved):

pkg/registry/common/expire/ns_server.go
```diff
@@ -30,88 +29,133 @@ import (
...
 			ns := nse.NetworkServiceNames[i]
-			var value int32
-			stored, _ := n.nsCounts.LoadOrStore(ns, &value)
-			if nse.ExpirationTime != nil && nse.ExpirationTime.Seconds < 0 {
-				atomic.AddInt32(stored, -1)
-			} else {
-				atomic.AddInt32(stored, 1)
+			state, ok := n.nsStates.Load(ns)
+			if !ok {
+				return
```
haiodo, line 57:
```suggestion
				continue
```

pkg/registry/common/expire/ns_server.go
```diff
@@ -30,88 +29,133 @@ import (
...
+				continue
 			}
+
+			if expirationDuration < 0 {
+				delete(state.Timers, nse.Name)
+				if timer.Stop() {
+					ctx := state.Context
+					state.Unlock()
+					_, _ = n.Unregister(ctx, &registry.NetworkService{Name: ns})
+					return
```
haiodo, line 82: You need to continue go over:
```for i := 0; i < len(nse.NetworkServiceNames); i++ {```

pkg/registry/common/expire/ns_server.go
denis-tingaikin: Thanks, fixed

pkg/registry/common/expire/ns_server.go
denis-tingaikin: Thanks, fixed

haiodo (approved):

PR #686, opened 2021-01-30
Fix go leak in registry refresh chain element

artem-belov:


pkg/registry/common/refresh/nse_registry_client.go
```diff
@@ -38,10 +38,10 @@ type refreshNSEClient struct {
 
 // NewNetworkServiceEndpointRegistryClient creates new NetworkServiceEndpointRegistryClient that will refresh expiration
 // time for registered NSEs
-func NewNetworkServiceEndpointRegistryClient(options ...Option) registry.NetworkServiceEndpointRegistryClient {
+func NewNetworkServiceEndpointRegistryClient(ctx context.Context, options ...Option) registry.NetworkServiceEndpointRegistryClient {
```
denis-tingaikin, line 41: Please use the option to pass the context.

denis-tingaikin (changes requested):

PR #688, opened 2021-02-01
Fix NSE memory registry server data race

Bolodya1997:
# Issue
NSE memory registry server doesn't clone when sending update, it results in data races.
# Solution
Clone endpoints on sending update.


denis-tingaikin (commented):
Should we put this with a test that reproduces this?

PR #689, opened 2021-02-01
Fix data race in NSE expire registry server

Bolodya1997:
# Issue
On low expire durations (100 ms) timer in expire registry server can be reset before its previous ran has been finished, it results in data races.
# Solution
Use synchronization, rework expire registry server.


pkg/registry/common/expire/nse_server_test.go
```diff
@@ -121,6 +125,35 @@ func TestExpireNSEServer_ShouldRemoveNSEAfterExpirationTime(t *testing.T) {
 	}, time.Second, time.Millisecond*100)
 }
 
+func TestExpireNSEServer_DataRace(t *testing.T) {
```
denis-tingaikin, line 128: This datarace fixes by more simple patch
```
diff --git a/pkg/registry/common/expire/nse_server.go b/pkg/registry/common/expire/nse_server.go
index a92044d..20b41bb 100644
--- a/pkg/registry/common/expire/nse_server.go
+++ b/pkg/registry/common/expire/nse_server.go
@@ -56,11 +56,10 @@ func (n *expireNSEServer) Register(ctx context.Context, nse *registry.NetworkSer
                defer cancel()
                _, _ = next.NetworkServiceEndpointRegistryServer(unregisterCtx).Unregister(unregisterCtx, unregisterNSE)
        })
-       if t, load := n.timers.LoadOrStore(resp.Name, timer); load {
+       if timer, ok := n.timers.LoadAndDelete(nse.Name); ok {
                timer.Stop()
-               t.Stop()
-               t.Reset(time.Until(expirationTime))
        }
+       n.timers.Store(resp.Name, timer)
 
        return resp, nil
 }
@@ -75,7 +74,7 @@ func (n *expireNSEServer) Unregister(ctx context.Context, nse *registry.NetworkS
                return nil, err
        }
 
-       if timer, ok := n.timers.Load(nse.Name); ok {
+       if timer, ok := n.timers.LoadAndDelete(nse.Name); ok {
                timer.Stop()
        }
 
```

denis-tingaikin (changes requested):

pkg/registry/common/expire/nse_server_test.go
Bolodya1997: Yes, it fixes data race but there comes issue with possibly not stopped timers that can execute after the successful refresh:
```
1. timer fires unregister
2. refresh comes
3. refresh reaches memory
4. unregister reaches memory
```
in such case we will have no entry in memory right after the refresh happen.

pkg/registry/common/expire/nse_server.go
```diff
@@ -20,71 +20,128 @@ import (
...
+		if stopped {
+			// Timer has been stopped, we need only to reset it.
+			t.timer.Reset(time.Until(expirationTime))
+		} else if loaded && !started {
+			// Timer function has been stopped with the `canceled` flag, we need to remove the flag.
+			t.executor.AsyncExec(func() {
+				t.canceled = false
+				if t.started {
+					// Timer function has been already finished, we need to reset the timer.
+					t.timer.Reset(time.Until(expirationTime))
```
haiodo, line 81: expirationTime is not defined for this branch.

pkg/registry/common/expire/nse_server.go
```diff
@@ -20,71 +20,128 @@ import (
...
 	if resp.ExpirationTime != nil {
 		if respExpirationTime := resp.ExpirationTime.AsTime().Local(); respExpirationTime.Before(expirationTime) {
 			expirationTime = respExpirationTime
 		}
 	}
 	resp.ExpirationTime = timestamppb.New(expirationTime)
 
-	unregisterNSE := resp.Clone()
+	t = n.newTimer(ctx, expirationTime, resp.Clone())
+	n.timers.Store(resp.Name, t)
```
haiodo, line 97: What if two requests will Register with same Endpoint at same time, is it will be one timer lost?

pkg/registry/common/expire/nse_server.go
Bolodya1997: Yep, replaced with `0`.

pkg/registry/common/expire/nse_server.go
Bolodya1997: Yep, fixed.

haiodo (approved):

denis-tingaikin (commented):
Currently, I  not like the idea of adding a new entity `unregisterTimer` 

Could you explain why do we need this?

Bolodya1997:
> Currently, I not like the idea of adding a new entity unregisterTimer
> 
> Could you explain why do we need this?
1. Timer needs to be fully stopped **before** passing the refresh `Register` further in the chain. If not, we can meet an issue with "canceled" `Unregister` reaching registry **after** the refresh `Register`. So we need:
   * executor
   * canceled flag - timer can fire but not yet lock the executor, so we need to prevent timer func from execution
2. Timer needs to be recovered if refresh `Register` fails. If not, we will not unregister NSE from registry. So we need:
   * expirationTime
   * started flag - timer func might not yet actually have been stopped, so we need to check if we don't need to reset the timer

haiodo (approved):

PR #693, opened 2021-02-02
Rework nsmlabels to nsurl.

edwarnicke:


pkg/tools/nsurl/nsurl.go
```diff
@@ -0,0 +1,76 @@
...
+// ToNSURL - convert u to an NSURL
+func ToNSURL(u *url.URL) *NSURL {
+	return &NSURL{
+		URL: u,
+	}
+}
+
+// Mechanism - return Mechanism for the requested Network Service
+func (n *NSURL) Mechanism() *networkservice.Mechanism {
+	mechanism := &networkservice.Mechanism{Cls: cls.LOCAL, Type: n.URL.Scheme}
```
denis-tingaikin, line 51:
```suggestion
	mechanism := &networkservice.Mechanism{Cls: cls.LOCAL, Type: strings.ToUpper(n.URL.Scheme)}
```

denis-tingaikin (approved):
I like this idea

pkg/tools/nsurl/nsurl.go
edwarnicke: Done.

go.mod
```diff
@@ -19,17 +18,16 @@ require (
 	github.com/nats-io/nats-streaming-server v0.17.0
 	github.com/nats-io/stan.go v0.6.0
 	github.com/networkservicemesh/api v0.0.0-20210202152048-ec956057eb3a
+	github.com/networkservicemesh/sdk-vpp v0.0.0-20210202164004-1ec8a9bd6287
```
denis-tingaikin, line 21: This looks like an artifact after rebase.
I guess here can be checkout go.mod, go.sum changes

pkg/tools/nsurl/nsurl.go
```diff
@@ -0,0 +1,76 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+	"github.com/networkservicemesh/api/pkg/api/networkservice/mechanisms/cls"
+	"github.com/networkservicemesh/api/pkg/api/networkservice/mechanisms/common"
+)
+
+// NSURL - wrapper around *url.URL to allow easy extraction of NSM related information
+type NSURL struct {
```
denis-tingaikin, line 38: Also, we can define a type based on another type.
For example:
```suggestion
type NSURL url.UR
```
In this case, we could use a typecast to covert `url.URL` to `NSURL` or vise versa.
That means L43-L47 in this case can be removed.

denis-tingaikin (approved):
@edwarnicke Nice!

PR #694, opened 2021-02-02
Remove pkg/tools/callback as its no longer used.

edwarnicke:


haiodo:
I've moved it into https://github.com/haiodo/callback

PR #698, opened 2021-02-04
Remove localbypass, urls chain elements

Bolodya1997:
# Intro
Existing NSMgr NSE registry chain has too many chain elements used to operate on `Endpoint.Url`:
```go
nseChain := chain_registry.NewNamedNetworkServiceEndpointRegistryServer(
    ...
    urlsRegistryServer,                                                  // <-- stores URLs
    interposeRegistry,                                                   // <-- stores forwarders URLs
    localbypassRegistryServer,                                           // <-- stores URLs except forwarders
    seturl.NewNetworkServiceEndpointRegistryServer(nsmRegistration.Url), // <-- replaces URL with NSMgr URL
    ...
)
```
and additionally we have some networkservice chain elements that actually need information about what endpoints do we have:
```go
rv.Endpoint = endpoint.NewServer(ctx,
    localbypass.NewServer(&localbypassRegistryServer), // <-- replaces NSMgr URL with URL
    ...
    interpose.NewServer(&interposeRegistry),           // <-- iterates over all forwarders
    filtermechanisms.NewServer(&urlsRegistryServer),   // <-- filters local endpoints and forwarders
    ...
)
```
# Issues
1. `urlRegistryServer`, `localbypassRegistryServer` do nothing except of storing URLs.
2. `seturl.NewNetworkServiceEndpointRegistryServer` replaces URL with NSMgr URL but doesn't set it back, so we need `localbypass.NewServer`.
# Solution
1. Remove `urlRegistryServer`, `localbypassRegistryServer` and store forwarders URLs in `interposeRegistry`, endpoints URLs in `seturl`.
2. Make `seturl` set back endpoint URL on `Register`, `Unregister`, `Find` responses.


denis-tingaikin (changes requested):
I agree that some things in the registry chains can be simplified.
But I don't like the idea to pass maps into chain elements. It can produce many unexpected bugs with map mutating.
I would suggest using a more private approach, for example, pass a read-only map over context and so on.

Bolodya1997:
> I agree that some things in the registry chains can be simplified.
But I don't like the idea to pass maps into chain elements. It can produce many unexpected bugs with map mutating.
I would suggest using a more private approach, for example, pass a read-only map over context and so on.

Yep, fixed.

pkg/networkservice/chains/nsmgr/server.go
```diff
@@ -83,8 +77,9 @@ var _ Nsmgr = (*nsmgrServer)(nil)
 func NewServer(ctx context.Context, nsmRegistration *registryapi.NetworkServiceEndpoint, authzServer networkservice.NetworkServiceServer, tokenGenerator token.GeneratorFunc, registryCC grpc.ClientConnInterface, clientDialOptions ...grpc.DialOption) Nsmgr {
 	rv := &nsmgrServer{}
 
-	var urlsRegistryServer registryapi.NetworkServiceEndpointRegistryServer
-	var localbypassRegistryServer registryapi.NetworkServiceEndpointRegistryServer
+	var setURLRegistryServer, interposeRegistryServer registryapi.NetworkServiceEndpointRegistryServer
+
+	filterMechanismsServer := filtermechanisms.NewServer(nsmRegistration.Url, &setURLRegistryServer)
```
denis-tingaikin, line 82: Why do we need to define this here? Can we move this to the chain at line 117?

pkg/registry/common/seturl/server.go
```diff
@@ -19,44 +19,72 @@ package seturl
...
+
 	"github.com/networkservicemesh/api/pkg/api/registry"
+
+	"github.com/networkservicemesh/sdk/pkg/registry/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/stringurl"
 )
 
-type setMgrServer struct {
-	url string
+type setURLNSEServer struct {
```
denis-tingaikin, line 33: Do not think that we should use chain elements that have more than one responsibility.  
I would be suggesting to keep seturl as simple as it was before and move the other logic into localbypass registry serever.

denis-tingaikin (changes requested):

pkg/networkservice/chains/nsmgr/server.go
Bolodya1997: It fills `setURLRegistryServer` which is used in line 101.

pkg/registry/common/seturl/server.go
Bolodya1997: I suppose that it is one responsibility to swap between NSMgr URL for Registry and NSE URL for NSMgr.

pkg/networkservice/chains/nsmgr/server.go
```diff
@@ -83,8 +77,9 @@ var _ Nsmgr = (*nsmgrServer)(nil)
 func NewServer(ctx context.Context, nsmRegistration *registryapi.NetworkServiceEndpoint, authzServer networkservice.NetworkServiceServer, tokenGenerator token.GeneratorFunc, registryCC grpc.ClientConnInterface, clientDialOptions ...grpc.DialOption) Nsmgr {
 	rv := &nsmgrServer{}
 
-	var urlsRegistryServer registryapi.NetworkServiceEndpointRegistryServer
-	var localbypassRegistryServer registryapi.NetworkServiceEndpointRegistryServer
+	var localBypassRegistryServer, interposeRegistryServer registryapi.NetworkServiceEndpointRegistryServer
+
+	filterMechanismsServer := filtermechanisms.NewServer(nsmRegistration.Url, &localBypassRegistryServer)
```
denis-tingaikin, line 82: why this outside of networkservice chain

Bolodya1997: https://github.com/networkservicemesh/sdk/pull/698#discussion_r571854769

denis-tingaikin, line 82: filtermechanisms should not produce localbypass registry server.
Chain elements should be as independent as possible

denis-tingaikin (changes requested):

pkg/networkservice/chains/nsmgr/server.go
Bolodya1997: Yep, fixed.

denis-tingaikin (approved):

PR #699, opened 2021-02-04
Rework memory registry server

Bolodya1997:
# Issues
There are data races, deadlocks and invalid behavior in both NSE, NS memory registry servers.
# Solution
1. Add unit tests for data races, slow receiver, all Register/Unregister received.
2. Rework NSE, NS memory registry servers.

### DIFF
--- a/pkg/registry/common/memory/nse_server_test.go
+++ b/pkg/registry/common/memory/nse_server_test.go
@@ -191,29 +191,31 @@ func TestNetworkServiceEndpointRegistryServer_DataRace(t *testing.T) {
 
 	s := memory.NewNetworkServiceEndpointRegistryServer()
 
-	go func() {
-		for ctx.Err() == nil {
-			nse, err := s.Register(ctx, &registry.NetworkServiceEndpoint{Name: "nse-1"})
-			require.NoError(t, err)
-
-			_, err = s.Unregister(ctx, nse)
-			require.NoError(t, err)
-		}
-	}()
+	_, err := s.Register(ctx, &registry.NetworkServiceEndpoint{Name: "nse"})
+	require.NoError(t, err)
 
 	c := adapters.NetworkServiceEndpointServerToClient(s)
 
-	wg := new(sync.WaitGroup)
+	var wgStart, wgEnd sync.WaitGroup
 	for i := 0; i < 10; i++ {
-		wg.Add(1)
+		wgStart.Add(1)
+		wgEnd.Add(1)
 		go func() {
-			defer wg.Done()
+			defer wgEnd.Done()
 
-			stream, err := c.Find(ctx, &registry.NetworkServiceEndpointQuery{
-				NetworkServiceEndpoint: &registry.NetworkServiceEndpoint{Name: "nse-1"},
+			findCtx, findCancel := context.WithTimeout(ctx, time.Second)
+			defer findCancel()
+
+			stream, err := c.Find(findCtx, &registry.NetworkServiceEndpointQuery{
+				NetworkServiceEndpoint: &registry.NetworkServiceEndpoint{Name: "nse"},
 				Watch:                  true,
 			})
 			assert.NoError(t, err)
+
+			_, err = stream.Recv()
+			assert.NoError(t, err)
+
+			wgStart.Done()
 
 			for j := 0; j < 100; j++ {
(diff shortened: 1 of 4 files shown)


pkg/registry/common/memory/nse_server_test.go
```diff
@@ -177,6 +183,180 @@ func TestNetworkServiceEndpointRegistryServer_RegisterAndFindByLabelWatch(t *tes
 	require.True(t, proto.Equal(expected, <-ch))
 }
 
+func TestNetworkServiceEndpointRegistryServer_DataRace(t *testing.T) {
```
denis-tingaikin, line 186: This test looks invalid to me.
Because this test made a race in the test (line 222)
This can not be reproduced in real work if we'll use a real client.

Please consider reworking this test to catch a true race condition.

pkg/registry/common/memory/nse_server_test.go
```diff
@@ -177,6 +183,180 @@ func TestNetworkServiceEndpointRegistryServer_RegisterAndFindByLabelWatch(t *tes
...
+
+				nse.Name = ""
+			}
+		}()
+	}
+
+	wg.Wait()
+}
+
+func TestNetworkServiceEndpointRegistryServer_SlowReceiver(t *testing.T) {
```
denis-tingaikin, line 230: This test is invalid.

You are trying to write 100 entries to a channel that has a limited 10 by default. Please look https://github.com/networkservicemesh/sdk/blob/master/pkg/registry/common/memory/common.go#L19
And check how the adapted client is actually working.

Please use option setChannelSize(100) or
Rework this test to correctly read/write from the channel to simulate the real client/server.

pkg/registry/common/memory/nse_server_test.go
```diff
@@ -177,6 +183,180 @@ func TestNetworkServiceEndpointRegistryServer_RegisterAndFindByLabelWatch(t *tes
...
+
+	defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
+
+	_, err = stream.Recv()
+	require.NoError(t, err)
+
+	findCancel()
+}
+
+func TestNetworkServiceEndpointRegistryServer_ShouldReceiveAllRegisters(t *testing.T) {
```
denis-tingaikin, line 265: Also works fine on master with increasing buffer size inside the test...

pkg/registry/common/memory/nse_server_test.go
```diff
@@ -177,6 +183,180 @@ func TestNetworkServiceEndpointRegistryServer_RegisterAndFindByLabelWatch(t *tes
...
+			assert.NoError(t, err)
+
+			_, err = stream.Recv()
+			assert.NoError(t, err)
+		}()
+	}
+	wg.Wait()
+}
+
+func TestNetworkServiceEndpointRegistryServer_ShouldReceiveAllUnregisters(t *testing.T) {
```
denis-tingaikin, line 304: Also works fine on master with increasing buffer size inside the test...

denis-tingaikin (changes requested):
Most covered issues related to _tests_ and not related to code.

I guess we could add the unlimited buffers for channels to avoid potential channel deadlocks.

pkg/registry/common/memory/nse_server_test.go
Bolodya1997: It is actually very similar to the `discover` server using watching `Find` and finishing receiving after the first match.

Bolodya1997:
> Most covered issues related to tests and not related to code of memory chain element.
>
> I guess we could add the unlimited buffers for channels to avoid potential channel deadlocks
OR do not try to send an update if the channel is blocked.

I am not sure that increasing buffer size is an option if we can find a solution without it.

pkg/registry/common/memory/nse_server_test.go
Bolodya1997: Yes, setting unlimited buffer size fixes almost all found issues :)

denis-tingaikin:
@Bolodya1997 Do we still have a problem with channel limitation with these changes?

Bolodya1997:
> @Bolodya1997 Do we still have a problem with channel limitation with these changes?

Nope, there is no more need to add unlimited channels with these changes.

denis-tingaikin (approved):

PR #700, opened 2021-02-05
Close client chain on failed grpc connection

artem-belov:
1. Closing client chain on failed grpc connection
2. Fix clients leak on failed first request


pkg/networkservice/common/clienturl/client.go
```diff
@@ -55,36 +56,58 @@ func NewClient(ctx context.Context, clientFactory func(ctx context.Context, cc g
...
-		if u.dialErr != nil {
+		go func() {
+			var dialErr error
+			cc, dialErr = grpc.DialContext(ctx, grpcutils.URLToTarget(clientURL), u.dialOptions...)
+			if dialErr != nil {
+				errCh <- dialErr
+				return
+			}
+			errCh <- nil
+		}()
```
denis-tingaikin, line 91: Why do we need a goroutine here?

pkg/networkservice/common/connect/server_test.go
```diff
@@ -14,18 +14,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package connect_test
```
denis-tingaikin, line 17: Please do not rename the test pkgs

pkg/networkservice/common/connect/server_test.go
```diff
@@ -150,6 +151,9 @@ func TestConnectServer_Request(t *testing.T) {
 
 		require.Equal(t, requestNext.Connection.String(), conn.String())
 
+		require.Equal(t, 1, len(cs.clientsCloseFuncs))
```
denis-tingaikin, line 154: Please do not use a white box testing

pkg/networkservice/common/connect/server_test.go
```diff
@@ -245,6 +256,8 @@ func TestConnectServer_RequestParallel(t *testing.T) {
 				conn, err := s.Request(clienturlctx.WithClientURL(ctx, urlA), request)
 				assert.NoError(t, err)
 
+				runtime.Gosched()
```
denis-tingaikin, line 259: Why do we need this in test?

pkg/tools/sandbox/utils.go
```diff
@@ -95,14 +97,24 @@ func NewEndpoint(ctx context.Context, nse *registry.NetworkServiceEndpoint, gene
...
-		client.NewClientFactory(
-			fmt.Sprintf("nsc-%v", uuid.New().String()),
-			nil,
-			generatorFunc,
-			additionalFunctionality...),
-		append(opentracing.WithTracingDial(), grpc.WithBlock(), grpc.WithInsecure())...)
+	type nsServer struct {
+		networkservice.NetworkServiceServer
+	}
+	rv := &nsServer{}
```
denis-tingaikin, line 103: Could you explain, why do we need this?

denis-tingaikin (changes requested):

pkg/networkservice/common/clienturl/client.go
artem-belov: To exit on timeout. We cannot use context with deadline here

pkg/networkservice/common/connect/server_test.go
artem-belov: Done

pkg/networkservice/common/connect/server_test.go
artem-belov: Done

pkg/tools/sandbox/utils.go
artem-belov: Removed from this PR

pkg/networkservice/common/connect/server_test.go
artem-belov: Removed

pkg/networkservice/common/clienturl/client.go
artem-belov: Fixed

pkg/networkservice/common/connect/server.go
```diff
@@ -95,60 +120,97 @@ func (s *connectServer) Close(ctx context.Context, conn *networkservice.Connecti
...
+	s.clientsMutex.Lock()
+	defer s.clientsMutex.Unlock()
 
-	return *clientPtr, cancel
+	if onClose, ok := s.clientsCloseFuncs[client]; ok {
+		for {
+			if onClose() {
+				return
+			}
+		}
```
denis-tingaikin, line 214: Why we cannot try to call close one time?

pkg/networkservice/common/connect/server.go
```diff
@@ -22,29 +22,34 @@ package connect
...
 
+type closeFunc func() bool
+
 type connectServer struct {
 	ctx               context.Context
 	clientFactory     func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
 	clientDialOptions []grpc.DialOption
 	connInfos         connectionInfoMap
 	clients           clientmap.RefcountMap
+	clientsMutex      sync.Mutex
```
denis-tingaikin, line 51: NOTE: sync map will work here faster because in some cases we'll not block a group of clients.
Also, consider do you need order here. Because in this case you also could look at `serialize.Executor` for sync.

denis-tingaikin (changes requested):

pkg/networkservice/common/connect/server.go
artem-belov: onClose hase to be called while exist at least one ref.

pkg/networkservice/common/connect/server.go
artem-belov: Another connection with the same client can break it. So mutex is required here

pkg/networkservice/common/clienturl/client.go
```diff
@@ -75,16 +75,22 @@ func (u *clientURLClient) init() error {
...
 		if u.dialErr != nil {
 			return
 		}
+
 		u.client = u.clientFactory(u.ctx, cc)
 
 		go func() {
 			defer func() {
 				_ = cc.Close()
+				cancel()
```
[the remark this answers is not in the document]
artem-belov: Done

pkg/networkservice/common/connect/server_test.go
```diff
@@ -216,6 +223,10 @@ func TestConnectServer_RequestParallel(t *testing.T) {
 		err := startServer(ctx, urlA, serverA)
 		require.NoError(t, err)
 
+		// Time for servers to start listening
+		<-time.After(10 * time.Millisecond)
```
denis-tingaikin, line 227: Can we avoid waiting for the tests?

pkg/networkservice/common/connect/server_test.go
```diff
@@ -113,6 +114,10 @@ func TestConnectServer_Request(t *testing.T) {
 		))
 		require.NoError(t, err)
 
+		// Time for servers to start listening
+		<-time.After(10 * time.Millisecond)
```
denis-tingaikin, line 118: Can we avoid waiting in the tests?

pkg/networkservice/common/connect/server.go
denis-tingaikin: What if we have 1e3 clients? Do we really need in this case to call on close 1e3 times?

pkg/networkservice/common/connect/server.go
denis-tingaikin: It looks like over synchronization.

pkg/networkservice/common/connect/server.go
artem-belov: Done

pkg/networkservice/common/connect/server_test.go
artem-belov: Done

pkg/networkservice/common/connect/server_test.go
artem-belov: Done

pkg/networkservice/common/connect/server.go
artem-belov: We have to be sure that other connection does not make write operation on the same client at the same time

pkg/networkservice/common/connect/server.go
```diff
@@ -22,29 +22,34 @@ package connect
...
-	"github.com/networkservicemesh/sdk/pkg/networkservice/core/chain"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/inject/injecterror"
 	"github.com/networkservicemesh/sdk/pkg/tools/clientmap"
 	"github.com/networkservicemesh/sdk/pkg/tools/clienturlctx"
+	"github.com/networkservicemesh/sdk/pkg/tools/grpcutils"
+	"github.com/networkservicemesh/sdk/pkg/tools/logger"
 )
 
+type closeFunc func() bool
```
denis-tingaikin, line 43: Why we can not use `func() bool` inline?

pkg/networkservice/common/connect/server.go
```diff
@@ -22,29 +22,34 @@ package connect
...
+
 type connectServer struct {
 	ctx               context.Context
 	clientFactory     func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient
 	clientDialOptions []grpc.DialOption
 	connInfos         connectionInfoMap
 	clients           clientmap.RefcountMap
+	clientsMutex      sync.Mutex
+	clientsCloseFuncs map[networkservice.NetworkServiceClient]closeFunc
```
denis-tingaikin, line 52: Why we can not put closeFunc into clientInfo?

pkg/networkservice/common/connect/server_test.go
```diff
@@ -62,6 +63,17 @@ func startServer(ctx context.Context, listenOn *url.URL, server networkservice.N
 	}
 }
 
+func checkServer(target *url.URL) error {
```
denis-tingaikin, line 66: Probably better to register health service on the tested server and test that server started to serve to verify the sever.

pkg/tools/clientmap/refcount.go
```diff
@@ -130,3 +130,10 @@ func (m *RefcountMap) Delete(key string) (deleted bool) {
 	_, _, deleted = m.LoadAndDelete(key)
 	return deleted
 }
+
+// Drop is deleting value for a key for any count value
+// count == 1  -->  delete
+// count > 1   -->  delete
+func (m *RefcountMap) Drop(key string) {
```
denis-tingaikin, line 137: Is this map using by someone else than `connect`?

denis-tingaikin: I would also suggest creating a private map based on this one for the connect goals to prevent API leaks and also to optimize synchronization on client creating/deleting

pkg/networkservice/common/connect/server.go
```diff
@@ -26,29 +26,40 @@ import (
...
+
+	connInfos connectionInfoMap
+	clients   clientInfoMap
+	executor  multiexecutor.MultiExecutor
+}
+
+type clientInfo struct {
+	client  networkservice.NetworkServiceClient
+	count   int
+	onClose closeFunc
```
Bolodya1997, line 57: Looks like `onClose` is a `context.CancelFunc` wrapped with `closeFunc`. Do we need this additional type?

pkg/networkservice/common/connect/server.go
```diff
@@ -65,22 +76,41 @@ func NewServer(
...
 
 func (s *connectServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
-	conn, err := s.client(ctx, request.GetConnection()).Request(ctx, request.Clone())
+	clientURL := clienturlctx.ClientURL(ctx)
+
+	clientInfo := s.client(ctx, request.GetConnection())
+	conn, err := clientInfo.client.Request(ctx, request.Clone())
 	if err != nil {
+		if _, ok := s.connInfos.Load(request.GetConnection().GetId()); !ok {
+			s.closeClient(clientInfo, clientURL.String())
```
Bolodya1997, line 85: What if there is no `clientURL` in the context?
We will come here with `injecterror.NewClient(err)` and then panic on `clientURL.String()` instead of returning an error.
Probably check from line 132 should be right after line 79.

pkg/networkservice/common/connect/server.go
Bolodya1997: And so it can be `cancel`, not `onClose`.

pkg/networkservice/common/connect/server_test.go
```diff
@@ -91,27 +128,31 @@ func TestConnectServer_Request(t *testing.T) {
...
 				Cls:  cls.LOCAL,
 				Type: memif.MECHANISM,
 			}),
-		))
+		)
+		err = startServer(ctx, urlB, nsServerB)
 		require.NoError(t, err)
+		require.NoError(t, checkServer(urlB, nsServerB))
+
+		ignoreCurrentGoroutines := goleak.IgnoreCurrent()
```
Bolodya1997, line 155: Can be replace with `defer goleak.VerifyNone(t, goleak.IgnoreCurrent())` since it is a function call.

pkg/networkservice/common/connect/server_test.go
```diff
@@ -179,6 +220,8 @@ func TestConnectServer_Request(t *testing.T) {
 		require.Nil(t, serverClient.capturedRequest)
 		require.Nil(t, serverB.capturedRequest)
 		require.Nil(t, serverNext.capturedRequest)
+
+		goleak.VerifyNone(t, ignoreCurrentGoroutines)
```
Bolodya1997, line 224: With removing this line.

pkg/networkservice/common/connect/server_test.go
```diff
@@ -264,6 +310,66 @@ func TestConnectServer_RequestParallel(t *testing.T) {
...
+			serverA,
+			newEditServer("a", "A", &networkservice.Mechanism{
+				Cls:  cls.LOCAL,
+				Type: kernel.MECHANISM,
+			}),
+		))
+		require.NoError(t, err)
+
+		// Time for servers to start listening
+		<-time.After(10 * time.Millisecond)
```
Bolodya1997, line 349: Shouldn't this be replaced with `checkServer`?

pkg/tools/multiexecutor/utils.go
```diff
@@ -14,15 +14,22 @@
...
 
 import (
 	"sync"
 
 	"github.com/edwarnicke/serialize"
 )
 
-type multiExecutor struct {
+// Executor is a serialize.Executor interface type
+type Executor interface {
```
Bolodya1997, line 27: I am not sure that this should be removed here from `serialize`, because almost no one who takes `Executor` from context knows anything about `MultiExecutor` but something about `serialize`.

artem-belov: Removed serialize from comment. All "context" methods already left at serialize

artem-belov: This interface required by MultiExecutor to pass all AsyncExec calls through it to be sure that executor will be deleted in time. 
All context methods are left at serialize

pkg/networkservice/common/connect/server.go
artem-belov: Done

pkg/networkservice/common/connect/server.go
artem-belov: Done

pkg/networkservice/common/connect/server_test.go
artem-belov: Done

pkg/networkservice/common/connect/server_test.go
artem-belov: Done

pkg/networkservice/common/connect/server_test.go
artem-belov: Done

pkg/networkservice/common/connect/server.go
```diff
@@ -65,22 +74,45 @@ func NewServer(
 }
 
 func (s *connectServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
-	conn, err := s.client(ctx, request.GetConnection()).Request(ctx, request.Clone())
+	clientURL := clienturlctx.ClientURL(ctx)
+	if clientURL == nil {
+		err := errors.Errorf("clientURL not found for incoming connection: %+v", request.GetConnection())
+		return injecterror.NewServer(err).Request(ctx, request)
```
Bolodya1997, line 80:
```go
return nil, errors.Errorf("clientURL not found for incoming connection: %+v", request.GetConnection())
```

pkg/tools/multiexecutor/utils.go
Bolodya1997: `MultiExecutor` is using `Executor` as a return value mostly to make code cleaner than it actually needs it. So it is not an any strong consumer, it is only a producer.

artem-belov: Yes, but it will create possible sittuation, when two Executors are exists for the same ID

Bolodya1997 (approved):

pkg/tools/multiexecutor/executor_func.go
```diff
@@ -0,0 +1,28 @@
...
+// ExecutorFunc - function type, simulates serialize.Executor behavior
+//		Can be used to guarantee exclusive, in order execution of functions.
+type ExecutorFunc func(f func()) <-chan struct{}
+
+// AsyncExec - simulates serialize.Executor AsyncExec call
+// 		Guarantees f() will be executed Exclusively and in the Order submitted.
+//		It immediately returns a channel that will be closed when f() has completed execution.
+func (e ExecutorFunc) AsyncExec(f func()) <-chan struct{} {
+	return e(f)
+}
```
denis-tingaikin, line 28: do we need this as a public?

pkg/tools/multiexecutor/utils.go
```diff
@@ -59,8 +63,9 @@ func (e *multiExecutor) AsyncExec(id string, f func()) (ch <-chan struct{}) {
 	return ch
 }
 
-func (e *multiExecutor) Executor(id string) Executor {
-	return executorFunc(func(f func()) <-chan struct{} {
+// Executor - returns Executor by ID
+func (e *MultiExecutor) Executor(id string) ExecutorFunc {
```
denis-tingaikin, line 67:
```suggestion
func (e *MultiExecutor) Executor(id string) Executor {
```

Bolodya1997: `serialize` already depends on `multiexecutor`. We cannot use direct `multiexecutor` -> `serialize` dependency. So the only other option is to add something like `serializectx`, but it really doesn't make sense if we can simply make `ExecutorFunc` public and return it.

artem-belov: There should be ExecutorFunc

artem-belov: Executor interface moved to other package

pkg/tools/multiexecutor/executor_func.go
artem-belov: Yes. At the moment it become a public tool ExecutorFunc has to be also public

pkg/tools/multiexecutor/utils.go
denis-tingaikin: > Executor interface moved to other package

Dont see this. Could you point a file?

denis-tingaikin: > serialize already depends on multiexecutor. We cannot use direct multiexecutor -> serialize dependency. So the only other option is to add something like serializectx, but it really doesn't make sense if we can simply make ExecutorFunc public and return it.

I did not quite follow this

denis-tingaikin: I would suggest to just move all common staff with context and itnerfaces from serialize to tools/serialize.

pkg/tools/multiexecutor/executor_func.go
denis-tingaikin: See this https://github.com/networkservicemesh/sdk/pull/700#discussion_r573101131

pkg/tools/multiexecutor/utils.go
```diff
@@ -14,15 +14,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package serialize
+// Package multiexecutor provides a structure MultiExecutor that can be used to guarantee exclusive by ID, in order execution of functions.
+package multiexecutor
```
edwarnicke, line 18: Would it make more sense to simply stick an executor into metadata?

pkg/tools/multiexecutor/utils.go
edwarnicke: > serialize already depends on multiexecutor.

I also don't follow this... I would further suggest we move to using https://github.com/edwarnicke/serialize rather than further clogging up tools (we need less in tools/ not more).

edwarnicke:
The complexity of this fix feels very very off.

Wouldn't it simply make more sense to have a closeonerror chain element you could stick into both client and server chains to close anything that has gone wrong along the way?

Increasing the complexity of connect, introducing complex new tools/ constructs etc feels like over kill here.

denis-tingaikin:
@edwarnicke Here we have one issue that we are using a refcount map and in case if the connection has died we should close the client ignoring refs. Are you have in mind any idea how we can close all ref clients from the separate chain element?

denis-tingaikin:
Changes moved to another PR: https://github.com/networkservicemesh/sdk/pull/700

pkg/tools/multiexecutor/executor_func.go
denis-tingaikin: Fixed by https://github.com/networkservicemesh/sdk/pull/703/commits/8635adba8e8e0d05eea0f4d889567564f62dd036

PR #703, opened 2021-02-10
Pr fix connect server

### LINKED ISSUE #704: Connect server doesn't handle GRPC connection error
# Intro
If GRPC connection failes, `connect` server does nothing to make a new GRPC connection, so all subsequent requests will fail with GRPC error until `discover` find a new URL.
# Issue
If remote server restarts, healing can't recover connection without selecting a new server.
# Solution
#703

If `connect` server receives a GRPC error from client request, it deletes it. Subsequent request to this URL:
* with the same connection ID - will create a new client and continue working after 1 error received.
* with different connection ID - will receive a GRPC error, because `connInfo` still refers to the old client but will replace old client with a new one on further requests.

Bolodya1997:
Continues #700.
Closes #704.


denis-tingaikin (approved):

pkg/networkservice/common/connect/server_test.go
```diff
@@ -267,6 +311,61 @@ func TestConnectServer_RequestParallel(t *testing.T) {
...
+	// 1. Create connectServer
+
+	s := connect.NewServer(context.TODO(),
+		func(_ context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
+			return injecterror.NewClient()
+		},
+		grpc.WithInsecure(),
+	)
+
+	func() {
```
denis-tingaikin, line 326: This function looks strange.

@Bolodya1997 Could we also remove this function and call `goleak.VerifyNone` in line 361?

pkg/networkservice/common/connect/server_test.go
```diff
@@ -267,6 +311,61 @@ func TestConnectServer_RequestParallel(t *testing.T) {
...
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		// 2. Setup servers
+
+		urlA := &url.URL{Scheme: "tcp", Host: "127.0.0.1:10000"}
```
denis-tingaikin, line 332: Let's do not hardcode port

Bolodya1997: Fixed.

pkg/networkservice/common/connect/server_test.go
Bolodya1997: Yep, fixed.

denis-tingaikin (approved):

denis-tingaikin:
@edwarnicke Could you take a look at this patch that's solve issue https://github.com/networkservicemesh/sdk/issues/704?

Any thoughts are welcome :)

denis-tingaikin:
@edwarnicke , @Bolodya1997 I think this also can be simplified if we will use refcountmap that allows to delete entry not look at ref count. What do you think?

artem-belov (approved):

pkg/networkservice/common/clienturl/client.go
```diff
@@ -75,11 +75,16 @@ func (u *clientURLClient) init() error {
 			u.dialErr = errors.New("cannot dial nil clienturl.ClientURL(ctx)")
 			return
 		}
+
+		ctx, cancel := context.WithTimeout(u.ctx, dialTimeout)
```
edwarnicke, line 79: Why are we introducing a new arbitrary dialTimeout instead of deriving it from the ctx of the request?

pkg/networkservice/common/connect/server.go
```diff
@@ -94,60 +125,67 @@ func (s *connectServer) Close(ctx context.Context, conn *networkservice.Connecti
...
+	<-s.executor.AsyncExec(clientURL.String(), func() {
+		// Fast path if we already have client for the clientURL and we chould not reconnect, use it.
+		var loaded bool
+		client, loaded = s.clients.Load(clientURL.String())
+		if !loaded {
+			// If not, create and LoadOrStore a new one.
+			client = s.newClient(clientURL)
+			s.clients.Store(clientURL.String(), client)
+		}
+		client.count++
```
edwarnicke, line 156: Why are we blocking globally when we could do something fairly straightforward and non-blocking like:

```suggestion
		for {
			// Create a client
			client = s.newClient(clientURL)
			atomic.AddInt64(&client.count,1)
			var loaded bool
			// Load or store the client
			client, loaded = s.clients.LoadOrStore(clientURL.String(),client)
			// If its loaded and incrementing the counter from 0... somebodies already deleted it, so try again
			if loaded && atomic.AddInt64(&client.count,1) > 1 {
				continue
			}
		}
```

pkg/networkservice/common/connect/server.go
```diff
@@ -94,60 +125,67 @@ func (s *connectServer) Close(ctx context.Context, conn *networkservice.Connecti
...
+func (s *connectServer) closeClient(client *clientInfo, clientURL string) {
+	<-s.executor.AsyncExec(clientURL, func() {
+		client.count--
+		if client.count == 0 {
+			if loadedClient, ok := s.clients.Load(clientURL); ok && client == loadedClient {
+				s.clients.Delete(clientURL)
+			}
+			client.onClose()
+		}
+	})
```
edwarnicke, line 181: Could be made non-blocking with something like:

```suggestion
	if atomic.AddInt64(&client.count,-1) == 0 {
		client,ok := s.clients.LoadAndDelete(clientURL)
		if ok {
			client.onClose()
		}
	}
```

pkg/networkservice/common/connect/server.go
```diff
@@ -65,22 +74,44 @@ func NewServer(
...
 	if err != nil {
+		if _, ok := s.connInfos.Load(request.GetConnection().GetId()); !ok {
+			s.closeClient(clientInfo, clientURL.String())
+		}
+
+		// Close current client chain if grpc connection was closed
+		if grpcutils.UnwrapCode(err) == codes.Canceled {
+			s.deleteClient(clientInfo, clientURL.String())
+			s.connInfos.Delete(request.GetConnection().GetId())
+		}
```
edwarnicke, line 93: Just because we get a codes.Canceled ... doesn't mean anything about the *client* ... it only tells you about the *connection*

pkg/networkservice/common/serialize/client.go
```diff
@@ -25,10 +25,11 @@ import (
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/multiexecutor"
 )
 
 type serializeClient struct {
-	executor multiExecutor
+	executor multiexecutor.MultiExecutor
```
edwarnicke, line 32: This feels like a separate PR to introduce multiexecutor into serialize.  Could we split it out?

edwarnicke:
One way or the other, we need to find a way to *not* be queueing up all Requests/Closes globally in connect.  Its fine to queue them up by Connection.ID... but not globally.  I think we can get there with a combination of sync.Map and atomics.

pkg/networkservice/common/connect/server.go
artem-belov: @edwarnicke That is the point, we want to close client if something wrong with connection. Right now if grpc connection is broken, we just fail all requests comming to this client and doing nothing to reconnect.

edwarnicke: Doesn't codes.Canceled mean something is wrong with the *request* not the *connection*?  I presume it comes back if the ctx for the request exceeds its timeout for example...

pkg/networkservice/common/connect/server.go
edwarnicke: Ah... I'd missed that this was the multiexecutor and thus per clientURL and not global.  This is fine as written.

pkg/networkservice/common/connect/server.go
edwarnicke: Ah... I'd missed that this was the multiexecutor and thus per clientURL and not global. This is fine as written.

pkg/networkservice/common/connect/server.go
edwarnicke: I still suspect this will unnecessarily cycle connections when we don't need to... However, the blast radius should be fine.  Let's move on :)

pkg/networkservice/common/clienturl/client.go
Bolodya1997: It will break discover + roundrobin, because it will wait for the first dial until requests lives and so there won't be any chance to try someone else.

PR #706, opened 2021-02-12
Fix refresh setting invalid expiration time

Bolodya1997:
# Intro
`refresh` uses for `nse.ExpirationTime` first timeout received from the subsequent chain. In case of remote memory registry we have such chain:
```
-- endpoint --
refresh(default_timeout)
...
-- NSMgr --
expire(timeout_1)
...
-- registry --
expire(timeout_2)
...
memory
```
Most probably `timeout_1 << timeout_2 << default_timeout`, because endpoint is on the same node with the NSMgr, but registry is somewhere else.
In such case we got following `nse.ExpirationTime` over time:
```
-- register --
refresh   -> now + default_timeout
expire(1) -> ...
expire(2) -> ...
expire(2) <- now + timeout_2 [starts timer on this time]
expire(1) <- now + timeout_1 [starts timer on this time]
refresh   <- now + timeout_1 [starts timer on this time]
-- refresh register --
refresh   -> now + timeout_1
expire(1) -> ...
expire(2) -> ...
expire(2) <- ... [starts timer on this time]
expire(1) <- ... [starts timer on this time]
refresh   <- ... [starts timer on this time]
```
This is happens because `expire` takes least expiration time from `{ incoming, self, returned }`.
# Issue
`timeout_1` is too small for `refresh -> expire(2)` refresh invocation, so endpoint will be unregistered.
# Solution
`refresh` should use either default expiry duration, or given from the initial Register call.

### DIFF
--- a/pkg/registry/common/refresh/nse_registry_client.go
+++ b/pkg/registry/common/refresh/nse_registry_client.go
@@ -55,18 +55,18 @@ func (c *refreshNSEClient) startRefresh(
 	ctx context.Context,
 	client registry.NetworkServiceEndpointRegistryClient,
 	nse *registry.NetworkServiceEndpoint,
+	expiryDuration time.Duration,
 ) {
 	logger := log.FromContext(ctx).WithField("refreshNSEClient", "startRefresh")
 
 	t := time.Unix(nse.ExpirationTime.Seconds, int64(nse.ExpirationTime.Nanos))
-	delta := time.Until(t)
 	go func() {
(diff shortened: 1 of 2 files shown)


denis-tingaikin (changes requested):
This will work incorrectly in case if remote  registry responses  slow.
I would suggest to calculate `delta` or `expiryDuration` time every time after call register.

denis-tingaikin (approved):

PR #707, opened 2021-02-15
feat: Add mechanisms client

denis-tingaikin:

## Motivation

1. This PR adds mechanisms client that can be used in nsc applications to simplify chains.

For example:

This https://github.com/networkservicemesh/cmd-nsc/blob/master/main.go#L125-L185 can be replaced to

```go
client.NewClient(..., 
    mechanisms.NewClient(map[string]networkservice.NetworkServiceClient {
        vfio.MECHANISM: chain.NewNetworkServiceClient(...),
        kernel.MECHANISM: chain.NewNetworkServiceClient(...),
})

```

2. It Fixes a few issues like adding the same error in error: https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/mechanisms/server.go#L74
3. Removes recurrent dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/google/go-cmp v0.5.2
 	github.com/google/uuid v1.1.2
 	github.com/grpc-ecosystem/grpc-opentracing v0.0.0-20180507213350-8e809c8a8645
-	github.com/hashicorp/go-multierror v1.0.0
 	github.com/nats-io/nats-streaming-server v0.17.0
 	github.com/nats-io/stan.go v0.6.0
 	github.com/networkservicemesh/api v0.0.0-20210202152048-ec956057eb3a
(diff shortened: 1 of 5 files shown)


edwarnicke:
The problem is that you need *each* mechanism client to run on the way down the chain to add its information to Mechanism Preferences...

denis-tingaikin:
My thoughts about this:

1.  The user passing `nslabel` so it means we can determine what mechanism we actually needed and we do not need to add other mechanisms on the way down.
2. Mechanism specific elements could have additional chain elements for example sriov `token`: https://github.com/networkservicemesh/cmd-nsc/blob/master/main.go#L168

@edwarnicke What do you think?

edwarnicke:
What do labels in the generic have to do with mechanisms?

denis-tingaikin:
@edwarnicke I meant `nsurl` https://github.com/networkservicemesh/sdk/blob/master/pkg/tools/nsurl/nsurl.go

edwarnicke:
Ah... now I see your point, and you are correct.

In a pure NSC, this is the right behavior.

In a Forwarder, or other NSE that passes through, its wrong... but it can be used *optionally* in a pure NSC.

PR #711, opened 2021-02-17
Fix memory tests

Bolodya1997:
# Issue
Sometimes `TestNetworkServiceEndpointRegistryServer_ShouldReceiveAllRegisters`, `TestNetworkServiceEndpointRegistryServer_ShouldReceiveAllUnregisters` fail due to the huge amount of NSEs, low timeout.
# Solution
Reduce amount, increase timeout.

### DIFF
--- a/pkg/registry/common/memory/common_test.go
+++ b/pkg/registry/common/memory/common_test.go
@@ -1,0 +1,40 @@
+// Copyright (c) 2021 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


denis-tingaikin (changes requested):
These tests should be refactored.
Please pass one context with deadline to each goroutine

And do select on wg.Done() or ctx.Done()

Note: unit test should run in 1s maximum

haiodo (approved):

Bolodya1997:
> Please pass one context with deadline to each goroutine

Done.

> And do select on wg.Done() or ctx.Done()

`sync.WaitGroup` is not a channel, we can't `select` on it. Context timeout would break `stream.Recv()` in all goroutines and so they would update `wg` - there is no need to select on `ctx.Done()`.

> Note: unit test should run in 1s maximum

Reduced to 1 second.

denis-tingaikin:
> sync.WaitGroup is not a channel, we can't select on it. Context timeout would break stream.Recv() in all goroutines and so they would update wg - there is no need to select on ctx.Done()

you could create a goroutine to wait for wg.Done.

also, can we reduce channel size and reduce find request count?

haiodo (approved):

denis-tingaikin (approved):

Issue #269, opened 2020-06-02
Add OPA policies for NSM registries


denis-tingajkin:
## Motivation

Probably we need to add authorization policies for NSM registries to cover scenarios such as 

1. NSE/NSMgr/xconnNSE should be able to unregister only itself. NSE/NSMgr/xconn NSE should not be able to unregister another NSE/NSMgr/xconn NSE.
2. Only NSMgr/vl3 NSE should be able to search NSE in the registry.
3. Only NSMgr and other registries could use `Bulk` API.

## Problem of implementation

We need to somehow pass Path data to registries.

## Potential solution

We can use pkg `grpc/metadata` for passing tokens for registries.
```
import "google.golang.org/grpc/metadata"
...  
ctx = metadata.NewContext( 
    ctx, 
    metadata.Pairs("path-data", "data"), 
) 
```

PR #716, opened 2021-02-18
feat: Convert updatetoken client to PerRPCCredentials to solve issue 702

### LINKED ISSUE #702: updatetoken client generates tokens using server authinfo in connect chain
## Issue

The issue is related to two problems:

1. When server `connect` uses `updatetoken` client then the client using the `connect` server's` authinfo.

Look at
https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/updatetoken/client.go#L45
https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/chains/client/client.go#L71
https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/utils/inject/injectpeer/client.go#L83

As you can see `injectpeer ` sets a peer on backward when to generate the token it should be done before calling next.

2. NSC can not set [aud](https://tools.ietf.org/html/rfc7519#section-4.1.3) in its JWT because at first NSC should resolve the service name.

## Solutions

Currently, we have two solutions for this issue:

1. Create `authorize` ns client chain element that will do handshake via
https://github.com/grpc/grpc-go/blob/master/credentials/credentials.go#L144 in this case updatetoken will correctly generate tokens.

2. Somehow match endpoint in the registry to spiffieID and use it in `aud`.

denis-tingaikin:

## Description
This PR is the first part of enabling OPA by default.

To enable opa we need to do the next things:

1. Fix #702
2. Refactor chains client to use options pattern
3. Enable opa by default and disable it for sandbox tests.

This PR does the first step.

## Motivation

This PR:

1. Closes #702
2. Removes `updatetoken` client because it cannot work correctly with grpc clients flow (look at issue 702).
3. Allows passing client's token via grpc.metadata with enabled security transport. 
4. Adds to `updatepath` server possible to put the token from the grpc.metadata into the previous path segment.
5. Adds token.PerRPCCredentials that should be used instead of  `updatetoken` client. 
6. Potentially unblocks https://github.com/networkservicemesh/sdk/issues/269


denis-tingaikin:
Blocked by:

https://github.com/networkservicemesh/api/pull/80
https://github.com/edwarnicke/grpcfd/issues/15

pkg/networkservice/common/clienturl/client.go
```diff
@@ -58,14 +58,14 @@ func (u *clientURLClient) Request(ctx context.Context, request *networkservice.N
 	if err := u.init(); err != nil {
 		return nil, err
 	}
-	return u.client.Request(ctx, request)
+	return u.client.Request(ctx, request, opts...)
```
edwarnicke, line 61: Good catch!

pkg/networkservice/common/heal/client_test.go
```diff
@@ -136,15 +136,13 @@ func TestHealClient_EmptyInit(t *testing.T) {
 		},
 	}
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
+	ctx, cancelFunc := context.WithCancel(token.NewContext(context.Background(), "token", time.Now().Add(time.Hour)))
```
edwarnicke, line 139: Does heal really need to be token aware?

denis-tingaikin: This unit test is using updatepath server and currently, to test it locally we should use token in the context

denis-tingaikin: Fixed.

fkautz (approved):

PR #718, opened 2021-02-19
refactor: Use Option pattern in chains/client

denis-tingaikin:
Depends on https://github.com/networkservicemesh/sdk/pull/716

### DIFF
--- a/pkg/networkservice/chains/client/client.go
+++ b/pkg/networkservice/chains/client/client.go
@@ -121,7 +121,6 @@ func NewCrossConnectClientFactory(clientOpts ...Option) Factory {
 func NewClientFactory(clientOpts ...Option) Factory {
 	return func(ctx context.Context, cc grpc.ClientConnInterface) networkservice.NetworkServiceClient {
 		return chain.NewNetworkServiceClient(
-			mechanismtranslation.NewClient(),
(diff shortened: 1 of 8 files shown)


fkautz (approved):

pkg/networkservice/chains/client/client.go
```diff
@@ -22,66 +22,106 @@ import (
...
+func WithAdditionalFunctionality(additionalFunctionality ...networkservice.NetworkServiceClient) Option {
+	return Option(func(c *clientOptions) {
+		c.additionalFunctionality = additionalFunctionality
+	})
+}
+
+// WithAuthorizeClient sets authorizeClient for the client chain.
+func WithAuthorizeClient(authorizeClient networkservice.NetworkServiceClient) Option {
+	if authorizeClient == nil {
+		panic("authorizeClient cannot be nil")
```
edwarnicke, line 72: Could we provide a bit more detail in the panic message about how to fix the problem (use WithAuthorizeClient etc?)

denis-tingaikin: @edwarnicke 

This message will be show when will be called:
```go
client.NewClient(ctx, conn, client.WithAuthorizeClient(nil))
```

PR #722, opened 2021-02-21
fix: kernel client should not add new mechanism if kernel mechanism already added into preference list

### LINKED ISSUE #721: mechanisms/clients should not duplicate preference mechanisms
## Issue

If any of our client {vfio, memif, [kernel](https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/common/mechanisms/kernel/client.go#L38)} that appends mechanisms  will be used with some prepared request that already contains mechanisms in the preference list then mechanisms count of the mechanism type will be increased on 1

Also, this can be important for refresh scenarios.

## Solution

If the request already has the needed mechanism then just modify it and do not add a new mechanism.

denis-tingaikin:

This PR fixes kernel client for #721

### DIFF
--- a/pkg/networkservice/common/mechanisms/kernel/client.go
+++ b/pkg/networkservice/common/mechanisms/kernel/client.go
@@ -1,4 +1,6 @@
-// Copyright (c) 2020 Cisco and/or its affiliates.
+// Copyright (c) 2020-2021 Cisco and/or its affiliates.
+//
+// Copyright (c) 2021 Doc.ai and/or its affiliates.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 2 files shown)


fkautz (approved):

pkg/networkservice/common/mechanisms/kernel/client.go
```diff
@@ -44,23 +46,40 @@ func NewClient(options ...Option) networkservice.NetworkServiceClient {
...
+// updateMechanismPreferences returns true if MechanismPreferences has updated
+func updateMechanismPreferences(request *networkservice.NetworkServiceRequest) bool {
+	var updated = false
+
+	for _, m := range request.GetRequestMechanismPreferences() {
+		if m.Type == kernel.MECHANISM {
+			if m.Parameters == nil {
+				m.Parameters = make(map[string]string)
+			}
+			m.Parameters[kernel.NetNSURL] = (&url.URL{Scheme: "file", Path: netNSFilename}).String()
```
haiodo, line 75: Should we add kernel.InterfaceNameKey with k.interfaceName?

denis-tingaikin: yes, we should, fixed.

haiodo (approved):

PR #732, opened 2021-02-24
fix: authorize client should check tokens on backward

denis-tingaikin:

## Issue
authorize client should check tokens on backward because on forward it have not server's cert and also it will cover much more cases.

### DIFF
--- a/pkg/networkservice/chains/client/client.go
+++ b/pkg/networkservice/chains/client/client.go
@@ -93,12 +93,12 @@ func NewClient(ctx context.Context, cc grpc.ClientConnInterface, clientOpts ...O
 		append(
 			append([]networkservice.NetworkServiceClient{
 				updatepath.NewClient(opts.name),
-				opts.authorizeClient,
 				serialize.NewClient(),
 				heal.NewClient(ctx, networkservice.NewMonitorConnectionClient(cc), opts.onHeal),
 				refresh.NewClient(ctx),
 				metadata.NewClient(),
 			}, opts.additionalFunctionality...),
+			opts.authorizeClient,
 			networkservice.NewNetworkServiceClient(cc),
 		)...)
 	return rv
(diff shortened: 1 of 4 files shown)


edwarnicke:
Should we also then put the authorize *last* in the client chain?  And if we reject, we should *definitely* make sure to Close down the chain.

denis-tingaikin:
> And if we reject, we should definitely make sure to Close down the chain.

Sure it is already done, have a look at:  pkg/networkservice/common/authorize/client.go:62

 > Should we also then put the authorize last in the client chain? 
 
 I'm not sure about this because in general, we need to verify tokens that returned from the remote side. 
 Do you see any scenario where it is required?

edwarnicke:
> > Should we also then put the authorize last in the client chain?
> 
> I'm not sure about this because in general, we need to verify tokens that returned from the remote side.
> Do you see any scenario where it is required?

That's actually my point.  Since we are checking policy on *return*, we want to accept/reject *before* doing any other work in the client on return.  So we want to do:

```golang
chain.NewNetworkServiceClient(
		append(
			append([]networkservice.NetworkServiceClient{
				updatepath.NewClient(opts.name),
				serialize.NewClient(),
				heal.NewClient(ctx, networkservice.NewMonitorConnectionClient(cc), opts.onHeal),
				refresh.NewClient(ctx),
				metadata.NewClient(),
			}, opts.additionalFunctionality...),
                          opts.authorizeClient,
			networkservice.NewNetworkServiceClient(cc),
		)...)
```

rather than:
https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/chains/client/client.go#L92-L103

So that we check the authorization as early as possible in the return part of the chain.

edwarnicke:
> > And if we reject, we should definitely make sure to Close down the chain.
> 
> Sure it is already done, have a look at: pkg/networkservice/common/authorize/client.go:62

Got it!

denis-tingaikin:
> So that we check the authorization as early as possible in the return part of the chain.

It makes sense, will fix

denis-tingaikin:
@edwarnicke Done!

PR #747, opened 2021-03-05
Add timer.AfterFunc

Bolodya1997:
# Issue
There is an issue with `time.AfterFunc` and `time.Timer` returned from there:
* if `Timer.Stop()` returns `false`, there is no way to get information about function status - we only knows that it has been started in some time before.
# Solution
Close `Timer.C` on successful `Timer.Stop` or function return. Now we can use such code to make sure that no timer logic is more executing:
```go
if !timer.Stop() {
    <-timer.C
}
// Here we know that function hasn't been started or already has been finished.
```

### DIFF
--- a/pkg/tools/timer/timer.go
+++ b/pkg/tools/timer/timer.go
@@ -1,0 +1,96 @@
+// Copyright (c) 2021 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


haiodo (approved):

pkg/tools/timer/timer.go
```diff
@@ -0,0 +1,96 @@
...
+	t := &Timer{c: make(chan time.Time, 1)}
+	t.C = t.c
+
+	t.timer = time.AfterFunc(d, func() {
+		f()
+		close(t.c)
+	})
+
+	return t
+}
```
denis-tingaikin, line 96:
```suggestion
func AfterFunc(t time.Duration, f func()) *time.Timer {
  done := make(chan time.Time)
  result := time.AfterFunc(t, func() {
	  f()
	  close(done)
  })
  timerr.C = done
  return result
}
```

denis-tingaikin (commented):
This looks overdesigned to me.

Please look at the suggestion.

Also, we can revert this and use the suggestion directly where it is needed.

Issue #724, opened 2021-02-22
nsmgr should call close in case if on backward from nse happens an error


denis-tingaikin:
## Steps to reproduce

1. create nse that opens some resource on request by connId or use ipam server.
2. add to forwarder logic that on backward something happens and  return an error
3. deploy nsmgr, frorwarder, nsc, nse

Actual: nse has leaked resource for a timeout.
Expected: nsmgr could call close for nse

artem-belov:
I suggest 2 possible solutions here:
1. Review all chain elements for posible issues: when Request passed successfully but then it got error on the chain server/client by other reasons. But it is not guarantee  all leakes be solved
2. Add `Close` on discovery server if Request was failed. Also it require removing protection of closing unexisting connection (for example in `timeout` server). But it will require Close through whole path on every failed Request

denis-tingaikin:
@edwarnicke ^^^

edwarnicke:
@artem-belov @denis-tingaikin lets think about this a bit broader than just the NSMgr.

What is the proper behavior of any Client when it receives an error?

The Client's chain may have done work that needs to be cleaned up, and so needs the 'Close' down the chain to behave properly.  But we probably also would rather *not* have every chain element 'Close' down the chain on return of error from something further down the chain.  Example of what we don't want would be:

edwarnicke:
Thinking about this a bit more... basically, we need it to be the case that:

a)  If an error is returned to a chain element from ```next.Client(ctx).Request(...)```, that chain element should Close downstream of itself
b) If a chain element receives a Close call for a connection id it is not active for, it should simply return ```&empty.Empty,nil``` without calling the Close down the chain

If we think that's the right behavior, then I would also say we are unlikely to get chain element authors to reliably get it right, and further, its going to be a burden on them to do so.

Would it then make sense to have a 'Closer' wrapper, similar to the wrapper we have for tracing to implement that logic for chain elements?  Could we then inject it via 'Next' ?

denis-tingaikin:
> If we think that's the right behavior, would it then make sense to have a 'Closer' wrapper, similar to the wrapper we have for tracing to implement that logic for chain elements? Could we then inject it via 'Next' ?

I also thought about this and it looks perfect to me :)

edwarnicke:
@denis-tingaikin Actually... I poked at doing this and no longer think its the right behavior.  Here's why:

[...]

PR #765, opened 2021-03-11
Fix ipam next error

Bolodya1997:
# Issue
`ipam` chain element doesn't free newly selected IP addresses if next server returns an error.

# Solution
If IP addresses have been selected in this `Request` and next server returns an error we should free them.

### DIFF
--- a/pkg/networkservice/ipam/point2pointipam/server.go
+++ b/pkg/networkservice/ipam/point2pointipam/server.go
@@ -90,16 +90,16 @@ func (s *ipamServer) Request(ctx context.Context, request *networkservice.Networ
 
 	excludeIP4, excludeIP6 := exclude(ipContext.GetExcludedPrefixes()...)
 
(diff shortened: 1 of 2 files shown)


denis-tingaikin:
It makes sense for pass thought cases.

It seems the root cause of this issue is https://github.com/networkservicemesh/sdk/issues/724

denis-tingaikin (changes requested):
Let's try to consider this issue in solution for https://github.com/networkservicemesh/sdk/issues/724

edwarnicke:
@denis-tingaikin As noted in https://github.com/networkservicemesh/sdk/issues/724#issuecomment-799524979 I think we may need to handle this in a more localized manner.

denis-tingaikin:
@edwarnicke ok, then this can be merged :)

denis-tingaikin (approved):

Issue #760, opened 2021-03-09
Serialize in registry chain


Bolodya1997:
# Intro
For `networkservice` chain we have `serialize` chain element serializing all `Request`/`Close` events by Connection ID. With such approach we have a guarantee that in any time in any chain element we never have 2 concurrent events for the same Connection ID.

# Issue
At the moment there is no such chain element in registry chain (serializing all `Register`/`Unregister` events by Endpoint Name) but we also need it to reduce concurrency complicity.
The main problem for creating such chain element for registry chain is Endpoint Name generated somewhere in the top right place in the hop chain.
```
   endpoint -> NSMgr ->
-> NSMgr    -> Registry ->
-> Registry -> another domain Registry <-- Name can be generated even here
```
So the approach we are using in the networkservice chain cannot be directly used for the first `Register` event in the chain.
```
   ...
   serialize <-- here we need to take an executor and store it in context, but we don't have a Name to use it as a key
   ...
   setid <-- here we have a Name
   ...
```

# Solution
Instead of storing `Executor`:
```go
type Executor interface {
    AsyncExec(f func()) <-chan struct{}
}
```
we can store in context `MultiExecutor`:
```go
type MultiExecutor interface {
    AsyncExec(id string, f func()) <-chan struct{}
}
```
and so it should resolve the issue.

PR #767, opened 2021-03-12
Fix localbypass NSE registry server

Bolodya1997:
# Intro
When endpoint registers on NSMgr, NSMgr stores its URL locally and passes NSE registration with URL replaced to NSMgr URL in `localbypass` NSE server chain element.
```
Endpoint  -Register->  NSMgr    : { URL: nseURL }
NSMgr     -Register->  Registry : { URL: nsmgrURL }
```
When NSMgr needs to find the endpoint, localbypass replaces URL back:
```
NSMgr (localbypass)  <-Find-  Registry            : { URL: nsmgrURL }
NSMgr                <-Find-  NSMgr (localbypass) : { URL: nseURL }
```
`localbypass` can't store `name -> URL` mapping before passing NSE registration to the next, because it doesn't yet have `name`. So there is a moment in time when NSE is already stored in Registry but not yet exists in `localbypass` mappings.

# Issue
Sometimes we can get locally registered NSE with `{ URL: nsmgrURL }` from `Find`.

# Solution
#770

### DIFF
--- a/pkg/registry/common/localbypass/find_server.go
+++ b/pkg/registry/common/localbypass/find_server.go
@@ -18,19 +18,32 @@ package localbypass
 
 import (
 	"github.com/networkservicemesh/api/pkg/api/registry"
-
-	"github.com/networkservicemesh/sdk/pkg/tools/stringurl"
 )
 
 type localBypassNSEFindServer struct {
-	nseURLs *stringurl.Map
-
+	*localBypassNSEServer
 	registry.NetworkServiceEndpointRegistry_FindServer
 }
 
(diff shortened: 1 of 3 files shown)


denis-tingaikin (approved):
In general, I don't mind these changes.

Question: Is it blocked by https://github.com/networkservicemesh/sdk/issues/760?

Bolodya1997:
> Question: Is it blocked by #760?

Nope.

denis-tingaikin:
@Bolodya1997 
OK, then lets continue discuss a solution here: https://github.com/networkservicemesh/sdk/issues/770

PR #769, opened 2021-03-15
Logrus logger shouldn't overwrite time for logs

Bolodya1997:
# Issue
`logruslogger.New` and `logruslogger.FromSpan` overwrite logs time with `time.Now`. So for any prolonged usage of this logger we get same timestamps for all logs.
# Solution
Stop using `.WithTime(time.Now())` and let logs to use their own timestamps.


denis-tingaikin (approved):
Good catch :)

PR #749, opened 2021-03-05
Rework registry NSE expire with timer.AfterFunc

Bolodya1997:
# Motivation
`time.AfterFunc` has few issues that have made registry NSE expire overcomplicated. `timer.AfterFunc` was designed to reduce such complicities.

PR #771, opened 2021-03-16
Add registry serialize chain elements

Bolodya1997:
# Issue
https://github.com/networkservicemesh/sdk/issues/760

# Solution
Extracts `serializectx` from `pkg/networkservice/serialize`, adds `pkg/registry/serialize`.

Closes https://github.com/networkservicemesh/sdk/issues/760.
Unblocks #749, #772.


pkg/networkservice/common/serialize/README.md
```diff
@@ -6,16 +6,18 @@
...
 
 Correct event firing chain element example:
 ```go
 func (s *eventServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
-	executor := serialize.Executor(ctx)
+	executor := serializectx.Executor(ctx, request.GetConnection().GetId())
 	go func() {
 		executor.AsyncExec(func() {
-			_, _ = next.Server(ctx).Request(serialize.WithExecutor(context.TODO(), executor), request)
+			_, _ = next.Server(ctx).Request(serializectx.WithExecutor(context.TODO(), executor), request)
```
haiodo, line 20: Why we use context.TODO()?

pkg/networkservice/common/serialize/README.md
```diff
@@ -26,7 +28,8 @@ func (s *eventServer) Request(ctx context.Context, request *networkservice.Netwo
 
 	go func() {
 		executor.AsyncExec(func() {
-			_, _ = next.Server(ctx).Close(serialize.WithExecutor(context.TODO(), executor), conn)
+			// We don't pass executor to the Close context. Please don't do it.
+			_, _ = next.Server(ctx).Close(context.TODO(), conn)
```
haiodo, line 32: Same here

pkg/networkservice/common/serialize/README.md
Bolodya1997: It is only to show that we are not using `Request` context (`ctx`) for generated `Request`, `Close` events.

Bolodya1997: Actualy it is always something like `s.chainCtx` instead of `context.TODO()` in real life examples.

haiodo (approved):

pkg/networkservice/common/serialize/README.md
haiodo: ok

PR #772, opened 2021-03-16
Unify expire servers

Bolodya1997:
# Issue
Currently we have 3 different expire servers implementing different partially duplicating each other expiration algorithms:
1. timeout in networkservice chain
2. expire in NSE registry chain
3. expire in NS registry chain

# Solution
Create `expire.Manager` incapsulating all expiration algorithm logic and use it in all 3 places.

~Depends on https://github.com/networkservicemesh/sdk/pull/862.~


denis-tingaikin (changes requested):
I think this is an incorrect way for NSM chains.
We should do not do mix registres and netwokservices...

Bolodya1997:
> We should do not do mix registres and netwokservices...

We are not actually mixing them - there is a common algorithm and tools for implementing common algorithm.

denis-tingaikin:
@edwarnicke Do you mind this direction? The main idea is to move all expiration logic into one entity and use it whatever where we have an entity that can expire.

edwarnicke (approved):
@Bolodya1997 In addition to being excellently coded, I really appreciate the quality of your documentation.  It is *excellent*.

edwarnicke:
@Bolodya1997 Looks like this needs a rebase.

denis-tingaikin (approved):

PR #780, opened 2021-03-19
fix: monitor server store duplicates with the same connections

denis-tingaikin:

### DIFF
--- a/pkg/networkservice/chains/nsmgr/server_test.go
+++ b/pkg/networkservice/chains/nsmgr/server_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/goleak"
 	"google.golang.org/grpc"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 	"github.com/networkservicemesh/api/pkg/api/networkservice/mechanisms/cls"
@@ -46,6 +47,7 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/chain"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/inject/injecterror"
+	"github.com/networkservicemesh/sdk/pkg/tools/grpcutils"
 	"github.com/networkservicemesh/sdk/pkg/tools/sandbox"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/networkservice/chains/nsmgr/server_test.go
```diff
@@ -350,6 +352,67 @@ func TestNSMGR_ConnectToDeadNSE(t *testing.T) {
...
+	stream, err := montiorClient.MonitorConnections(ctx, &networkservice.MonitorScopeSelector{
+		PathSegments: conn.Path.PathSegments,
+	})
+	require.NoError(t, err)
+	require.NoError(t, ctx.Err())
+
+	msg, err := stream.Recv()
+	require.NoError(t, err)
+
+	require.Len(t, msg.Connections, 1)
```
denis-tingaikin, line 412: Note: On the master branch this check fails due to len(msg.Connections) == 2

pkg/networkservice/common/monitor/server.go
```diff
@@ -79,14 +79,22 @@ func (m *monitorServer) MonitorConnections(selector *networkservice.MonitorScope
...
 	if err == nil {
+		if len(conn.GetPath().PathSegments) == 0 {
+			return nil, errors.New("resp.Path.PathSegments has inccorect lengrh = 0")
+		}
+		eventConn := conn.Clone()
 		m.executor.AsyncExec(func() {
-			m.connections[eventConn.GetId()] = eventConn
+			id := eventConn.GetPath().PathSegments[0].GetId()
+			eventConn.Id = id
+			eventConn.Path.Index = 0
```
denis-tingaikin, line 90: I'm not sure that we should change index here, probabbly we should store the current index.

pkg/networkservice/common/monitor/server.go
```diff
@@ -79,14 +79,22 @@ func (m *monitorServer) MonitorConnections(selector *networkservice.MonitorScope
...
 	conn, err := next.Server(ctx).Request(ctx, request)
-	eventConn := conn.Clone()
 	if err == nil {
+		if len(conn.GetPath().PathSegments) == 0 {
+			return nil, errors.New("resp.Path.PathSegments has inccorect lengrh = 0")
+		}
+		eventConn := conn.Clone()
 		m.executor.AsyncExec(func() {
-			m.connections[eventConn.GetId()] = eventConn
+			id := eventConn.GetPath().PathSegments[0].GetId()
```
denis-tingaikin, line 88: Root cause here: nsmgr could duplicate request from nsc and from forwarder here(because they have different connection id).

pkg/networkservice/common/monitor/server.go
```diff
@@ -79,14 +79,20 @@ func (m *monitorServer) MonitorConnections(selector *networkservice.MonitorScope
...
 func (m *monitorServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
 	conn, err := next.Server(ctx).Request(ctx, request)
-	eventConn := conn.Clone()
 	if err == nil {
+		if len(conn.GetPath().PathSegments) == 0 {
+			return nil, errors.New("resp.Path.PathSegments has inccorect lengrh = 0")
+		}
+		eventConn := conn.Clone()
 		m.executor.AsyncExec(func() {
-			m.connections[eventConn.GetId()] = eventConn
```
denis-tingaikin, line 85: The root cause here: nsmgr could duplicate request from nsc and from forwarder here(because they have different connection id).

denis-tingaikin:
Is not actual more

PR #768, opened 2021-03-12
Add support ipv6 for point2pointipam

artem-belov:
Fixes #411 

IPPool is a RB tree, containing ip intervals storing in the pool
IPPool adds 128 bit support to keep IPv6 addresses

BenchmarkIPPool | ops | ns/op | B/op | allocs/op
-- | -- | -- | -- | --
BenchmarkIPPool/IPPool | 566 | 2161121 | 440206 | 22007
BenchmarkIPPool/RoaringBitmap | 1359 | 824205 | 218841 | 11106
BenchmarkIPPool/PrefixPool | 2 | 1803298500 | 1001995744 | 18689007
  |   |   |   |  
BenchmarkIPPool/IPPool-2 | 1009 | 1079308 | 440223 | 22008
BenchmarkIPPool/RoaringBitmap-2 | 2985 | 351601 | 218850 | 11106
BenchmarkIPPool/PrefixPool-2 | 2 | 1355259309 | 1021856016 | 19835510
  |   |   |   |  
BenchmarkIPPool/IPPool-4 | 1126 | 1005249 | 440240 | 22008
BenchmarkIPPool/RoaringBitmap-4 | 5494 | 193221 | 218847 | 11106
BenchmarkIPPool/PrefixPool-4 | 2 | 1365976629 | 1021857340 | 19835525
  |   |   |   |  
BenchmarkIPPool/IPPool-8 | 1011 | 1095871 | 440265 | 22009
BenchmarkIPPool/RoaringBitmap-8 | 6398 | 163107 | 218842 | 11106
BenchmarkIPPool/PrefixPool-8 | 2 | 1319444212 | 1012589580 | 19802446
  |   |   |   |  
BenchmarkIPPool/IPPool-16 | 996 | 1102098 | 440250 | 22008
BenchmarkIPPool/RoaringBitmap-16 | 7626 | 179327 | 218851 | 11106
BenchmarkIPPool/PrefixPool-16 | 2 | 1408024424 | 1012589068 | 19802437

PR #787, opened 2021-03-24
Replace PrefixPool with IPPool in excludedPrefixesServer

### LINKED ISSUE #389: Rework prefixpool to work with using RoaringBitmap
## Motivation

We've migrated prefix pool from mono repo to sdk https://github.com/networkservicemesh/sdk/tree/master/pkg/tools/prefixpool, but It looks like many things can be refactored and simplified with using this library:
https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/ipam/point2pointipam/server.go#L31

artem-belov:
Fixes #389
Benchmark results #768


pkg/networkservice/common/excludedprefixes/server_test.go
```diff
@@ -59,7 +59,7 @@ func TestNewExcludedPrefixesService(t *testing.T) {
 }
 
 func TestCheckReloadedPrefixes(t *testing.T) {
-	prefixes := []string{"172.16.1.0/24", "10.32.0.0/12", "10.96.0.0/12"}
+	prefixes := []string{"10.32.0.0/12", "10.96.0.0/12", "172.16.1.0/24"}
```
denis-tingaikin, line 62: Why tests have changed? Looks like it is more correct, but could you please clarify this :)

pkg/tools/ippool/ippool.go
```diff
@@ -553,6 +582,102 @@ func (node *treeNode) maximumNode() *treeNode {
...
+		// exclude the last /64 network if it is not full
+		if end.low != math.MaxUint64 {
+			end.high--
+		}
+
+		// get available prefixes bigger than /64
+		for start.high <= end.high {
+			z := trailingZeros(start.high)
+
+			if z == 64 {
```
denis-tingaikin, line 617: Can we make `64` as a constant?

pkg/tools/ippool/ippool.go
```diff
@@ -230,6 +235,30 @@ func (tree *IPPool) PullP2PAddrs(exclude ...*IPPool) (srcNet, dstNet *net.IPNet,
...
+	}
+	for it.node.Left != nil {
+		it.node = it.node.Left
+	}
+
+	for node := it.Next(); node != nil; node = it.Next() {
+		prefixes = append(prefixes, node.getPrefixes(clone.ipLength)...)
+	}
+
+	return
```
denis-tingaikin, line 259: I wonder why linter does not trigger at this line. 
Can we do not use return without values?

denis-tingaikin (approved):

pkg/networkservice/common/excludedprefixes/server_test.go
artem-belov: IPPool sort subnets with the bit order within it's RB tree, so I changed prefixes order to check in the test

pkg/tools/ippool/ippool.go
artem-belov: Done

pkg/tools/ippool/ippool.go
artem-belov: Done

PR #789, opened 2021-03-25
Generate nse.Name on the Endpoint side

### LINKED ISSUE #770: Issue with swapping between Endpoint URL and NSMgr URL
# Intro
When endpoint registers on NSMgr, NSMgr stores its URL locally and passes NSE registration with URL replaced to NSMgr URL in localbypass NSE server chain element.

```
Endpoint  -Register->  NSMgr    : { URL: nseURL }
NSMgr     -Register->  Registry : { URL: nsmgrURL }
```
When NSMgr needs to find the endpoint, localbypass replaces URL back:
```
NSMgr (localbypass)  <-Find-  Registry            : { URL: nsmgrURL }
NSMgr                <-Find-  NSMgr (localbypass) : { URL: nseURL }
```
localbypass can't store name -> URL mapping before passing NSE registration to the next, because it doesn't yet have name. So there is a moment in time when NSE is already stored in Registry but not yet exists in localbypass mappings.

# Issue
Sometimes we can get locally registered NSE with `{ URL: nsmgrURL }` from `Find`.

# Possible solutions

## Without changing api

`localbypass` can make `Register` event + refresh `Register` event and filter out from `Find` all endpoints with `nse.Url == NSMgr.Url` and not presented in the local mappings - #767.

## With changing api

Instead of single `nse.Url` field we can make `nse.Url`, `nse.NsmgrUrl` fields and don't make any substitutions on `Register`, `Unregister`, `Find`. We only need to select right URL and we can easy make it comparing `nse.NsmgrUrl` with our `NSMgr.Url`.

Bolodya1997:
Closes #770.


pkg/registry/chains/memory/server.go
```diff
@@ -40,9 +40,9 @@ import (
 func NewServer(ctx context.Context, expiryDuration time.Duration, proxyRegistryURL *url.URL, options ...grpc.DialOption) registryserver.Registry {
 	nseChain := chain.NewNetworkServiceEndpointRegistryServer(
 		serialize.NewNetworkServiceEndpointRegistryServer(),
+		setid.NewNetworkServiceEndpointRegistryServer(),
```
haiodo, line 43: Why we need setid here?

Bolodya1997: It is a chain for the memory registry. So we need to check for the collisions in registry.

haiodo: maybe setid is a wrong name in this case?

Bolodya1997: Extracted server + error to `checkid` package.

denis-tingaikin (approved):
In general looks good to me

pkg/registry/checkid/error.go
```diff
@@ -0,0 +1,28 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checkid
+
+import "fmt"
+
+// DuplicateError is an error type for the duplicate name error
+type DuplicateError struct {
```
denis-tingaikin, line 22: Can we make this private? As I can see this is public to be useable in the tests. So let's try to use chain with server/client to get the error naturally.

pkg/registry/checkid/gen.go
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checkid
+
+import "sync"
+
+//go:generate go-syncmap -output names_urls_map.gen.go -type namesUrlsMap<string,string>
+
+type namesUrlsMap sync.Map
```
denis-tingaikin, line 23: Can we just use `stringurl.Map`?

pkg/registry/checkid/error.go
Bolodya1997: We have different `setid`, `checkid` packages for client/server side (https://github.com/networkservicemesh/sdk/pull/789#discussion_r603436796). So this error can't be private.

pkg/registry/checkid/gen.go
Bolodya1997: Yep, replaced with `stringurl.Map`.

denis-tingaikin (approved):

PR #793, opened 2021-03-25
Close connection in connect on next.Request error

artem-belov:
Fixes example:
If forwarder cannot setup mechanism, already opened connection should be closed

### DIFF
--- a/pkg/networkservice/common/connect/server.go
+++ b/pkg/networkservice/common/connect/server.go
@@ -116,7 +116,7 @@ func (s *connectServer) Request(ctx context.Context, request *networkservice.Net
 	conn, err = next.Server(ctx).Request(ctx, request)
 	// Close connection if next.Server Request finished with error
 	if err != nil && !refreshRequest {
-		_, cErr := c.client.Close(ctx, request.Connection.Clone())
+		_, cErr := s.Close(ctx, request.Connection.Clone())
 		if cErr != nil {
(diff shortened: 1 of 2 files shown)


artem-belov:
Related to #724

denis-tingaikin (approved):

denis-tingaikin:
@Bolodya1997 Could you also have a look?

Bolodya1997:
I am afraid that it wouldn't actually solve the problem with unclosed connection on a remote side:
```
... -> discover -> ... -> connect -> remote
```
1. remote returns an error, but there is an unclosed connection on remote
2. connect drops the corresponding client, because it was the first request to the remote URL and it has failed
3. discover tries to close connection
4. connect tells that there is no client and returns an error

As a result we still have a connection on a remote side.

artem-belov:
@Bolodya1997  You are presenting some different behaviour. That PR is fixing next issue:
1. Forwarder send request to the next path element (nsmgr). It successfuly done and client.Request returns connection
2. Forwarder is trying to configure mechanisms and failes. It just returns an error to previous steps and does not close already configured connection.
With the fix, if forwarder failed configuring mechanisms, it send `Close` to the client chain, so connection will be closed. Client chain will be deleted only if there is no more connections on it.
Discover server does not involve in the process totally.

Bolodya1997:
Oops, got something wrong with my eyes :)

Bolodya1997 (approved):

PR #795, opened 2021-03-29
Refactor chains/nsmgr to use Option pattern

### LINKED ISSUE #733: Refactor chains/nsmgr to use Option pattern
## Motivation

We can simplify using of our chains with
[Option pattern](https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md)

See at an example: https://github.com/networkservicemesh/sdk/commit/9007c4530373d301993ecd8985a1c9abebe359b2

Moreover, this can super help us with minimization of manual changes of upstream repositories.

Entry point: https://github.com/networkservicemesh/sdk/blob/master/pkg/networkservice/chains/nsmgr/server.go

haiodo:
Fixes #733 
Refactor chains/nsmgr to use Option pattern

### DIFF
--- a/pkg/networkservice/chains/nsmgr/server.go
+++ b/pkg/networkservice/chains/nsmgr/server.go
@@ -23,6 +23,8 @@ import (
 	"context"
 	"time"
 
+	"github.com/google/uuid"
+
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
@@ -30,6 +32,7 @@ import (
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/chains/client"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/chains/endpoint"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/common/authorize"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/connect"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/discover"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/excludedprefixes"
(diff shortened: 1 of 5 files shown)


pkg/networkservice/chains/nsmgr/server.go
```diff
@@ -69,20 +70,76 @@ type nsmgrServer struct {
...
+// WithName - set a nsmgr name, a default name is `Nsmgr`.
+func WithName(name string) Option {
+	return func(o *serverOptions) {
+		o.name = name
+	}
+}
+
+// WithURL - set a public URL address for NetworkServiceManager, it is used to access endpoints within this NSMgr from remote clients.
+// Default value is not set, and endpoints address will not be changed.
+func WithURL(url string) Option {
```
denis-tingaikin, line 117: URL should be required parameter.

pkg/networkservice/chains/nsmgr/server.go
```diff
@@ -69,20 +70,76 @@ type nsmgrServer struct {
...
-//           authzServer - authorization server chain element
 //           tokenGenerator - authorization token generator
-//           registryCC - client connection to reach the upstream registry, could be nil, in this case only in memory storage will be used.
-// 			 clientDialOptions -  a grpc.DialOption's to be passed to GRPC connections.
-func NewServer(ctx context.Context, nsmRegistration *registryapi.NetworkServiceEndpoint, authzServer networkservice.NetworkServiceServer, tokenGenerator token.GeneratorFunc, registryCC grpc.ClientConnInterface, clientDialOptions ...grpc.DialOption) Nsmgr {
+//			 options - a set of NSMGR options.
+func NewServer(ctx context.Context, tokenGenerator token.GeneratorFunc, options ...Option) Nsmgr {
+	opts := &serverOptions{
+		authorizeServer: authorize.NewServer(authorize.Any()),
+		name:            "Nsmgr",
```
denis-tingaikin, line 131:
```suggestion
		name:            "nsmgr-"+uuid.New().String(),
```

denis-tingaikin (changes requested):

pkg/networkservice/chains/nsmgr/server.go
haiodo: It is not required for Local -> Nsmgr -> local only cases and for cases if we doesn't use public Url for nsmgr.

denis-tingaikin: OK, currently we have not a test for this scenario actually. Let's then cover it :)

pkg/networkservice/chains/nsmgr/server.go
haiodo: Thanks.

pkg/networkservice/chains/nsmgr/server.go
haiodo: Added test with a Unix Socket nsmgr and empty Url.

denis-tingaikin:
Note: For some reason new test produces goleaks: https://github.com/networkservicemesh/sdk/pull/795/checks?check_run_id=2218240245

pkg/tools/sandbox/builder.go
```diff
@@ -269,20 +300,37 @@ func (b *Builder) newNSMgr(ctx context.Context, address string, registryURL *url
...
+	if b.useUnixSockets {
+		serveURL = grpcutils.TargetToURL(address)
+	} else {
+		listener, err := net.Listen("tcp", address)
+		b.require.NoError(err)
+		serveURL = grpcutils.AddressToURL(listener.Addr())
+		b.require.NoError(listener.Close())
+	}
+
+	nsmgrName := "nsmgr-" + uuid.New().String()
```
denis-tingaikin, line 314: let's use default value
```suggestion
```

pkg/tools/sandbox/builder.go
```diff
@@ -360,6 +409,16 @@ func (b *Builder) SetupRegistryClients(ctx context.Context, node *Node) {
...
 }
 
+// newAddress - will return a new public address, if unixSockets are used prefix will be used to make uniq files.
+func (b *Builder) newAddress(prefix string) string {
+	if !b.useUnixSockets {
+		return "127.0.0.1:0"
+	}
+
+	b.addressID++
+	return fmt.Sprintf("unix:%s/%s_%d.sock", b.sockPath, prefix, b.addressID)
```
denis-tingaikin, line 419: Can we use uuid instead of `b.addressID`?

haiodo: Sure :-) Fixed.

haiodo:
@denis-tingaikin Unix sockets with uuid id inside are not working, so I've revert a logic to use an index, and since index are inside random temporary folder it should be fully safe :-)

denis-tingaikin (approved):

PR #796, opened 2021-03-30
Fix breakage in heal due to sending incorrect Connection on Close

edwarnicke:


belov3004:
@edwarnicke The issue with cached connection in monitor is based on Client does not know that connection was changed by heal until first refresh request. So we have to change Close some way to put there correct connection. The other way is to inform Client on connection changes. For example we can make Long Poll refresh requests... Or some other way.

artem-belov:
@edwarnicke The issue with cached connection in monitor is based on Client does not know that connection was changed by heal until first refresh request. So we have to change Close some way to put there correct connection. The other way is to inform Client on connection changes. For example we can make Long Poll refresh requests... Or some other way.

haiodo:
@edwarnicke As I understand @artem-belov correctly, since we do not have a way to have latest connection, on some heals, in case endpoint is different, Close could go with a different way if it is not updated to latest one.

Where few variants possible, on client, before doing Close, we should be sure we had a latest connection object. 
This issue is created to cover this stuff: https://github.com/networkservicemesh/sdk/issues/797

Also @denis-tingaikin is looking into this issue right now, why it had problems without current PR, and possible root cause is serialize chain element is doing something wrong on Request and Close going on same time.

denis-tingaikin:
>Also @denis-tingaikin is looking into this issue right now, why it had problems without current PR, and possible root cause is serialize chain element is doing something wrong on Request and Close going on same time.

Potentially serialize doing things wrong due to caching connection in heal servers.

denis-tingaikin (approved):
I think we can go with this patch. It workarounds the issue at this moment.

Heal rework will be continued in https://github.com/networkservicemesh/sdk/issues/797

edwarnicke:
> @edwarnicke As I understand @artem-belov correctly, since we do not have a way to have latest connection, on some heals, in case endpoint is different, Close could go with a different way if it is not updated to latest one.

Ah... OK... so the problem we are dealing with is that if a heal occurs, it may get a different Path than we got as a return to the original request.  The original requestor won't know till the first Refresh cycle.

> 
> Where few variants possible, on client, before doing Close, we should be sure we had a latest connection object.
> This issue is created to cover this stuff: #797
> 
> Also @denis-tingaikin is looking into this issue right now, why it had problems without current PR, and possible root cause is serialize chain element is doing something wrong on Request and Close going on same time.

Good to know :)

PR #798, opened 2021-03-31
Fix issue with heal close

Bolodya1997:
# Issue
`Close` doesn't stop healing.

# Solution
Use context created for healing instead of `f.ctx` in healing.

### DIFF
--- a/pkg/networkservice/common/heal/server.go
+++ b/pkg/networkservice/common/heal/server.go
@@ -179,7 +179,7 @@ func (f *healServer) healAsNeeded(baseCtx context.Context, request *networkservi
 		return
 	}
 
-	ctx, cancel := context.WithCancel(f.ctx)
+	ctx, cancel := f.healContext(baseCtx)
 	defer cancel()
 	id := request.GetConnection().GetId()
 	<-f.cancelHealMapExecutor.AsyncExec(func() {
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/heal/server.go
```diff
@@ -356,43 +350,38 @@ func (f *healServer) restoreConnection(ctx, baseCtx context.Context, request *ne
...
 
 	if err != nil {
-		f.processHeal(baseCtx, request.Clone(), opts...)
+		f.processHeal(ctx, baseCtx, request.Clone(), opts...)
 	}
 }
 
-func (f *healServer) processHeal(baseCtx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) {
-	logEntry := log.FromContext(f.ctx).WithField("healServer", "processHeal")
+func (f *healServer) processHeal(ctx, baseCtx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) {
```
denis-tingaikin, line 367: Let's just pass one heal context which will include candidates :) 
```suggestion
func (f *healServer) processHeal(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) {
```

Bolodya1997: Yep, fixed.

denis-tingaikin (approved):

PR #810, opened 2021-04-08
Add Client Monitor to the Heal Client

### LINKED ISSUE #797: Add Client Monitor to the Heal Client
- Move `Monitor Connections` client from `heal server` to `heal client`
- Inform `Heal Server` from `Heal Client` about new connections and connection changes if `Heal Server` is presented in the chain
- Cache changed connections on `Heal Client` and swap connection to the newer version on `Close` request.

d-uzlov:
Fixes #797


pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
+	//  will we receive a new event with this channel?
+	//  If not, then we will block on waiting and return an error eventually
+	verifySuccessCh, err := u.addToExpectedConnections(request.GetConnection())
+	if err != nil {
+		return nil, err
+	}
+
 	conn, err := next.Client(ctx).Request(ctx, request, opts...)
+	if err != nil {
+		return conn, err
```
Bolodya1997, line 85: We have the following pattern for such case:
```go
func Request() {
    conn, err := next.Request()
    if err != nil {
        return nil, err // <-- return `nil` instead of `conn`
    }
    err := doSomething()
    if err != nil {
        next.Close(conn) // <-- close `conn` if we failed to process it
        return nil, err // <-- return `nil` instead of `conn`
    }
    return conn, nil
}
```

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
+	//if err != nil {
+	//	return conn, errors.Wrapf(err, "error converting pathSegment.GetExpires() to time.Time: %+v", pathSegment.GetExpires())
+	//}
+
+successVerifyLoop:
+	for {
+		select {
+		case <-u.ctx.Done():
+			u.removeFromExpectedConnections(conn)
+			return conn, errors.Errorf("can't wait for connection %v (segment %v) verification: chain context was canceled", conn.GetId(), conn.GetNextPathSegment())
```
Bolodya1997, line 100: Please look pattern.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
-		registerClient(u.ctx, conn, u.cc)
+			// TODO
+			//  original code used expire time for waiting but it didn't matter because
+			//  originally we created a monitor and received INITIAL_STATE_TRANSFER event which always was instantaneous,
+			//  regardless if there were errors
+			//  now it's unclear how long we should wait
+		//case <-time.After(time.Until(expireTime)):
+		case <-time.After(time.Second):
+			u.removeFromExpectedConnections(conn)
+			return conn, errors.New("timeout expired but we couldn't verify that connection was established")
```
Bolodya1997, line 116: Please look pattern.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
+				return false
+			}
+		}
+		return true
+	}
+
+	for u.ctx.Err() == nil {
+		event, err := monitorClient.Recv()
+		if err != nil {
+			if u.ctx.Err() != nil {
```
Bolodya1997, line 157: I suppose that we just need to call heal server for all the connections we own. There possibly can be a case when gRPC connection breaks and so connect server closes the client context, so context can be closed even if we really need to heal server. So heal server would decide by itself if there is a need to heal connections IDs given from the heal client.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
+		conn = cachedConn
+	} else {
+		// TODO log this?
+	}
+
 	return next.Client(ctx).Close(ctx, conn, opts...)
 }
+
+func (u *healClient) listenToConnectionChanges(heal HealRequestFunc, errCh chan error) {
+	monitorClient, err := u.cc.MonitorConnections(u.ctx, &networkservice.MonitorScopeSelector{})
```
Bolodya1997, line 134: We can use such path segments as a selector for all connections that have us as a previous path segment.
```
[]{ { Name: currentPathSegment.Name }, { Name: "" } }
```

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
+	if err != nil {
+		errCh <- errors.Wrap(err, "MonitorConnections failed")
+		return
+	}
+	close(errCh)
+
+	// TODO should we really compare paths from the beginning?
+	pathStartsWith := func(c *networkservice.Connection, reference *networkservice.Connection) bool {
```
Bolodya1997, line 142: If we use the selector mentioned above, we only need to check `conn.GetPrevPathSegment().Id` instead of full path prefix.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
+				return
+			}
+			continue
+		}
+
+		var connsToHeal []string
+
+		switch event.GetType() {
+		case networkservice.ConnectionEventType_INITIAL_STATE_TRANSFER: // TODO is this case special?
+			fallthrough
```
Bolodya1997, line 172: Just do nothing, there are no requests from us at the moment.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,243 @@ package heal
...
+				for _, svReq := range u.expectedConns {
+					if pathStartsWith(conn, svReq.prefixConn) {
+						svReq.ch <- conn.GetPath().GetPathSegments()[svReq.prefixConn.GetPath().GetIndex()+1]
+					}
+				}
+			}
+			u.expectedConnsMutex.Unlock()
+
+		case networkservice.ConnectionEventType_DELETE:
+			if heal != nil {
```
Bolodya1997, line 201: It must be not nil, so you may add this check somewhere out of this function.

d-uzlov: When I talked with Denis Tingaikin about heal callback few days ago, he said that heal server may not be present in the chain, therefore it's allowed for this variable to be nil, and heal client in this case should still listen to changes to replace the connection in the Close() call.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,208 @@ package heal
...
+		}
+		return true
+	})
+}
+
+func (u *healClient) addToExpectedConnections(conn *networkservice.Connection) *connVerificationRequest {
+	connAlreadyExists := false
+	<-u.conCacheExecutor.AsyncExec(func() {
+		if cachedConn, ok := u.connCache[conn.GetId()]; ok {
+			if cachedConn.GetCurrentPathSegment().GetId() == conn.GetCurrentPathSegment().GetId() {
```
Bolodya1997, line 174: `conn.GetCurrentPathSegment().GetId()` is the same as `conn.GetId()`.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,208 @@ package heal
...
+
+	return &req
+}
+
+func (u *healClient) confirmConnectionSuccess(conn *networkservice.Connection, req *connVerificationRequest) error {
+	timeoutCh := time.After(time.Millisecond * 100)
+	for {
+		select {
+		case segment := <-req.connectionSuccessCh:
+			if conn.GetNextPathSegment().Equal(segment) {
```
Bolodya1997, line 199: I don't think that we need such check. We have already filtered out all connections except the ones going from us. We now that for the `segment` previous `conn.Id` matches this `conn.Id`. So there is no need to send a `segment` and check it for something, because we have already checked everything we want.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,208 @@ package heal
...
-type RegisterClientFunc func(context.Context, *networkservice.Connection, networkservice.MonitorConnectionClient)
-
 type healClient struct {
-	ctx context.Context
-	cc  networkservice.MonitorConnectionClient
+	ctx              context.Context
+	cc               networkservice.MonitorConnectionClient
+	initOnce         sync.Once
+	initErr          error
+	connCache        map[string]*networkservice.Connection
```
Bolodya1997, line 40: Probably it would be better to create a `sync.Map` of something like that:
```go
type connnectionInfo struct {
    conn *networkservice.Connection
    state connState // { initial, ready, broken }
    updateCh chan struct{}
}
```
instead of managing multiple maps? And so it would be easier to handle different refresh/refresh on broken scenarios because there is a `state`.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,208 @@ package heal
...
+		}
+
+		switch event.GetType() {
+		case networkservice.ConnectionEventType_INITIAL_STATE_TRANSFER:
+			continue
+		case networkservice.ConnectionEventType_UPDATE:
+			<-u.conCacheExecutor.AsyncExec(func() {
+				for _, conn := range event.GetConnections() {
+					if _, ok := u.connCache[conn.GetId()]; ok {
+						u.connCache[conn.GetId()] = conn.Clone()
```
Bolodya1997, line 128: Probably `conn.GetPrevPathSegment().GetId()`?

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,208 @@ package heal
...
+	}
+	close(errCh)
+
+	for {
+		event, err := monitorClient.Recv()
+		if err != nil {
+			if heal != nil {
+				<-u.conCacheExecutor.AsyncExec(func() {
+					for id := range u.connCache {
+						heal(id, true)
```
Bolodya1997, line 114: We probably want to pass updated path from `connCache` to the `heal` to make it work properly. I suppose that it is not enough to pass only `id` because in such case `heal` will be invoked with a new client and there will be no `connCache`.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,208 @@ package heal
...
+	if u.initErr != nil {
+		return nil, u.initErr
+	}
+
+	verificationRequest := u.addToExpectedConnections(request.GetConnection())
+	if verificationRequest != nil {
+		defer close(verificationRequest.stopCh)
+	}
+
 	conn, err := next.Client(ctx).Request(ctx, request, opts...)
```
Bolodya1997, line 71: We probably want to use here updated path from `connCache`.

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,208 @@ package heal
...
-	return conn, err
+
+	return conn, nil
 }
 
 func (u *healClient) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
+	if cachedConn, ok := u.removeConnectionFromMonitor(conn); ok {
+		conn = cachedConn
+	} else {
+		log.FromContext(ctx).WithField("healClient", "Close").Warnf("can't find connection in cache, id=%v", conn.GetId())
```
Bolodya1997, line 92: No entry in cache means no connection exists, here we can simply return `new(empty.Empty), nil`.

pkg/networkservice/common/heal/client.go
d-uzlov: Heal client does 2 things:
1. Listening of it's own connection, to make sure Close() always gets the correct connection information (UPDATE event).
2. Monitoring of next segment of connection, to request heal if there are issues (DELETE event).

So on UPDATE event we search for our own connections, and in DELETE we search for connections that have our connections as previous path segment.

pkg/networkservice/common/heal/client.go
d-uzlov: I don't think so.
When Request() is called on client, it gets fresh path from previous chain elements.
If we replace it with the path we have, that path may be outdated.

pkg/networkservice/common/heal/client.go
Bolodya1997: How can we receive current connection from the next hop if the next hop sends updates only for its connections?
https://github.com/networkservicemesh/sdk/blob/2de39a89505e17de4e4c3daac4161f02bfbd88d0/pkg/networkservice/common/monitor/server.go#L89-L100

pkg/networkservice/common/heal/context.go
```diff
@@ -18,23 +18,32 @@ package heal
...
 
 type contextKeyType string
 
-func withRegisterClientFunc(parent context.Context, registerClientFunc RegisterClientFunc) context.Context {
+// requestHealFuncType - function used to inform heal server that connection should be healed
+//                       - conn              - connection to be healed
+//                       - restoreConnection - flag to specify whether what to heal.
+//                                             When false, only heals this particular connection.
+//                                             When true also tries to heal physical connection to next hop.
+type requestHealFuncType func(conn *networkservice.Connection, restoreConnection bool)
```
denis-tingaikin, line 36: It looks like we actually have two functions that hid by the boolean flag. Please consider removing the boolean flag and create two separate functions:
1. onRestoreConnection(conn *networkservice.Connection)
2. onHealConnection(conn *networkservice.Connection)

pkg/networkservice/common/heal/client.go
```diff
@@ -45,15 +57,174 @@ func NewClient(ctx context.Context, cc networkservice.MonitorConnectionClient) n
...
+	if err != nil {
+		errCh <- errors.Wrap(err, "MonitorConnections failed")
+		return
+	}
+
+	close(errCh)
+
+	for {
+		event, err := monitorClient.Recv()
+		if err != nil {
```
denis-tingaikin, line 146: Please verify that we don't get 'EOF'. `EOF` is a valid error for streams.

pkg/networkservice/common/heal/client.go
```diff
@@ -45,15 +57,174 @@ func NewClient(ctx context.Context, cc networkservice.MonitorConnectionClient) n
...
+				info.mut.Lock()
+				defer info.mut.Unlock()
+				heal(info.conn, true)
+				return true
+			})
+			return
+		}
+
+		for _, eventConn := range event.GetConnections() {
+			eventConn := eventConn
```
denis-tingaikin, line 157:
```suggestion
```

pkg/networkservice/common/heal/client.go
```diff
@@ -45,15 +57,174 @@ func NewClient(ctx context.Context, cc networkservice.MonitorConnectionClient) n
...
+}
+
+// confirmConnectionSuccess - waits for monitor thread to receive event confirming successful connection to next node
+func (u *healClient) confirmConnectionSuccess(conn *networkservice.Connection, successVerificationCh chan struct{}) error {
+	timeoutCh := time.After(time.Millisecond * 100)
+	select {
+	case <-successVerificationCh:
+		return nil
+	case <-timeoutCh:
+		return errors.Errorf("healClient: timeout expired but we couldn't verify that connection was established, connection id: %v", conn.GetId())
```
denis-tingaikin, line 190:
```suggestion
		return errors.Errorf("couldn't verify that connection: %v was established", conn.GetId())
```

pkg/networkservice/common/heal/client.go
```diff
@@ -45,15 +57,174 @@ func NewClient(ctx context.Context, cc networkservice.MonitorConnectionClient) n
...
+	}
+	info.mut.Lock()
+	defer info.mut.Unlock()
+	fun(info)
+}
+
+// applyLocked - searches the map for entry with key=id and runs provided function while entry mutex is locked.
+//               If map doesn't have entry with this key creates a new entry.
+//               Function is informed whether it receives new key or already existing key via first argument.
+func (m *connectionInfoMap) applyLockedOrNew(id string, fun func(created bool, info *connectionInfo)) {
```
denis-tingaikin, line 219: This is using only in one place. Please consider using this inline.

pkg/networkservice/common/heal/gen.go
```diff
@@ -19,11 +19,30 @@
...
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
 )
 
-//go:generate go-syncmap -output connection_map.gen.go -type connectionMap<string,connection>
-//go:generate go-syncmap -output client_connection_map.gen.go -type clientConnMap<string,clientConnInfo>
+type connectionInfo struct {
+	mut                   sync.Mutex
+	conn                  *networkservice.Connection
+	state                 connectionState
+	successVerificationCh chan struct{}
```
denis-tingaikin, line 32: Why we can't use `sync.Cond` as @Bolodya1997 suggested for simplifying the code?

pkg/networkservice/common/heal/gen.go
```diff
@@ -19,11 +19,30 @@
...
+	request *networkservice.NetworkServiceRequest
+	ctx     context.Context
+	cancel  func()
+}
+
+//go:generate go-syncmap -output context_wrapper_map.gen.go -type ctxWrapperMap<string,*ctxWrapper>
 
-type connectionMap sync.Map
-type clientConnMap sync.Map
+type ctxWrapperMap sync.Map
```
denis-tingaikin, line 48: Seems to me this is not using and can be removed.

denis-tingaikin (changes requested):

pkg/networkservice/common/heal/client.go
Bolodya1997: I suppose that receiving `EOF` in such case still means that we need to start healing :)

d-uzlov: What would EOF change here?
If I understand correctly, we should get EOF when connection to the next hop is closed because it is no longer needed, and in case this happens, connection map in this client should already be empty.
So if I add a check for EOF and just stop executing without requesting heal, this would change nothing.
Am I missing something?

pkg/networkservice/common/heal/client.go
d-uzlov: Without this line linter complains that I have unsafe usage of loop variable.

denis-tingaikin: OK, I see, linter is correct.

Then let's rework this to the index loop.

```suggestion
			eventConn := event.GetConnections()[i]
```

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,177 @@ package heal
...
+	connInfo.conn = conn.Clone()
+	if connInfo.state != connStateReady {
+		ch := make(chan struct{})
+		go func() {
+			connInfo.cond.Wait()
+			close(ch)
+		}()
+		select {
+		case <-ch:
+		case <-time.After(100 * time.Millisecond):
```
denis-tingaikin, line 119: Why 100 ms is enough?

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,177 @@ package heal
...
-		registerClient(u.ctx, conn, u.cc)
+	connInfo.cond.L.Lock()
+	defer connInfo.cond.L.Unlock()
+	connInfo.conn = conn.Clone()
+	if connInfo.state != connStateReady {
+		ch := make(chan struct{})
+		go func() {
+			connInfo.cond.Wait()
+			close(ch)
+		}()
```
denis-tingaikin, line 116: Oh, I see for cond we need a goroutine for the `select` statement. Probably the channel version is better.
@Bolodya1997 Thoughts?

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,177 @@ package heal
...
+		case <-ch:
+		case <-time.After(100 * time.Millisecond):
+			connInfo.cond.Broadcast()
+			<-ch
+			if connInfo.state == connStateInitial {
+				u.conns.Delete(conn.GetId())
+			}
+			_, _ = next.Client(ctx).Close(ctx, conn, opts...)
+			return nil, errors.Errorf("couldn't verify that connection: %v was established", conn.GetId())
+		}
```
denis-tingaikin, line 127: I think it is not a good solution to wait to sync with the monitoring server. Can we manage to miss of initial event in a separate goroutine and notify the server on timeout?

denis-tingaikin (changes requested):

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,177 @@ package heal
...
+		select {
+		case <-ch:
+		case <-time.After(100 * time.Millisecond):
+			connInfo.cond.Broadcast()
+			<-ch
+			if connInfo.state == connStateInitial {
+				u.conns.Delete(conn.GetId())
+			}
+			_, _ = next.Client(ctx).Close(ctx, conn, opts...)
+			return nil, errors.Errorf("couldn't verify that connection: %v was established", conn.GetId())
```
denis-tingaikin, line 126: Finally, by my local investigation, I found that we are using the monitor because we incorrectly handling grpc connection death.

denis-tingaikin (changes requested):
Please add keepalive.`ClientParameters` into https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/connect/client.go#L62 
https://pkg.go.dev/google.golang.org/grpc#WithKeepaliveParams

denis-tingaikin:
@edwarnicke , @Bolodya1997 Could you review this PR?

denis-tingaikin (approved):
In general, It looks good to me.

edwarnicke (approved):

pkg/networkservice/common/heal/client.go
```diff
@@ -19,41 +19,150 @@ package heal
...
-		registerClient(u.ctx, conn, u.cc)
+	connInfo, _ := u.conns.LoadOrStore(conn.GetId(), &connectionInfo{
+		conn: conn.Clone(),
+		mut:  sync.Mutex{},
+	})
+	u.replaceConnectionPath(conn, connInfo)
+
+	conn, err := next.Client(ctx).Request(ctx, request, opts...)
+	if err != nil {
+		return nil, err
```
Bolodya1997, line 97: If it is a first request for the connection, we need to remove `connInfo` from `u.conns`.

pkg/networkservice/common/heal/server.go
```diff
@@ -409,7 +227,7 @@ func (f *healServer) processHeal(ctx context.Context, request *networkservice.Ne
 		}
 	} else {
 		// Huge timeout is not required to close connection on a current path segment
-		closeCtx, closeCancel := clockTime.WithTimeout(ctx, time.Second)
+		closeCtx, closeCancel := context.WithTimeout(ctx, time.Second)
```
Bolodya1997, line 230: Please don't use direct `context.WithTimeout`, `.WithDeadline` and `time.` methods. Use tools from `pkg/tools/clock`.

Bolodya1997 (approved):

PR #838, opened 2021-04-16
Skip heal restore tests

Bolodya1997:
## Issue
https://github.com/networkservicemesh/sdk/issues/840

### DIFF
--- a/pkg/networkservice/chains/nsmgr/server_heal_test.go
+++ b/pkg/networkservice/chains/nsmgr/server_heal_test.go
@@ -275,8 +275,6 @@ func testNSMGRHealForwarder(t *testing.T, nodeNum int, restored bool, customConf
 }
 
 func TestNSMGR_HealRemoteNSMgrRestored(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/networkservice/chains/nsmgr/server_heal_test.go
```diff
@@ -368,7 +374,7 @@ func testNSMGRHealNSMgr(t *testing.T, nodeNum int, customConfig []*sandbox.NodeC
 	e, err := nsc.Close(ctx, conn)
 	require.NoError(t, err)
 	require.NotNil(t, e)
-	require.Equal(t, int32(2), atomic.LoadInt32(&counter.Requests))
```
denis-tingaikin, line 371: @Bolodya1997 Why this is changed?

Bolodya1997: There was a bug in code accidentally made by Artem. Initially there was a test without refresh request, after it was skipped due to the https://github.com/networkservicemesh/sdk/issues/770 and after there were added refresh request (but the test has already been skipped and so this count was not updated).

Issue #734, opened 2021-02-24
Refactor chains/registryproxy to use Option pattern


denis-tingaikin:
## Motivation

We can simplify using of our chains with
[Option pattern](https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md)

See at an example: https://github.com/networkservicemesh/sdk/commit/9007c4530373d301993ecd8985a1c9abebe359b2

Moreover, this can super help us with minimization of manual changes of upstream repositories.

Entry point: https://github.com/networkservicemesh/sdk/blob/master/pkg/registry/chains/proxydns/server.go

PR #852, opened 2021-04-20
Refactor chains/registryproxy to use options

glazychev-art:
#734 


denis-tingaikin:
We discussed with @edwarnicke about a modification of the interdomain protocol. Now interdomain case Path should include registry-proxy, nsmgr-proxy from the remote side. It means that chains/registryproxy will be reworked in https://github.com/networkservicemesh/sdk/issues/854

denis-tingaikin:
Will apply this patch in https://github.com/networkservicemesh/sdk/issues/854

Issue #764, opened 2021-03-10
Use clock library to enhance testing


Bolodya1997:
# Issue
We are running all our unit tests on multiple platforms with `--race` flag. It does mean that any timeout test becomes either slow or not stable or both, because we have totally no idea how long could it take for something to happen.

# Solution
Instead of using timeouts in tests we can use some clock library (e.g. https://github.com/benbjohnson/clock) to use mocked time in tests and real time in applications.

denis-tingaikin:
@edwarnicke Can we go with this? 

This can speed up our unit tests. This problem is most actual for https://github.com/networkservicemesh/sdk/pull/650 where we have long tests to check functionality based on timers.

edwarnicke:
@denis-tingaikin I had a long talk with @Bolodya1997 about this... and am fully in support :)

Bolodya1997:
Linked PRs:
* add `clock` package - https://github.com/networkservicemesh/sdk/pull/782
* rework `pkg/networkservice` to clock - https://github.com/networkservicemesh/sdk/pull/783
* rework `pkg/registry` to clock - https://github.com/networkservicemesh/sdk/pull/784
* rework `sandbox` to clock - https://github.com/networkservicemesh/sdk/pull/785

Bolodya1997:
Blocker for https://github.com/networkservicemesh/sdk/issues/823.

edwarnicke:
@Bolodya1997 Is this still in progress?

Bolodya1997:
It is actually done (#785) but blocked by #811. Just need to merge dependent PR and rebase.

PR #862, opened 2021-04-23
Edit clockmock.Mock to slowly run time all the time

Bolodya1997:
#764

## Issue
If code uses timer added with 0 duration, it will never fire in test with `clockmock.Mock` until time will be modified.

## Solution
Make time in `clockmock.Mock` always running but slowed down at 100000 times.

Closes https://github.com/networkservicemesh/sdk/issues/865.

### DIFF
--- a/pkg/tools/clockmock/mock.go
+++ b/pkg/tools/clockmock/mock.go
@@ -36,10 +36,25 @@ type Mock struct {
 }
 
 // NewMock returns a new mocked clock
-func NewMock() *Mock {
-	return &Mock{
+func NewMock(ctx context.Context) *Mock {
+	m := &Mock{
 		mock: libclock.NewMock(),
 	}
+
+	// Timers added with zero or negative duration will never run if time stands still. So mocked time
+	// should be slowed down (~100000 times) but never frozen.
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(100 * time.Microsecond):
+				m.Add(time.Nanosecond)
+			}
+		}
+	}()
+
+	return m
 }
 
 // Set sets the current time of the mock clock to a specific one.
--- a/pkg/tools/clockmock/mock.go
+++ b/pkg/tools/clockmock/mock.go
@@ -35,30 +35,14 @@ type Mock struct {
 	mock *libclock.Mock
 }
 
-// NewMock returns a new mocked clock
-func NewMock(ctx context.Context) *Mock {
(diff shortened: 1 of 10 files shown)


Bolodya1997:
Tests in NS were incorrect - waiting with `updateServer` was not enough to guarantee that NSEs have been added to the NS expire server. Replaced with `time.Sleep`.

denis-tingaikin:
> If code uses timer added with 0 duration, it will never fire in test with clockmock.Mock until time will be modified.

@Bolodya1997 Could you explain where we add 0 duration for timers?

Bolodya1997:
> > If code uses timer added with 0 duration, it will never fire in test with clockmock.Mock until time will be modified.
> 
> @Bolodya1997 Could you explain where we add 0 duration for timers?

In #772 I use `timer.Reset(0)` to force run expiration in expire manager.

pkg/tools/clockmock/timer.go
```diff
@@ -31,5 +33,12 @@ func (t *mockTimer) C() <-chan time.Time {
 }
 
 func (t *mockTimer) Reset(d time.Duration) bool {
-	return t.Timer.Reset(safeDuration(d))
+	if d = safeDuration(d); d == 0 {
```
denis-tingaikin, line 36:
```suggestion
	if  safeDuration(d) == 0 {
```

pkg/tools/clockmock/mock.go
```diff
@@ -82,11 +80,19 @@ func (m *Mock) Sleep(d time.Duration) {
 
 // Timer returns a timer that will fire when the mock current time becomes > m.Now().Add(d)
 func (m *Mock) Timer(d time.Duration) clock.Timer {
-	m.lock.RLock()
-	defer m.lock.RUnlock()
+	if d = safeDuration(d); d > 0 {
```
denis-tingaikin, line 83:
```suggestion
	if safeDuration(d) > 0 {
```

denis-tingaikin (commented):
Added minor comments. In general looks better than previous version :)

pkg/tools/clockmock/timer.go
Bolodya1997: It is used in line 43.

pkg/tools/clockmock/mock.go
Bolodya1997: It is used in line 95.

denis-tingaikin (approved):

PR #868, opened 2021-04-26
Add defaultLogger and use it by default

### LINKED ISSUE #758: Use go 1.16 log.Default by default in sdk/pkg/tools/log
https://golang.org/pkg/log/#Default

Bolodya1997:
Closes #758.


denis-tingaikin (changes requested):
If passed empty logger then nothing should be printed.
Please consider to create defaultLogger and use it if logger is missed in context.

denis-tingaikin (approved):

Issue #870, opened 2021-04-26
Common NSM scenarios are unstable in sandbox tests


Bolodya1997:
## Expected Behavior

All sandbox tests should work correctly.

## Current Behavior

Random sandbox tests failing from times to times.

## Steps to Reproduce

1. Run sandbox test on CI multiple times.

## Failure Logs

[5_Test.txt](https://github.com/networkservicemesh/sdk/files/6375052/5_Test.txt)

Bolodya1997:
Blocker for https://github.com/networkservicemesh/sdk/issues/823.

PR #874, opened 2021-04-27
Fix data race in NSMgr tests

Bolodya1997:
Issue #870.


pkg/networkservice/chains/nsmgr/nsmgr_test.go
```diff
@@ -93,7 +93,11 @@ func (s *nsmgrSuite) Test_Remote_ParallelUsecase() {
 	request := defaultRequest()
 	counter := &counterServer{}
 
+	var unregisterWG sync.WaitGroup
```
denis-tingaikin, line 96: Why don't we use `unregisterED.Wait`?

Bolodya1997: Oops, fixed.

denis-tingaikin (approved):

PR #884, opened 2021-04-28
Remove ns expire

### LINKED ISSUE #877: NS expire chain element logic is invalid
## Intro
Currently NS expire implements the following logic for single NS:
```
0. start listening for NSEs updates
1. register NS
2. receive update for NSE implementing NS
   * not expired - create a timer with NSE expiration time, store it to the list
   * expired - delete NSE timer from the list
      * list is empty - delete NS
```
Every NSE implementing NS added after NS registration owns it. When last NSE owning NS is removed, NS is removed.

Common scenario for running NSEs currently is:
```
1.   start NSE
2.   register NS
3.   register NSE
4-n. refresh NSE registration
```
If there are multiple NSEs implementing same NS, (2) does nothing for all of them except first.

## Issue
If we have 2 endpoints implementing same service and one of them starts after the previous death/unregistration there can be such case:
```
1.  (NSE-1) starts
2.  (NSE-1) registers NS
3.  (NSE-1) registers NSE-1
4.  (NSE-1) dies
5.  (NSE-2) starts
6.  (NSE-2) registers NS
7.          expiration time for NSE-1
8.          NSE-1 unregisters
9.          NS unregisters
10. (NSE-2) registers NSE-2
11. (NSC)   starts
12. (NSC)   requests NS
13. (NSC)   ...
14. (NSC)   timeout for the request
```

## Problem
1. We have different entities - NS, NSE. NSE implements NS, but doesn't own it. There possibly can be a case when there is no single NSE fully implementing NS (composite NS).
2. It is OK if we have NS that is not implemented by any NSE - in some future there can possibly appear some NSE implementing this NS. For the NSC perspective both cases (NS without NSEs, no NS) are the same - request should fail on timeout with _no NS_ error.

It means that currently existing NS expire chain element is invalid.

### Don't expire NS
The one who registers NS should also unregister it. If we have not used NS's, they can be removed by NSM admin manually.

### Add expiration time for NS
Add expiration time for NS registration and use refresh logic as for NSEs. NS's added as CRDs by k8s api wouldn't be expired in such case.

Bolodya1997:
Closes #877.


pkg/networkservice/chains/nsmgr/utils_test.go
```diff
@@ -0,0 +1,163 @@
...
+	kernelmech "github.com/networkservicemesh/api/pkg/api/networkservice/mechanisms/kernel"
+	"github.com/networkservicemesh/api/pkg/api/registry"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/sandbox"
+)
+
+func defaultRegistryService() *registry.NetworkService {
+	return &registry.NetworkService{
+		Name: "ns-1",
```
denis-tingaikin, line 40:
```suggestion
		Name: fmt.Sprintf("ns-%v", uuid.New().String())
```

pkg/networkservice/chains/nsmgr/utils_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+func defaultRegistryEndpoint() *registry.NetworkServiceEndpoint {
```
denis-tingaikin, line 44:
```suggestion
func defaultRegistryEndpoint(ns string) *registry.NetworkServiceEndpoint {
```

pkg/networkservice/chains/nsmgr/utils_test.go
```diff
@@ -0,0 +1,163 @@
...
+	return &registry.NetworkServiceEndpoint{
+		Name:                "final-endpoint",
+		NetworkServiceNames: []string{"ns-1"},
+	}
+}
+
+func defaultRequest() *networkservice.NetworkServiceRequest {
```
denis-tingaikin, line 51:
```suggestion
func defaultRequest(ns string) *networkservice.NetworkServiceRequest {
```

pkg/networkservice/chains/nsmgr/single_test.go
```diff
@@ -0,0 +1,262 @@
+// Copyright (c) 2020-2021 Doc.ai and/or its affiliates.
```
denis-tingaikin, line 1: Seems to me these changes are not related to the issue.

pkg/networkservice/chains/nsmgr/utils_test.go
Bolodya1997: Yep, done.

pkg/networkservice/chains/nsmgr/utils_test.go
Bolodya1997: Yep, done.

pkg/networkservice/chains/nsmgr/utils_test.go
Bolodya1997: Yep, done.

pkg/networkservice/chains/nsmgr/single_test.go
Bolodya1997: Moved to another PR - https://github.com/networkservicemesh/sdk/pull/885.

denis-tingaikin (approved):
Nice!

Issue #848, opened 2021-04-19
Test_WatchFile is unstable


Bolodya1997:
## Expected Behavior

`Test_WatchFile` should not fail.

## Current Behavior

`Test_WatchFile` fails sometimes on windows.

## Steps to Reproduce

1. Create a PR to `sdk` to start automation.
2. Restart automation several times.

## Failure Logs

https://github.com/networkservicemesh/sdk/runs/2378604509?check_suite_focus=true#step:5:114
```
--- FAIL: Test_WatchFile (0.01s)
    utils_test.go:75: 
        	Error Trace:	utils_test.go:75
        	Error:      	Expected nil, but got: &fs.PathError{Op:"open", Path:"C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\Test_WatchFile\\A\\file1.txt", Err:0x3}
        	Test:       	Test_WatchFile
FAIL
```

PR #886, opened 2021-04-28
Fix Test_WatchFile instability

d-uzlov:
fixes https://github.com/networkservicemesh/sdk/issues/848

### DIFF
--- a/pkg/tools/fs/utils_test.go
+++ b/pkg/tools/fs/utils_test.go
@@ -18,7 +18,6 @@ package fs_test
 
 import (
 	"context"
-	iofs "io/fs"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -36,16 +35,6 @@ import (
 
 const macOSName = "darwin"
 
-func checkPathError(t *testing.T, err error) error {
-	if fErr, ok := err.(*iofs.PathError); ok {
-		if u := fErr.Unwrap(); u != nil {
-			t.Log(u) // Without unwrapping printing gives only numeric error code.
-		}
-	}
-
-	return err
-}
-
 func Test_WatchFile(t *testing.T) {
 	t.Cleanup(func() { goleak.VerifyNone(t) })
 
@@ -56,7 +45,7 @@ func Test_WatchFile(t *testing.T) {
 	defer func() {
 		_ = os.RemoveAll(path)
 	}()
-	require.NoError(t, checkPathError(t, err))
+	require.NoError(t, err)
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
@@ -78,7 +67,7 @@ func Test_WatchFile(t *testing.T) {
 	require.Nil(t, expectEvent(), filePath) // initial file read, nil because file doesn't exist
 
 	err = ioutil.WriteFile(filePath, []byte("data"), os.ModePerm)
-	require.NoError(t, checkPathError(t, err))
+	require.NoError(t, err)
 	require.NotNil(t, expectEvent(), filePath) // file created
 
 	// https://github.com/fsnotify/fsnotify/issues/11
@@ -87,7 +76,7 @@ func Test_WatchFile(t *testing.T) {
 	}
 
 	err = os.RemoveAll(path)
-	require.NoError(t, checkPathError(t, err))
+	require.NoError(t, err)
 	if runtime.GOOS != macOSName {
 		require.Nil(t, expectEvent(), filePath) // file removed
 	} else {
@@ -98,18 +87,17 @@ func Test_WatchFile(t *testing.T) {
 		require.Nil(t, event, filePath)
 	}
 
-	err = checkPathError(t, os.MkdirAll(path, os.ModePerm))
-	if err != nil {
+	if os.MkdirAll(path, os.ModePerm) != nil {
 		// Removing file is async operation.
 		// Waiting for events should theoretically sync us with the filesystem,
 		// but apparently sometimes it's not enough, so MkdirAll can fail because the folder is still locked by the remove operation.
 		// Particularly, this can be observed on slow Windows systems.
 		require.Eventually(t, func() bool {
-			return checkPathError(t, os.MkdirAll(path, os.ModePerm)) == nil
+			return os.MkdirAll(path, os.ModePerm) == nil
 		}, time.Millisecond*300, time.Millisecond*50)
 	}
 
 	err = ioutil.WriteFile(filePath, []byte("data"), os.ModePerm)
-	require.NoError(t, checkPathError(t, err))
+	require.NoError(t, err)
 	require.NotNil(t, expectEvent(), filePath) // file created
 }
--- a/pkg/tools/fs/utils_test.go
+++ b/pkg/tools/fs/utils_test.go
@@ -32,8 +32,6 @@ import (
 )
 
 func Test_WatchFile(t *testing.T) {
-	t.Skip("https://github.com/networkservicemesh/sdk/issues/848")
-
 	root := filepath.Join(os.TempDir(), t.Name())
 	defer func() {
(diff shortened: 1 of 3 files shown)


pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,88 @@ import (
...
+	require.Nil(t, expectEvent(), filePath) // initial file read, nil because file doesn't exist
 
-	require.Nil(t, ioutil.WriteFile(filePath, []byte("data"), os.ModePerm))
-
-	expectEvent(require.NotNil)
+	err = ioutil.WriteFile(filePath, []byte("data"), os.ModePerm)
+	checkPathError(t, err)
+	require.NotNil(t, expectEvent(), filePath) // file created
 
+	macOSName := "darwin"
```
Bolodya1997, line 82: Probably this could be done as a constant out of test.

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,88 @@ import (
...
+		event := expectEvent()
+		if event != nil { // ...but sometimes we still get write events on MacOS. Very rarely, about 1/500 test retries
+			event = expectEvent()
+		}
+		require.Nil(t, event, filePath)
+	}
 
-	_ = os.MkdirAll(path, os.ModePerm)
-	require.Nil(t, ioutil.WriteFile(filePath, []byte("data"), os.ModePerm))
+	err = os.MkdirAll(path, os.ModePerm)
```
Bolodya1997, line 100: If you modify `checkPathError` to simply log and return an error instead of making `require` stuff, you can replace this loop with:
```go
require.Eventually(t, func() {
    return checkPathError(os.MkdirAll(path, os.ModePerm)) == nil
}, timeout, tick)
```

d-uzlov: `require.Eventually` waits `tick` time before first condition check, which makes the test unnecessarily slower on average, so we can't reduce this code to only `require.Eventually`.
But it will probably be more readable than this raw loop, so I replaced it.

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,89 @@ import (
...
+		if event != nil { // ...but sometimes we still get write events on MacOS. Very rarely, about 1/500 test retries
+			event = expectEvent()
+		}
+		require.Nil(t, event, filePath)
+	}
 
-	_ = os.MkdirAll(path, os.ModePerm)
-	require.Nil(t, ioutil.WriteFile(filePath, []byte("data"), os.ModePerm))
+	err = checkPathError(t, os.MkdirAll(path, os.ModePerm))
+	if err != nil {
```
Bolodya1997, line 102: Why do we need this extra check before `require.Eventually`?

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,89 @@ import (
...
 	}
 
-	expectEvent(require.Nil)
-
-	require.Nil(t, ioutil.WriteFile(filePath, []byte("data"), os.ModePerm))
+	require.Nil(t, expectEvent(), filePath) // initial file read, nil because file doesn't exist
 
-	expectEvent(require.NotNil)
+	err = ioutil.WriteFile(filePath, []byte("data"), os.ModePerm)
+	require.Nil(t, checkPathError(t, err))
```
Bolodya1997, line 81:
```suggestion
	require.NoError(t, checkPathError(t, err))
```

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,89 @@ import (
...
+	return err
+}
+
 func Test_WatchFile(t *testing.T) {
-	t.Skip("https://github.com/networkservicemesh/sdk/issues/848")
+	t.Cleanup(func() { goleak.VerifyNone(t) })
 
 	root := filepath.Join(os.TempDir(), t.Name())
+
+	path := filepath.Join(root, uuid.New().String())
```
denis-tingaikin, line 54: Please don't use uuid for paths in tests. Look at the previous comment.

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,89 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
```
denis-tingaikin, line 30:
```suggestion
```

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,89 @@ import (
...
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
 	filePath := filepath.Join(path, "file1.txt")
 	ch := fs.WatchFile(ctx, filePath)
 
-	expectEvent := func(assertFunc func(require.TestingT, interface{}, ...interface{})) {
+	expectEvent := func() []byte {
 		select {
 		case <-time.After(time.Second):
```
denis-tingaikin, line 69: Actually, this is not correct to use a timeout in `1s` for each event. Please let's set one timeout in `1s` for all events.
```suggestion
		case <-timeoutCh:
```

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,89 @@ import (
...
+	require.NoError(t, checkPathError(t, err))
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
 	filePath := filepath.Join(path, "file1.txt")
 	ch := fs.WatchFile(ctx, filePath)
 
-	expectEvent := func(assertFunc func(require.TestingT, interface{}, ...interface{})) {
+	expectEvent := func() []byte {
```
denis-tingaikin, line 67:
```suggestion
	readEvent := func() []byte {
```

pkg/tools/fs/utils_test.go
```diff
@@ -26,55 +27,89 @@ import (
...
+
+func checkPathError(t *testing.T, err error) error {
+	if fErr, ok := err.(*iofs.PathError); ok {
+		if u := fErr.Unwrap(); u != nil {
+			t.Log(u) // Without unwrapping printing gives only numeric error code.
+		}
+	}
+
+	return err
+}
```
denis-tingaikin, line 47: As I can see this function just logs PathError. 
Can we just use err.Error() to get information about the error?
```suggestion
```

denis-tingaikin (changes requested):

pkg/tools/fs/utils_test.go
```diff
@@ -21,60 +21,78 @@ import (
...
 
 	"github.com/networkservicemesh/sdk/pkg/tools/fs"
 )
 
 func Test_WatchFile(t *testing.T) {
-	t.Skip("https://github.com/networkservicemesh/sdk/issues/848")
+	t.Cleanup(func() { goleak.VerifyNone(t) })
 
-	root := filepath.Join(os.TempDir(), t.Name())
+	root := filepath.Join(t.TempDir(), t.Name())
```
denis-tingaikin, line 37:
```suggestion
	root := t.TempDir()
```

pkg/tools/fs/utils_test.go
```diff
@@ -21,60 +21,78 @@ import (
...
 	defer func() {
 		_ = os.RemoveAll(root)
 	}()
 
-	path := filepath.Join(root, "A")
-	_ = os.MkdirAll(path, os.ModePerm)
+	err := os.MkdirAll(root, os.ModePerm)
```
denis-tingaikin, line 42:
```suggestion
```

pkg/tools/fs/utils_test.go
```diff
@@ -21,60 +21,78 @@ import (
...
+	require.NoError(t, err)
```
denis-tingaikin, line 43:
```suggestion
```

denis-tingaikin (approved):
Looks good!

PR #897, opened 2021-05-03
fix: Improve condition for 'Update dependent repositories' job

denis-tingaikin:

## Description

Condition `github.event_name == 'push'` is more strict then `github.ref == 'refs/heads/main`

### DIFF
--- a/.github/workflows/update-dependent-repositories-gomod.yaml
+++ b/.github/workflows/update-dependent-repositories-gomod.yaml
@@ -28,7 +28,7 @@ jobs:
           - cmd-nsc-init
     name: Update ${{ matrix.repository }}
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' && github.actor == 'nsmbot' || github.ref == 'refs/heads/main' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.actor == 'nsmbot' || github.event_name == 'push' }}
     steps:
       - name: Checkout ${{ github.repository }}
         uses: actions/checkout@v2


.github/workflows/update-dependent-repositories-gomod.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
           - cmd-nsc-init
     name: Update ${{ matrix.repository }}
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' && github.actor == 'nsmbot' || github.ref == 'refs/heads/main' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.actor == 'nsmbot' || github.event_name == 'push' }}
```
edwarnicke, line 31: So... this looks like it could result in pushing out updates simply for incoming events to update/* branches... not just when they are merged to main...

denis-tingaikin: There we have two events then this workflow will run:

1. https://github.com/networkservicemesh/sdk/blob/main/.github/workflows/update-dependent-repositories-gomod.yaml#L4-L6 -- on pushing in main.
2. https://github.com/networkservicemesh/sdk/blob/main/.github/workflows/update-dependent-repositories-gomod.yaml#L7-L11 -- on automerge job compling.

This change is needed to avoid potential issues from events with type 2 that triggered on the main branch.

denis-tingaikin: > So... this looks like it could result in pushing out updates simply for incoming events to update/* branches... not just when they are merged to main...

For events with type 1  update/* branches should be fine. For them we have cond `${{ github.event.workflow_run.conclusion == 'success' && github.actor == 'nsmbot'`.

Issue #880, opened 2021-04-27
NSE Composition: Sandbox tests


Bolodya1997:
## Overview

1. Rework existing passthrough tests with composite NS. Add tests for complex tree scenarios.
2. Cover additional scenarios:
   1. Tree
      1. Add NS:
      ```yaml
      matches:
      - routes:
        - destination_selector:
            step: aa
      - source_selector:
          step: aa
        routes:
        - destination_selector:
            step: ab
      - source_selector:
          step: ab
        routes:
        - destination_selector:
            step: c
      - routes:
        - destination_selector:
            step: ba
      - source_selector:
          step: ba
        routes:
        - destination_selector:
            step: bb
      - source_selector:
          step: bb
        routes:
        - destination_selector:
            step: c
      - source_selector:
          step: c
        routes:
        - destination_selector:
            step: d
      ```
      2. Register `aa`, `ba`, `bb`, `c`, `d` Endpoints.
      3. Make and validate Request with a new Client : `aa` - 0, `ba` - 1, `bb` - 1, `c` - 1, `d` - 1.
      4. Register `ab` Endpoint.
      5. Make and validate Request with a new Client : `aa` - 1, `ab` - 1, `ba` - 1, `bb` - 1, `c` - 2, `d` - 2.
      6. Refresh and validate both Requests: `aa` - 2, `ab` - 2, `ba` - 2, `bb` - 2, `c` - 4, `d` - 4.
      7. Close both Requests.
   2. Multi label
      1. Add NS:
      ```yaml
      matches:
      - routes:
        - destination_selector:
            label_a: a
      - source_selector:
          label_a: a
          label_b: b
        routes:
        - destination_selector:
            label_a: aa
      - source_selector:
          label_a: aa
          label_b: bb
        routes:
        - destination_selector:
            label_a: aaa
            label_b: bbb
      ```
      2. Register Endpoints: `label_a: a, label_b: b`, `label_a: a, label_b: bb`, ..., `label_a: aaa, label_b: bbb` with 300 ms Request timeout for the client side
      3. Make and validate Request with a new Client: `label_a: a, label_b: b` -> `label_a: aa, label_b: bb` -> `label_a: aaa, label_b: bbb`

## References

* existing passthrough tests:
   * https://github.com/networkservicemesh/sdk/blob/5ab6206d4dcc62f09f79cc308aab1d5f6dc9202e/pkg/networkservice/chains/nsmgr/suite_test.go#L391
[...]

PR #923, opened 2021-05-14
Rework of sandbox passthrough tests for using matches

Mixaster995:

## Description
With this changes passthrough cases now using labels matching to discover endpoint chain, also added case for testing multi label scenario

## Issue link
fixes https://github.com/networkservicemesh/sdk/issues/880

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [x] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/nsmgr/suite_test.go
+++ b/pkg/networkservice/chains/nsmgr/suite_test.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/url"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -42,9 +43,9 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/connect"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/mechanisms/kernel"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/mechanismtranslation"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/common/replacelabels"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/chain"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/inject/injecterror"
-	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/inject/injectlabels"
 	"github.com/networkservicemesh/sdk/pkg/tools/sandbox"
 )
 
@@ -401,48 +402,17 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
 	nsReg, err := s.domain.Nodes[0].NSRegistryClient.Register(ctx, nsReg)
 	require.NoError(t, err)
 
-	var nsesReg [nodesCount]*registry.NetworkServiceEndpoint
+	var nsesReg []*registry.NetworkServiceEndpoint
 	for i := 0; i < nodesCount; i++ {
-		nsesReg[i] = &registry.NetworkServiceEndpoint{
-			Name:                fmt.Sprintf("endpoint-%v", i),
-			NetworkServiceNames: []string{nsReg.GetName()},
-			NetworkServiceLabels: map[string]*registry.NetworkServiceLabels{
-				nsReg.GetName(): {
-					Labels: map[string]string{
-						step: fmt.Sprintf("%v", i),
-					},
-				},
-			},
+		labels := map[string]string{
+			step: fmt.Sprintf("%v", i),
 		}
-
-		var additionalFunctionality []networkservice.NetworkServiceServer
-		if i != nodesCount-1 {
-			additionalFunctionality = []networkservice.NetworkServiceServer{
-				chain.NewNetworkServiceServer(
-					clienturl.NewServer(s.domain.Nodes[i].NSMgr.URL),
-					connect.NewServer(ctx,
-						client.NewClientFactory(
-							client.WithName(fmt.Sprintf("endpoint-client-%v", i)),
-							client.WithAdditionalFunctionality(
-								mechanismtranslation.NewClient(),
-								injectlabels.NewClient(nsesReg[i].NetworkServiceLabels[nsReg.Name].Labels),
-								kernel.NewClient(),
-							),
-						),
-						connect.WithDialTimeout(sandbox.DialTimeout),
-						connect.WithDialOptions(sandbox.DefaultDialOptions(sandbox.GenerateTestToken)...),
-					),
-				),
-			}
-		}
-
-		_, err = s.domain.Nodes[i].NewEndpoint(ctx, nsesReg[i], sandbox.GenerateTestToken, additionalFunctionality...)
-		require.NoError(t, err)
+		nsesReg = append(nsesReg, newPassThroughEndpoint(ctx, t, s, labels, fmt.Sprintf("%v", i), nsReg.Name, i != nodesCount-1, i))
 	}
 
 	nsc := s.domain.Nodes[0].NewClient(ctx, sandbox.GenerateTestToken)
 
-	request := defaultRequest(nsReg.GetName())
+	request := defaultRequest(nsReg.Name)
 
 	// Request
 	conn, err := nsc.Request(ctx, request)
@@ -452,6 +422,9 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
 	// Path length to first endpoint is 5
 	// Path length from NSE client to other remote endpoint is 8
 	require.Equal(t, 8*(nodesCount-1)+5, len(conn.Path.PathSegments))
+	for i := 0; i < len(nsesReg); i++ {
+		require.Contains(t, nsesReg[i].Name, conn.Path.PathSegments[i*8+4].Name)
+	}
 
 	// Refresh
 	conn, err = nsc.Request(ctx, request)
@@ -470,7 +443,7 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
 
 func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
 	t := s.T()
-	const nsesCount = 5
+	const nsesCount = 7
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
@@ -479,43 +452,13 @@ func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
 	nsReg, err := s.domain.Nodes[0].NSRegistryClient.Register(ctx, nsReg)
 	require.NoError(t, err)
 
-	var nsesReg [nsesCount]*registry.NetworkServiceEndpoint
+	var nsesReg []*registry.NetworkServiceEndpoint
 	for i := 0; i < nsesCount; i++ {
-		nsesReg[i] = &registry.NetworkServiceEndpoint{
-			Name:                fmt.Sprintf("endpoint-%v", i),
-			NetworkServiceNames: []string{nsReg.GetName()},
-			NetworkServiceLabels: map[string]*registry.NetworkServiceLabels{
-				nsReg.GetName(): {
-					Labels: map[string]string{
-						step: fmt.Sprintf("%v", i),
-					},
-				},
-			},
+		labels := map[string]string{
+			step: fmt.Sprintf("%v", i),
 		}
 
-		var additionalFunctionality []networkservice.NetworkServiceServer
-		if i != nsesCount-1 {
-			additionalFunctionality = []networkservice.NetworkServiceServer{
-				chain.NewNetworkServiceServer(
-					clienturl.NewServer(s.domain.Nodes[0].NSMgr.URL),
-					connect.NewServer(ctx,
-						client.NewClientFactory(
-							client.WithName(fmt.Sprintf("%v", i)),
-							client.WithAdditionalFunctionality(
-								mechanismtranslation.NewClient(),
-								injectlabels.NewClient(nsesReg[i].NetworkServiceLabels[nsReg.Name].Labels),
-								kernel.NewClient(),
-							),
-						),
-						connect.WithDialTimeout(sandbox.DialTimeout),
-						connect.WithDialOptions(sandbox.DefaultDialOptions(sandbox.GenerateTestToken)...),
-					),
-				),
-			}
-		}
-
-		_, err = s.domain.Nodes[0].NewEndpoint(ctx, nsesReg[i], sandbox.GenerateTestToken, additionalFunctionality...)
-		require.NoError(t, err)
+		nsesReg = append(nsesReg, newPassThroughEndpoint(ctx, t, s, labels, fmt.Sprintf("%v", i), nsReg.Name, i != nsesCount-1, 0))
 	}
 
 	nsc := s.domain.Nodes[0].NewClient(ctx, sandbox.GenerateTestToken)
@@ -529,6 +472,9 @@ func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
 	// Path length to first endpoint is 5
 	// Path length from NSE client to other local endpoint is 5
 	require.Equal(t, 5*(nsesCount-1)+5, len(conn.Path.PathSegments))
+	for i := 0; i < len(nsesReg); i++ {
+		require.Contains(t, nsesReg[i].Name, conn.Path.PathSegments[(i+1)*5-1].Name)
+	}
 
 	// Refresh
 	conn, err = nsc.Request(ctx, request)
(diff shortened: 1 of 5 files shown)


pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -394,42 +397,56 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
 
+	labels := []string{"a", "b", "c", "d", "e", "f", "g"}
```
Bolodya1997, line 400: I suppose that `nodesCount` or `nsesCount` is enough for the linear case, there is no need for the `labels` list.
You can simply create `linearNS(count)` with `step: i --> step: i+1` routes.
And so it is the same for the NSE registration loop, there is no need for `labels[i]` when you can simply use `i`.

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -394,42 +397,56 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
...
+				},
+			},
+		}
+
 		var additionalFunctionality []networkservice.NetworkServiceServer
-		if i != 0 {
-			// Passtrough to the node i-1
+		if i != nodesCount-1 {
 			additionalFunctionality = []networkservice.NetworkServiceServer{
 				chain.NewNetworkServiceServer(
```
d-uzlov, line 423: Seems like all test methods use the same code for generating the endpoint.
I suggest extracting a separate function for this.

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -451,53 +472,147 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
...
+	// Endpoint unregister
+	for i := 0; i < len(nsesReg); i++ {
+		_, err = s.domain.Nodes[0].EndpointRegistryClient.Unregister(ctx, nsesReg[i])
 		require.NoError(t, err)
+	}
+}
 
+func (s *nsmgrSuite) Test_PassThroughLocalUsecaseMultiLabel() {
+	t := s.T()
+	const nsesCount = 9
```
d-uzlov, line 554: When you are using arrays of labels for endpoins, you can remove `nsesCount` and use `len(labels)` instead.

Mixaster995: it is used for array 'nsesReg' for initialization

pkg/networkservice/chains/nsmgr/suite_test.go
Mixaster995: it is exactly the same only in 2 tests, i think function is not necessary

d-uzlov: You have 3 passthrough test functions, and all of which create their servers the same way.
Even if we don't add more tests, this is too much duplicated code in my opinion.

pkg/networkservice/chains/nsmgr/suite_test.go
d-uzlov: You can create that array dynamically.
Main point is to not being forced to manually sync `nsesCount` value each time you adjust labels array.

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -394,42 +397,56 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
...
+					},
+				},
+			},
+		}
+
 		var additionalFunctionality []networkservice.NetworkServiceServer
-		if i != 0 {
-			// Passtrough to the node i-1
+		if i != nodesCount-1 {
 			additionalFunctionality = []networkservice.NetworkServiceServer{
```
d-uzlov, line 420: I don't know why original tests had an array of one element here, but we don't need this array here.
Please, replace it with one `networkservice.NetworkServiceServer element`.

Mixaster995: because in case of single server it would be nil if condition not met

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -451,53 +470,145 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
+	defer cancel()
+
+	nsReg := multiLabelNS()
+	nsReg, err := s.domain.Nodes[0].NSRegistryClient.Register(ctx, nsReg)
+	require.NoError(t, err)
+
+	var nsesReg [nsesCount]*registry.NetworkServiceEndpoint
+	for i := 0; i < nsesCount; i++ {
```
Bolodya1997, line 562: We can use here such loop instead of this:
```go
var labelA, labelB string
for i := 0; i < 3; i++ {
    labelA += "a"
    for k := 0; k < 3; k++ {
        labelB += "b"
        labels := map[string]string{
            "label_a": labelA,
            "label_b": labelB,
        }
        ...
        if i != 2 || k != 2 { // let expect any "aaa" or "bbb" endpoint not to be passthrough
            ...
        }
    }
}
```
So we don't actually need either `nsesCount` or `multiLabelEndpoints()`.

pkg/networkservice/chains/nsmgr/suite_test.go
Bolodya1997: I agree, you can add some `passthroughFunctionality(...)` method to be used in all passthrough cases.

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -438,6 +453,10 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
 	// Path length from NSE client to other remote endpoint is 8
 	require.Equal(t, 8*(nodesCount-1)+5, len(conn.Path.PathSegments))
```
Bolodya1997, line 454: Probably it would be better to check that we have visited all required endpoints only once in the required order.

pkg/networkservice/chains/nsmgr/suite_test.go
Mixaster995: ok, done

pkg/networkservice/chains/nsmgr/suite_test.go
Mixaster995: done

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -394,42 +397,54 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
...
 
+	nsReg := linearNS(nodesCount)
+	nsReg, err := s.domain.Nodes[0].NSRegistryClient.Register(ctx, nsReg)
+	require.NoError(t, err)
+
 	var nsesReg [nodesCount]*registry.NetworkServiceEndpoint
 	for i := 0; i < nodesCount; i++ {
+		nsesReg[i] = &registry.NetworkServiceEndpoint{
+			Name:                fmt.Sprintf("endpoint-%v", i),
+			NetworkServiceNames: []string{nsReg.GetName()},
```
Bolodya1997, line 408: Probably just `.Name` is more common for this in tests.

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -451,45 +470,57 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
 
 func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
 	t := s.T()
-	const nsesCount = 7
+	const nsesCount = 5
```
Bolodya1997, line 473: Do we need this count for any reason to be changed?

Mixaster995: not needed after review fixes, returned old value

pkg/networkservice/chains/nsmgr/suite_test.go
Mixaster995: done

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -394,49 +398,57 @@ func (s *nsmgrSuite) Test_PassThroughRemoteUsecase() {
...
+	// Request
 	conn, err := nsc.Request(ctx, request)
 	require.NoError(t, err)
 	require.NotNil(t, conn)
 
 	// Path length to first endpoint is 5
 	// Path length from NSE client to other remote endpoint is 8
 	require.Equal(t, 8*(nodesCount-1)+5, len(conn.Path.PathSegments))
+	for i := 0; i < len(nsReg.Matches); i++ {
+		if i == 0 {
```
Bolodya1997, line 442: I guess that there is no need for the additional _if 0_ case :)
(0 + 1) * 5 - 1 = 4
0 * 8 + 4 = 4

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -498,6 +507,101 @@ func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
...
+							labelA: labelAvalue,
+							labelB: labelBvalue,
+						},
+					},
+				},
+			}
+
+			var additionalFunctionality []networkservice.NetworkServiceServer
+			if i != 2 || j != 2 {
+				additionalFunctionality = []networkservice.NetworkServiceServer{
```
Bolodya1997, line 561: This can also be replaced with `additionalFunctionalityChain`.

pkg/networkservice/utils/inject/injectlabels/client.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (s *injectLabelsClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (conn *networkservice.Connection, err error) {
+	oldConnLabels := request.Connection.Labels
+	oldNetworkServiceEndpointName := request.Connection.NetworkServiceEndpointName
+
+	request.Connection.Labels = s.labels
+	request.Connection.NetworkServiceEndpointName = ""
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
```
Bolodya1997, line 43: Please create some testing `requesttimeout` chain element instead of this.

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -528,3 +725,201 @@ func (s *nsmgrSuite) Test_ShouldCleanAllClientAndEndpointGoroutines() {
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func additionalFunctionalityChain(ctx context.Context, clientURL *url.URL, clientName string, labels map[string]string, empty bool) []networkservice.NetworkServiceServer {
+	if empty {
```
Bolodya1997, line 904: Probably it would be better to do this check outside of the method and just not call it in such case.

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -498,6 +497,204 @@ func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
...
+	firstExpectedPath := []string{"ba", "bb", "c", "d"}
+	secondExpectedPath := []string{"aa", "ab", "c", "d"}
+
+	nsReg := treeNS()
+	nsReg, err := s.domain.Nodes[0].NSRegistryClient.Register(ctx, nsReg)
+	require.NoError(t, err)
+
+	var nsesReg []*registry.NetworkServiceEndpoint
+	for i := 0; i < len(labels); i++ {
+		nseReg := &registry.NetworkServiceEndpoint{
```
Bolodya1997, line 536: Please create a separate method `newPassthroughEndpoint` for this:
```go
nseReg := &registry.NetworkServiceEndpoint{
	Name:                fmt.Sprintf("endpoint-%v", labels[i]),
	NetworkServiceNames: []string{nsReg.Name},
	NetworkServiceLabels: map[string]*registry.NetworkServiceLabels{
		nsReg.Name: {
			Labels: map[string]string{
				step: labels[i],
			},
		},
	},
}

additionalFunctionality := additionalFunctionalityChain(ctx, s.domain.Nodes[0].NSMgr.URL,
	fmt.Sprintf("endpoint-client-%v", labels[i]), nseReg.NetworkServiceLabels[nsReg.Name].Labels, i == len(labels)-1)

_, err = s.domain.Nodes[0].NewEndpoint(ctx, nseReg, sandbox.GenerateTestToken, additionalFunctionality...)
require.NoError(t, err)
```

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -498,6 +497,204 @@ func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
...
+	labels := []string{"aa", "ba", "bb", "c", "d"}
+	firstExpectedPath := []string{"ba", "bb", "c", "d"}
+	secondExpectedPath := []string{"aa", "ab", "c", "d"}
+
+	nsReg := treeNS()
+	nsReg, err := s.domain.Nodes[0].NSRegistryClient.Register(ctx, nsReg)
+	require.NoError(t, err)
+
+	var nsesReg []*registry.NetworkServiceEndpoint
+	for i := 0; i < len(labels); i++ {
```
Bolodya1997, line 535: I don't think that we actually need `len(labels)` here. Such thing probably should be better:
```go
for _, labelValue := range labels {
	...
	if labelValue != "d" { // or you can create some "lastLabel" const
		...
	}
}
```

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -498,6 +497,204 @@ func (s *nsmgrSuite) Test_PassThroughLocalUsecase() {
...
+
+	request := defaultRequest(nsReg.Name)
+
+	conn, err := nsc.Request(ctx, request)
+	require.NoError(t, err)
+	require.NotNil(t, conn)
+
+	// Path length to first endpoint is 5
+	// Path length from NSE client to other local endpoint is 5
+	require.Equal(t, 5*(len(labels)-1), len(conn.Path.PathSegments))
```
Bolodya1997, line 567: So here the same, please don't use `len(labels)` in a tree case - here should be used `len(firstExpectedPath)`.

Mixaster995: ok, done

pkg/networkservice/chains/nsmgr/suite_test.go
Mixaster995: done

pkg/networkservice/chains/nsmgr/suite_test.go
Mixaster995: done

pkg/networkservice/chains/nsmgr/suite_test.go
Mixaster995: done

pkg/networkservice/common/discover/context.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Cisco Systems, Inc.
+// Copyright (c) 2020-2021 Cisco Systems, Inc.
```
d-uzlov, line 1:
```suggestion
// Copyright (c) 2020 Cisco Systems, Inc.
// Copyright (c) 2021 Doc.ai and/or its affiliates.
```

pkg/networkservice/common/discover/server_test.go
```diff
@@ -187,7 +187,7 @@ func TestMatchEmptySourceSelector(t *testing.T) {
 		checkcontext.NewServer(t, func(t *testing.T, ctx context.Context) {
 			nses := discover.Candidates(ctx).Endpoints
 			require.Len(t, nses, 1)
-			require.Equal(t, want, nses[0].NetworkServiceLabels)
+			require.Equal(t, want, nses[0][0].NetworkServiceLabels)
```
d-uzlov, line 190: Please, add additional check: `require.Len(t, nses[0], 1)`
Same for other tests with such checks.

pkg/networkservice/common/roundrobin/server.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Cisco Systems, Inc.
+// Copyright (c) 2020-2021 Cisco Systems, Inc.
```
d-uzlov, line 1: Same as above, correct copyright it to Doc.ai.

pkg/networkservice/common/roundrobin/server.go
```diff
@@ -50,19 +50,21 @@ func (s *selectEndpointServer) Request(ctx context.Context, request *networkserv
...
-		if err != nil {
-			return nil, errors.WithStack(err)
-		}
-		ctx = clienturlctx.WithClientURL(ctx, u)
-		request.GetConnection().NetworkServiceEndpointName = endpoint.Name
-		resp, err := next.Server(ctx).Request(ctx, request)
-		if err == nil {
-			return resp, err
+		for j := 0; j < len(candidates.Endpoints[i]); j++ {
+			endpoint := s.selector.selectEndpoint(candidates.NetworkService, candidates.Endpoints[i])
```
d-uzlov, line 54: I'm not sure how critical this is, but this will now work correctly.
`s.selector` holds an index of endpoint for this network service. This was _probably_ fine when we had an array of candidates for NS, but now we will have a shared index for all of the sub-arrays.
It would probably be better to modify the selector to handle double array instead of making double loop here.

d-uzlov: Though, I didn't research this, but I feel like the selector already had potential issues when we had network services defined with matches.

pkg/networkservice/utils/inject/injectlabels/client.go
```diff
@@ -0,0 +1,62 @@
...
+
+type injectLabelsClient struct {
+	labels map[string]string
+}
+
+func (s *injectLabelsClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (conn *networkservice.Connection, err error) {
+	oldConnLabels := request.Connection.Labels
+	oldNetworkServiceEndpointName := request.Connection.NetworkServiceEndpointName
+
+	request.Connection.Labels = s.labels
```
denis-tingaikin, line 39: Should we join labels here?

Mixaster995: Probably confusing naming: renamed package and moved it to appropriate folder

pkg/networkservice/common/replacelabels/client.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"google.golang.org/grpc"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+type injectLabelsClient struct {
```
denis-tingaikin, line 31: Should we also rename the struct?

Mixaster995: ok, done

d-uzlov (approved):

pkg/networkservice/common/replacelabels/client.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, err
+	}
+
+	conn.Labels = oldConnLabels
+	conn.NetworkServiceEndpointName = oldNetworkServiceEndpointName
+
+	return conn, nil
+}
+
+func (s *replaceLabelsClient) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
```
Bolodya1997, line 53: Please ensure in tests that `Close` is also being invoked for the whole chain. If there is no `conn.NetworkServiceEndpointName` replacing, it wouldn't go to the subsequent Endpoints.

Mixaster995: ok, done

Bolodya1997 (approved):

pkg/networkservice/common/replacelabels/client.go
denis-tingaikin: Why we don't replace labels on close?

Mixaster995: done

PR #924, opened 2021-05-14
[qfix] Fix data race caused by querycache

Bolodya1997:
## Description
Adds missed NSE clone on return in `querycache`.

## Logs
```
2021-05-13T06:34:36.2512451Z ==================
2021-05-13T06:34:36.2512901Z WARNING: DATA RACE
2021-05-13T06:34:36.2513439Z Read at 0x00c000970c98 by goroutine 21:
...
2021-05-13T06:34:36.2526316Z   github.com/networkservicemesh/api/pkg/api/registry.(*NetworkServiceEndpoint).Clone()
2021-05-13T06:34:36.2527457Z       C:/Users/runneradmin/go/pkg/mod/github.com/networkservicemesh/api@v0.0.0-20210509180413-5753c9f30588/pkg/api/registry/helpers.go:101 +0xd7
2021-05-13T06:34:36.2528765Z   github.com/networkservicemesh/sdk/pkg/registry/common/querycache.(*queryCacheNSEClient).findInCache()
...
2021-05-13T06:34:36.2622739Z Previous write at 0x00c000970c98 by goroutine 102:
2021-05-13T06:34:36.2623804Z   github.com/networkservicemesh/sdk/pkg/registry/common/localbypass.(*localBypassNSEFindServer).Send()
...
```

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/registry/common/querycache/nse_client.go
+++ b/pkg/registry/common/querycache/nse_client.go
@@ -64,7 +64,7 @@ func (q *queryCacheNSEClient) Find(ctx context.Context, query *registry.NetworkS
 
 	resultCh := make(chan *registry.NetworkServiceEndpoint, len(nses))
 	for _, nse := range nses {
-		resultCh <- nse
+		resultCh <- nse.Clone()
 		q.storeInCache(ctx, nse, opts...)
 	}
 	close(resultCh)
--- a/pkg/registry/common/querycache/nse_client.go
+++ b/pkg/registry/common/querycache/nse_client.go
@@ -64,8 +64,8 @@ func (q *queryCacheNSEClient) Find(ctx context.Context, query *registry.NetworkS
 
 	resultCh := make(chan *registry.NetworkServiceEndpoint, len(nses))
 	for _, nse := range nses {
-		resultCh <- nse.Clone()
-		q.storeInCache(ctx, nse, opts...)
+		resultCh <- nse
+		q.storeInCache(ctx, nse.Clone(), opts...)
 	}
 	close(resultCh)
 


glazychev-art (approved):

d-uzlov (approved):

Mixaster995 (approved):

pkg/registry/common/querycache/nse_client.go
```diff
@@ -64,7 +64,7 @@ func (q *queryCacheNSEClient) Find(ctx context.Context, query *registry.NetworkS
 
 	resultCh := make(chan *registry.NetworkServiceEndpoint, len(nses))
 	for _, nse := range nses {
-		resultCh <- nse
+		resultCh <- nse.Clone()
 		q.storeInCache(ctx, nse, opts...)
```
denis-tingaikin, line 68: Can we also store the clone in the cache to prevent possible issues?
```suggestion
                nseClone := nse.Clone()
		resultCh <- nseClone
		q.storeInCache(ctx, nseClone, opts...)
```

Bolodya1997: There is no actual need for this, because we are already receiving cloned entities from `Find`. So the problem is that `querycache` violates this assumption returning not cloned entity from `Find`.

denis-tingaikin: Will localbypass then modify the stored value in the cache?

Bolodya1997: In existing code / with your suggestion - yes.
In such case: `resultCh <- nse.Clone()` - no.

denis-tingaikin: ok, as I understand we still have the problem if we'll move local bypass after querycache. Is it correct?
Suggestion updated.

Bolodya1997: Yep, the problem with data race will be revealed somewhere else in the future.

denis-tingaikin (approved):

Issue #925, opened 2021-05-17
Connect client connection and heal client monitor stream fail in different times


Bolodya1997:
## Intro

We have connect client with a gRPC connection to the remote side. We have heal client with a monitor gRPC stream over this connection.

## Expected Behavior

1. If monitor gRPC stream closes, gRPC connection should be closed or be already closed.
2. If monitor gRPC stream fails to start, gRPC connection should be closed and all Requests should return an error.

## Current Behavior

1. Sometimes we have gRPC stream closed, but gRPC connection still working and passing Requests.
2. Sometimes gRPC stream fails to start, but Requests return success.

## Steps to Reproduce

1. Run multiple times heal Endpoint tests.

## Failure logs

```
2021-05-14T11:06:05.4654511Z 2021/05/14 11:04:05 [DEBUG] --- START: TestNSMGR_HealEndpoint/Remote_New
2021-05-14T11:06:05.4655234Z 2021/05/14 11:04:05 [INFO]  Started listening Registry on: tcp://127.0.0.1:36973
2021-05-14T11:06:05.4656499Z 2021/05/14 11:04:05 [INFO]  Started listening NSMgr nsmgr-263cb81b-eebf-4eef-b3bf-87c8f6d111ee on tcp://127.0.0.1:45251
2021-05-14T11:06:05.4658517Z 2021/05/14 11:04:05 [INFO]  Started listening endpoint 7556d6a6-d0f5-49c6-9aec-90d9e47ffe97-forwarder-722ddefe-9a47-4333-ad16-668898ca774b#interpose-nse on tcp://127.0.0.1:37911
2021-05-14T11:06:05.4660415Z 2021/05/14 11:04:05 [INFO]  Started listening NSMgr nsmgr-beb8804b-e1d0-4986-8262-d311316ee98b on tcp://127.0.0.1:44327
2021-05-14T11:06:05.4662452Z 2021/05/14 11:04:05 [INFO]  Started listening endpoint baf61bc0-1a6b-44f2-a0b3-488b83cd4cd2-forwarder-bebfdb01-0f58-4810-8497-eda567dbcbfc#interpose-nse on tcp://127.0.0.1:36087
2021-05-14T11:06:05.4664415Z 2021/05/14 11:04:05 [INFO]  Started listening endpoint 890dc25b-8749-4522-8461-2e4524f293f7-final-endpoint on tcp://127.0.0.1:43643
[...]

PR #926, opened 2021-05-17
[sdk#925] Add waiting for the monitor updates

Bolodya1997:
### Problem:
gRPC requests and streaming requests have different requirements for the underlying gRPC connection. There can be a case on the closing gRPC connection, when it is still sending requests but fails to keep or start a streaming request. In such case we can get a situation when we have a Connection with no monitor stream, so it would never be healed.

### Solution:
We should check that monitor stream is ready for the Connection, before returning it back. This check should wait for some time after the next.Request(), because for the [NSMgr -> Endpoint] case response may come faster than monitor stream update.

## Issue link
https://github.com/networkservicemesh/sdk/issues/925

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/heal/client.go
+++ b/pkg/networkservice/common/heal/client.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/clock"
 )
 
 type connectionInfo struct {
@@ -120,7 +121,7 @@ func (u *healClient) Request(ctx context.Context, request *networkservice.Networ
 	case <-ctx.Done():
 		return nil, ctx.Err()
 	case <-condCh:
-	case <-time.After(100 * time.Millisecond):
+	case <-clock.FromContext(ctx).After(100 * time.Millisecond):
 		_, _ = next.Client(ctx).Close(ctx, conn)
 		return nil, errors.Errorf("timeout waiting for connection monitor: %s", conn.GetId())
 	}
--- a/pkg/networkservice/common/heal/client.go
+++ b/pkg/networkservice/common/heal/client.go
@@ -20,6 +20,7 @@ package heal
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
@@ -30,7 +31,7 @@ import (
 )
 
 type connectionInfo struct {
-	mut    sync.Mutex
+	cond   *sync.Cond
 	conn   *networkservice.Connection
 	active bool
 }
@@ -76,8 +77,8 @@ func (u *healClient) init(ctx context.Context, conn *networkservice.Connection) 
 	return nil
 }
 
-func (u *healClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
-	conn := request.GetConnection()
+func (u *healClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (conn *networkservice.Connection, err error) {
+	conn = request.GetConnection()
 	u.initOnce.Do(func() {
 		u.initErr = u.init(ctx, conn)
 	})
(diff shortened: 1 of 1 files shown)


Mixaster995 (approved):

denis-tingaikin:
With this PR `result` of networkservice client will depend on the result of connecting to the `MonitorConnectionServer`. This a bit confusing me. I would expect that both services are working separately. 

/cc @edwarnicke

pkg/networkservice/common/heal/client.go
```diff
@@ -88,25 +89,72 @@ func (u *healClient) Request(ctx context.Context, request *networkservice.Networ
...
+	//   it would never be healed.
+	// Solution:
+	//   We should check that monitor stream is ready for the Connection, before returning it back. This check should
+	//   wait for some time after the next.Request(), because for the [NSMgr -> Endpoint] case response may come faster
+	//   than monitor stream update.
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-condCh:
+	case <-time.After(100 * time.Millisecond):
```
d-uzlov, line 123: we should probably use clocktime.After here

Bolodya1997: Yes, sure, thank you for the notice :)

Bolodya1997:
> With this PR `result` of networkservice client will depend on the result of connecting to the `MonitorConnectionServer`. This a bit confusing me. I would expect that both services are working separately.

Connecting to the `MonitorConnectionServer` initialize the healing for the Connection. So if we fail to do some initialization stuff, it is the same for all chain elements - we should close next and return an error.

denis-tingaikin:
> Connecting to the MonitorConnectionServer initialize the healing for the Connection. So if we fail to do some initialization stuff, it is the same for all chain elements - we should close next and return an error.


1. Remote side could not have MonitorConnectionServer.
2. If networkservice.Server returned OK response then we should think that everything is OK. We should not wait for MonitorConnectionServer in networkservice.Server.Request. Because otherwise, these two servers start to look strongly connected.
3. Responsibility of grpc.Conn should be deligated to `connect` server. I think it is not good to do `connect` things in other chain elements...

Bolodya1997:
> 1. Remote side could not have MonitorConnectionServer.

Heal client does the only thing - it listens for the monitor server updates and translates these updates into the heal, restore invocations. So I suppose there is totally no reason to use heal client without monitor server on the remote side and it should definitely return an error if we are expecting a monitor server on the remote side and there is no.

> 2. If networkservice.Server returned OK response then we should think that everything is OK. We should not wait for MonitorConnectionServer in networkservice.Server.Request. Because otherwise, these two servers start to look strongly connected.

We have some count of chain elements first doing request and after failing if something goes wrong. Next chain element returning OK doesn't mean that everything is OK for the current chain element.
And still the same, heal client and monitor server **are** strongly connected.

> 3. Responsibility of grpc.Conn should be deligated to `connect` server. I think it is not good to do `connect` things in other chain elements...

Probably it is related to the #927, because there are no changes related to the connect in this PR :)
There are 2 different areas of responsibility: _managing gRPC connection_ and _monitoring Connection liveness_. I suppose that liveness of Connection is not strictly related to the gRPC connection state and so `connect` shouldn't make any suggestions about the Connection liveness.

pkg/networkservice/common/heal/client.go
```diff
@@ -88,25 +90,72 @@ func (u *healClient) Request(ctx context.Context, request *networkservice.Networ
...
+	//   it would never be healed.
+	// Solution:
+	//   We should check that monitor stream is ready for the Connection, before returning it back. This check should
+	//   wait for some time after the next.Request(), because for the [NSMgr -> Endpoint] case response may come faster
+	//   than monitor stream update.
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-condCh:
+	case <-clock.FromContext(ctx).After(100 * time.Millisecond):
```
edwarnicke, line 124: What was the thinking behind the choice of 100 ms here?

Bolodya1997: 100 ms is a common value for the unit testing.
It can be less or more, probably we will have more knowledge about common time for the request/response after the scalability testing.

edwarnicke: Any reason not to simply use the ctx.Done()  and wrap ctx.Err() it with your error message?

Bolodya1997: If connection is dying, Request can be success for it but monitor fails to start. In such case we need to close this Connection --> close gRPC connection and try again. Waiting just on context wouldn't help here.

denis-tingaikin: I think 100ms would produce problems in the future. So this value could be less or high for different systems. I would be suggesting to not use 100ms and look at something smarter from grpc box.

edwarnicke: Another possibility might be a heuristic around a percentage of the time between now and the context deadline...

denis-tingaikin:
I would be recommending to fix this something like: 

**Behavior with this patch**

1. heal client initializes monitoring in the goroutine of request on succeed response from the remote side.
2. heal client waits for a successful response from the monitor server in the goroutine of request.

**Suggestion:**

1. heal client should schedule async work with monitoring on listening events from the remote side on succeed request.
2. if heal client received error/delete event from the monitor then heal client could try to wait for the monitor to come back and after some timeout heal client could trigger heal.

edwarnicke:
@denis-tingaikin I like non-blocking :)

@Bolodya1997 thoughts?

pkg/networkservice/common/heal/client.go
Bolodya1997: So actually we can wait here additionally on `<-u.ctx.Done()` instead of timeout, because `u.ctx` will live no longer than underlying gRPC connection.

Bolodya1997:
I am afraid that if we don't require monitor success for the Request success, we need to handle few problematic scenarios as:
1. Next Request returns OK after we have already stopped the monitor stream and started healing for all connections - we need to add additional entry point for the healing.
2. Next Request returns OK for the restoring connection, but monitor fails - we restarting heal for actually no reason.

The changes I propose doesn't affect on refresh Request, it blocks only initial Request. My point is the following - if heal fails to initialize, it would be better to return an error and to try with a new Endpoint in discover. If we return OK, but we are not able to restore this connection, there will be a healing and Endpoint change after some time, so we don't actually win something here with this OK.

We already have "monitor healing" - it is the first phase of healing when we are trying to restore gRPC connection. In most cases, it wouldn't affect on the connection (if there is no Request/Close right now).

edwarnicke:
@Bolodya1997 WRT:
> I am afraid that if we don't require monitor success for the Request success

I believe at one point in my work on heal I came to a similar conclusion.  While I'd like non-blocking, it seems likely that you are correct on this point.

Bolodya1997:
> I believe at one point in my work on heal I came to a similar conclusion. While I'd like non-blocking, it seems likely that you are correct on this point.

Does it mean _approve_ or _it will be better to try some other solution_?

denis-tingaikin:
@edwarnicke Can we merge this one? We also can create an issue to rework this in the future to make non-blocking monitoring. Please share your thoughts.

denis-tingaikin:
@edwarnicke , @Bolodya1997 

With this PR we have added a few confusing changes:

1. https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/heal/client.go#L124
2. networkservice depends on monitoring service.

I think we need to create an issue to consider these problems in the future. 

@Bolodya1997 Could you please create an issue?

@edwarnicke Please correct me if I wrong.

edwarnicke:
@denis-tingaikin Concur... this is going to require rework in the future.

Bolodya1997:
> 1. https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/heal/client.go#L124

Totally missed to add [a fix](https://github.com/networkservicemesh/sdk/pull/926#discussion_r635858267) into the PR, so here is an additional one: https://github.com/networkservicemesh/sdk/pull/946.

PR #927, opened 2021-05-17
[sdk#925] Use monitor stream for canceling connect client

Bolodya1997:
### Problem
gRPC stream handles gRPC connection state change faster than our existing check and so we can have monitor stream closed, but connect client still running.

### Solution
Use monitor stream to track gRPC connection liveness.

## Issue link
#925 

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/connect/client.go
+++ b/pkg/networkservice/common/connect/client.go
@@ -24,11 +24,9 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/connectivity"
-
-	"github.com/networkservicemesh/api/pkg/api/networkservice"
 
 	"github.com/networkservicemesh/sdk/pkg/networkservice/chains/client"
 	"github.com/networkservicemesh/sdk/pkg/tools/clienturlctx"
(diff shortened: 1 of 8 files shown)


denis-tingaikin:
https://github.com/networkservicemesh/sdk/pull/926#issuecomment-842109614

denis-tingaikin (commented):
Can we get rid of using `cancelctx`? So for example, we can add grpc.Connection `cc` parameter into healClient.NewClient(...). In this case, we could also close the grpc.Connection on the error from the stream. 
@Bolodya1997 WDYT?

Bolodya1997:
> Can we get rid of using `cancelctx`? So for example, we can add grpc.Connection `cc` parameter into healClient.NewClient(...). In this case, we could also close the grpc.Connection on the error from the stream.
> @Bolodya1997 WDYT?

Right now `cc` in `client.Factory` is a `grpc.ClientConnInterface`, not a `*grpc.ClientConn` so this can't be just done without slightly modifying the API.
And so even closing the `*grpc.ClientConn` doesn't mean canceling the client context. So we still need to additionally monitor `cc` state in the connect client to close the client context - it looks dirty for me.

Probably we can do the following - `heal` knows about the `connect`, because:
1. `heal` means that there is something remote
2. `heal` means that there is a connection to this remote created in the chain
So we can add `connect` context methods:
```go
func WithClientCtxSupplier(parent context.Context, clientCtxSupplier func() context.Context) {...}
func supplyClientCtx(requestCtx, chainCtx context.Context) context.Context {...}
```
`heal` server would create a `clientCtxSupplier` working as a `cancelctx` but with private key only for `heal` package. So it will be almost the same, but only `heal` client will be possible to get `cancel` from the Request context.

pkg/networkservice/common/connect/client.go
```diff
@@ -68,25 +64,14 @@ func (u *connectClient) init() error {
...
-			}()
-			for cc.WaitForStateChange(u.ctx, cc.GetState()) {
-				switch cc.GetState() {
-				case connectivity.Connecting, connectivity.Idle, connectivity.Ready:
-					continue
-				default:
-					return
-				}
-			}
+			<-u.ctx.Done()
```
edwarnicke, line 73: Keep in mind that if the cancel function for u.ctx is called, this won't proceed to close the connection until some point *after* that.  There are no guarantees that it happens before some action you take immediately after calling the cancel function.

This may (or may not) be a problem depending on interactions with other parts of the system.

Bolodya1997: You are right and this actually could be a problem while `ctx.Error` was checked only on Request error.
Fixed code to first check `ctx.Error` and after make a Request - it doesn't affect anything except the heal but does guarantee that restore Request will always meet canceled or already restarting connect client.

denis-tingaikin (approved):
I'm fine with this PR if we'll rework some things after release.

@edwarnicke Be free to merge this if you have not any additional comments on this PR.

Issue #928, opened 2021-05-17
Make onidle disable itself when timeout is set to 0


d-uzlov:
## Overview

`onidle` constructor should check the timeout value, and in case timeout is `0`, `null.NewServer` should be used instead.

## References

https://github.com/networkservicemesh/sdk/issues/892

PR #929, opened 2021-05-17
Make onidle disable itself when timeout is set to 0

d-uzlov:
## Issue link

fixes https://github.com/networkservicemesh/sdk/issues/928

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


denis-tingaikin:
As I can see this is related to https://github.com/networkservicemesh/cmd-nse-icmp-responder/issues/188#issuecomment-841688543

@edwarnicke Could you look at this PR?

pkg/networkservice/common/onidle/server.go
```diff
@@ -60,6 +62,10 @@ func NewServer(ctx context.Context, notify func(), options ...Option) networkser
 		opt(t)
 	}
 
+	if t.timeout == 0 {
```
edwarnicke, line 65: We should probably *default* t.timeout = 0 above one line 56.

d-uzlov: I don't think default value 0 in the `onidle` server itself makes sense, because then we get the `null` server. So we would have to specify `onidle.WithTimeout()` always, which doesn't make sense for an optional parameter.
So, I went ahead and made `timeout` parameter mandatory.

pkg/networkservice/common/onidle/server.go
```diff
@@ -46,27 +46,23 @@ type onIdleServer struct {
...
-	for _, opt := range options {
-		opt(t)
-	}
-
-	if t.timeout == 0 {
-		return null.NewServer()
-	}
-
-	t.timer = clockTime.AfterFunc(t.timeout, func() {
+	t.timer = clock.FromContext(ctx).AfterFunc(t.timeout, func() {
```
edwarnicke, line 65: Why did we remove options?

Issue #918, opened 2021-05-12
Heal doesn't update restored endpoint name


Bolodya1997:
## Expected Behavior

If endpoint restarts, it registers with a new name. Heal should update endpoint name in the connection for such endpoint.

## Current Behavior

Heal doesn't update endpoint name and so refresh requests start failing after expire happen for the old endpoint registration.

edwarnicke:
@Bolodya1997 Should Endpoint names be ephemeral between restarts?

Bolodya1997:
There is a problem that we are trying to restore existing connection and it knows about the last Endpoint name. But restored endpoint doesn't and so it registers with a new name.
It actually becomes a problem when old Endpoint registration expires and there is no more Endpoint with the old name in registry, but we are keep trying to wait until it arrives (because connection is OK and no healing starts again).

I suppose that solution should be the following:
1. Endpoint has some format of name how it can be generated - e.g. (pod_name-uuid)
2. During the registration Endpoint tries to find from the NSMgr Endpoints registered with `Name: pod_name-..., URL: my-url`
3. If there is some Endpoint with such name+url, it means that Endpoint has been restarted and so it registers itself with this name.

So all of this logic should be included into the `setid` NSE registry chain element.

edwarnicke:
OK... let me break this down to make sure I'm understanding correctly.

1. NSC1 is connected via a Path to NSE1
2. NSE1 restarts and reregisters with NSE1'

the first thing I'm confused about is: how does heal to NSE1 work at all?  NSE1' having restarted would have a completely different callback unix file socket than NSE1, and so heal should simply *fail*.

Bolodya1997:
You are right, there is no way to restore NSE if it is connected via the unix socket. But we doesn't forbid Endpoints to be registered with the IP address, so in such case the address will be the same after the restart and we will face such problem.

edwarnicke:
> But restored endpoint doesn't and so it registers with a new name.

Why not just have the endpoints register with their Pod name?

Bolodya1997:
> Why not just have the endpoints register with their Pod name?

Let's continue this discussion in the PR thread - https://github.com/networkservicemesh/sdk/pull/932#issuecomment-850088506.

Bolodya1997:
We are no more supporting restore case, so this should be closed.

PR #932, opened 2021-05-18
[sdk#918] Enable restore NSE name in setid

Bolodya1997:
## Description
Now `setid` checks if it is able to restore NSE name from registry instead of generating a new one.

## Issue link
#918 

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/registry/common/setid/client.go
+++ b/pkg/registry/common/setid/client.go
@@ -23,13 +23,12 @@ import (
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/google/uuid"
+	"github.com/networkservicemesh/api/pkg/api/registry"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
-
-	"github.com/networkservicemesh/api/pkg/api/registry"
 
 	"github.com/networkservicemesh/sdk/pkg/registry/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/grpcutils"
 )
 
 type setIDClient struct {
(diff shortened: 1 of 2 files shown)


edwarnicke:
This looks good as far as it goes... but it feels like our basic underlying problem here is that the nse name that we are given by the NSE registering is *not* the one we actually put in the registry.  I feel like we may need to simply respect the name we are given, and treat the issue of duplication as one of *authorization* from the upstream registry.

Please note in the K8s case, a good approach to this is likely to start with simply having the NSEs register themselves with their Pod name.

Thoughts?

Bolodya1997:
@edwarnicke 
> I feel like we may need to simply respect the name we are given, and treat the issue of duplication as one of authorization from the upstream registry.

Do you mean simply not checking for duplication at all?

Our logic for now is the following - it is OK to take NSE registration with already registered name if it has the same URL as the existing registry entry. Here is a problem with simple using pod name for the NSE registration without some additional generation in such case - in unix + restart case we will get the following:
```
1. Register NSE (pod-name, unix-socket-1)
2. Endpoint restarts
3. Register NSE (pod-name, unix-socket-2) <-- failure
```

Now there is a chain element on the Endpoint side that is responsible to generate a unique name for the NSE registration and a server chain element on the NSMgr, Registry sides that is responsible to give a feedback of uniqueness of the name:
```
-(name)-> setid -(uuid-name)-> checkid
* found duplication *
setid <-(aready exists error)- checkid
setid -(uuid2-name)-> checkid --> registry
<-(new uuid-name)- setid
```
So it solves the problem with the unix + restart case, but provides a new problem with the IP + restart case:
```
1. Register NSE (uuid-pod-name, ip)
2. Endpoint restart
3. Register NSE (uuid2-pod-name, ip) <-- not a failure in terms of registration but it is now
                                         a new NSE instead of refreshing the old one
```
And this PR should solve this problem with restoring back the same name in such case.

Not checking for duplication should solve both of these problems, but here can be an issue, when for some reason we have 2 Endpoints registering with the same name (probably from the different clusters, so it even can be the pod name) and it would totally break NSM work for both of these Endpoints.

What do you think of this?

edwarnicke:
@Bolodya1997 I understand the issue with NSE restart causing a different URL for the unix file socket, which means you can't match it.

I do not understand 
```
-(name)-> setid -(uuid-name)-> checkid
* found duplication *
setid <-(aready exists error)- checkid
setid -(uuid2-name)-> checkid --> registry
<-(new uuid-name)- setid
```

Its not clear to me at all what's happening there, other than a uuid is coming from somewhere... and appears to be being prepended to the name?

edwarnicke:
@Bolodya1997 Is 'uuid' for you the Pod.metadata.uid ?

Bolodya1997:
@edwarnicke 
Sorry for this raw scheme. We have such NSE registry chain:
```
setid -> ... -> checkid (NSMgr) -> ... -> checkid (Registry) -> ... -> memory
```
1. Endpoint registers itself with some `name` name, in k8s case it would be `pod-name`.
  `-(name)-> setid`
2. `setid` generates some random prefix, so `uuid` here is just some random string.
  `setid -(uuid-name)-> ... -> checkid`
3. `checkid` checks if it already has such name registered with another URL, if so, it returns gRPC error containing `already exists` error code
  `setid <-(aready exists error)- checkid`
4. `setid` receives this error, generates a new random prefix for the name and tries to register again.
  `setid -(uuid2-name)-> checkid`
5. `checkid` doesn't find any duplication and so simply passes this registration further to the chain
  `checkid -> ... -> memory`
6. So the final name returned back with the registration will be `uuid2-name`.

PR #956, opened 2021-06-02
Introducing the begin chain element

edwarnicke:
## Description

Package begin provides a chain element that can be put at the beginning of the chain, after Connection.Id has been set
but before any chain elements that would mutate the Connection on the return path.
the begin.New{Client,Server}() guarantees:

### Scope

All `Request()` or Close() events are scoped to a particular Connection, uniquely idenfitied by its Connection.Id

### Exclusivity

Only one event is processed for a Connection.Id at a time

### Order

Events for a given Connection.Id are processed in the order in which they are received

### Close Correctness

When a `Close(Connection)` event is received, begin will replace the `Connection` provided with a the last `Connection`
successfully returned from the chain for `Connection.Id`

### Midchain Originated Events

A midchain element may originate a `Request()` or `Close()` event to be processed
from the beginning of the chain (Timeout, Refresh,Heal):

```go
begin.FromContext(ctx).Request(ctx)
begin.FromContext(ctx).Close(ctx)
```

Both `Request(ctx)` and `Close(ctx)` respect `<-ctx.Done()`.

Note: if a chain is a server chain continued by a client chain, the beginning of the chain is at the beginning of
the server chain, even if there is a subsequent begin.NewClient() in the client chain.

### Midchain Originated Request Event

Example:

```go
begin.FromContext(ctx).Request(ctx)
```

will use the `networkservice.NetworkServiceRequest` from the chain's last successfully completed `Request()` event
with `networkservice.NetworkServiceRequest.Connection` replaced with the `Connection` returned by the chain's last
successfully completed `Request()` event

Optionally, Midchain elements may provide multiple:

```go
type RequestMutatorFunc func(request *networkservice.NetworkServiceRequest)
```

using:
```go
func requestMutator(request *networkservice.NetworkServiceRequest) {...}
ctx = begin.WithRequestMutator(ctx,requestMutator)
```

to be applied mutate the default Request for Midchain Originated Requests.

### Chain Placement

begin.New{Server/Client} should always proceed any chain element which:
- Maintains state
- Mutates the Connection object along the return path of processing a Request() event.

### Reasoning

networkservice.NetworkService{Client,Server} processes two kinds of events:
- `Request()`
- `Close()`

Each `Request()` or `Close()` event is scoped to a networkservice.Connection, which can be uniquely identified by its Connection.Id

For a given Connection.Id, at most one event can be processed at a time (exclusivity).
For a given Connection.Id, events must be processed in the order they were received (order).
For `Close()`, the Connection passed to it must be identical to the last one returned by the chain to insure all state
is correctly cleared (close correctness).

Typically, a chain element receives a `Request()` or `Close()` event from the element before it in the chain
and sends a `Request()` or `Close()` and either terminates processing returning an error, or sends a `Request()` or `Close()`
event to the next element in the chain.

There are some circumstances in which a `Request()` or `Close()` event needs to be originated by a chain element
in the middle of the chain, but processed from the beginning of the chain.  Examples include (but are not limited to):
- A server timing out an expired Connection
- A client refreshing a Connection so that it does not expire
- A client healing from a lost Connection
  In all of these cases, the `Request()` or `Close()` event should be processed starting at the beginning of the chain, to ensure
  that all of the proper side effects occur within the chain.

## Issue link
This is related to healing and also has applicability to Refresh and Timeout.

## How Has This Been Tested?

- [ ] Tested manually
- [x] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [x] Documentation
- [ ] Refactoring
- [ ] CI


edwarnicke:
Note:  If we choose to adopt begin, it would replace serialize in chains [here](https://github.com/networkservicemesh/sdk/blob/eaf26022ad561d74786be04d70c1aaf1f56765c5/pkg/networkservice/chains/client/client.go#L89) and [here](https://github.com/networkservicemesh/sdk/blob/eaf26022ad561d74786be04d70c1aaf1f56765c5/pkg/networkservice/chains/endpoint/server.go#L108)

with adaptation to refresh and timeout.

fkautz (approved):

edwarnicke:
@denis-tingaikin @Bolodya1997 Thoughts?

pkg/networkservice/common/begin/client.go
```diff
@@ -0,0 +1,123 @@
...
+	entry := FromContext(ctx)
+	if entry == nil {
+		entry, _ = b.LoadOrStore(request.GetConnection().GetId(), newEventOriginator(ctx, nil, b))
+		ctx = withEntry(ctx, entry)
+	}
+	recurse := false
+	select {
+	case <-ctx.Done():
+		return nil, errors.Wrap(err, "timeout in beginClient.Request")
+	case <-entry.executor.AsyncExec(func() {
```
Bolodya1997, line 52: Looks like this would lead us to deadlock if we have multiple `begin` elements in chain.
```
(1) begin server -> ... -> connect server -> (2) begin client -> ...
```
(1) first locks executor, then (2) tries to lock the same executor.

Is there any actual reason why we have this option and so want to make such chain?

pkg/networkservice/common/begin/client.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (b *beginClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (conn *networkservice.Connection, err error) {
+	// If we already have a entry from some previous begin.NewServer or begin.NewClient, use that
+	// otherwise, LoadOrStore it and add it to the ctx
+	entry := FromContext(ctx)
+	if entry == nil {
+		entry, _ = b.LoadOrStore(request.GetConnection().GetId(), newEventOriginator(ctx, nil, b))
+		ctx = withEntry(ctx, entry)
+	}
+	recurse := false
```
Bolodya1997, line 48: We have a [multiexecutor](https://github.com/networkservicemesh/sdk/blob/main/pkg/tools/multiexecutor/multi_executor.go) for this, probably it would be better to have single `multiexecutor` in a `begin` chain element and create event originators based on it how we do it in serializectx:
https://github.com/networkservicemesh/sdk/blob/77508104772665de3d55ba9fb9a29cb2333f8fa8/pkg/tools/serializectx/context.go#L55-L60

pkg/networkservice/common/begin/event_originator.go
```diff
@@ -0,0 +1,113 @@
...
+to mutate the default Request sent.
+
+Note: if a chain is a server chain continued by a client chain, the beginning of the chain is at the beginning of
+the server chain, even if there is a subsequent begin.NewClient() in the client chain.
+*/
+func (e *EventOriginator) Request(ctx context.Context) (conn *networkservice.Connection, err error) {
+	if e == nil {
+		return nil, errors.Errorf("EventOriginator.Request should not be calle on a nil receiver")
+	}
+	req := e.req.Clone()
```
Bolodya1997, line 72: Looks like this would update outdated request clone if we have 2 requests processing in the same time (e.g. Refresh + update from the Client):
1. Client update comes and locks the executor.
2. Refresh happens and takes the stored Request clone.
3. Refresh starts waiting on the executor.
4. Client unlocks the executor and updates the stored Request value.
5. Refresh starts executing with the old Request clone.
6. Refresh finishes and updates the stored Request value with an old value (and so it updates it back for the whole chain on the right).

Probably it would be better to first lock the executor and only then take a Request clone.

Bolodya1997: And so this also can lead us to the data race :)

pkg/networkservice/common/begin/client.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+type beginClient struct {
+	Map
```
Bolodya1997, line 32: Probably it would be better to use metadata for storing event originators, because in such case they will be automatically cleaned up on `Close`.

Bolodya1997:
I very like this idea :)

PR #962, opened 2021-06-04
[qfix] Don't start healing if ctx is canceled

Bolodya1997:
## Description
Restore shouldn't start heal if heal context is canceled.

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/heal/server.go
+++ b/pkg/networkservice/common/heal/server.go
@@ -193,7 +193,9 @@ func (f *healServer) restoreConnection(ctx context.Context, request *networkserv
 		}
 	}
 
-	f.processHeal(ctx, request.Clone(), opts...)
+	if ctx.Err() == nil {
+		f.processHeal(ctx, request.Clone(), opts...)
+	}
 }
 
 func (f *healServer) processHeal(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) {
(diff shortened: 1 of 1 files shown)


pkg/networkservice/common/heal/server.go
```diff
@@ -193,7 +193,9 @@ func (f *healServer) restoreConnection(ctx context.Context, request *networkserv
 		}
 	}
 
-	f.processHeal(ctx, request.Clone(), opts...)
+	if ctx.Err() == nil {
```
denis-tingaikin, line 196: As I can see we already have a check on `ctx.Err()` at line 217. Should we remove the check from line 217 with this patch?

Bolodya1997: You mean this for loop condition?
https://github.com/networkservicemesh/sdk/blob/d38c49bc61201044ab9e4060e915171881ba2826/pkg/networkservice/common/heal/server.go#L217

Bolodya1997: If so, no we shouldn't :)

Bolodya1997: Moved this check to `processHeal`.

denis-tingaikin (approved):

Issue #994, opened 2021-06-23
Heal blocks all other Requests/Closes


Bolodya1997:
## Expected Behavior

During the connection healing subsequent events should:
* Request - fails with error or succeeds and stops healing.
* Close - stops healing.

## Current Behavior

Heal makes Request without any deadline, so it hangs in discover and blocks serialize chain element, so all subsequent Requests/Closes blocks.

## Steps to Reproduce

1. Edit `TestNSMGR_CloseHeal` or `Test_ConnectToDeadNSEUsecase` adding `time.Sleep(10 * time.Millisecond)` after the NSE context cancel.
2. Run test.
3. It fails with context deadline error.

## Failure Logs

https://github.com/networkservicemesh/sdk/runs/2891847654?check_suite_focus=true#step:5:820
https://github.com/networkservicemesh/sdk/runs/2891245838?check_suite_focus=true#step:5:377

PR #989, opened 2021-06-18
Fix unstable tests in onidle package

d-uzlov:
## Issue link

fixes https://github.com/networkservicemesh/sdk/issues/988

## How Has This Been Tested?

- [x] Added unit testing to cover
- [ ] Tested manually
- [ ] Tested by integration testing
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

PR #995, opened 2021-06-24
[sdk#994] Heal deadline

Bolodya1997:
## Description
1. Adds timeout to heal requests.
2. Fixes interpose not to mark itself as "clear" on failed heal request.

## Issue link
Closes https://github.com/networkservicemesh/sdk/issues/994.

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


Mixaster995 (approved):

pkg/networkservice/common/interpose/server.go
```diff
@@ -108,8 +108,6 @@ func (l *interposeServer) Request(ctx context.Context, request *networkservice.N
 			return result, nil
 		}
 
-		l.activeConnection.Delete(activeConnID)
```
denis-tingaikin, line 111: Why do we need to delete this?

pkg/networkservice/chains/nsmgr/heal_test.go
```diff
@@ -346,8 +346,9 @@ func TestNSMGR_CloseHeal(t *testing.T) {
 	conn, err = nsc.Request(ctx, request.Clone())
 	require.NoError(t, err)
 
-	// 3. Stop endpoint
+	// 3. Stop endpoint and wait for the heal to start
 	nseCtxCancel()
+	time.Sleep(100 * time.Millisecond)
```
denis-tingaikin, line 351: Why do we need to add sleep here?

Bolodya1997: This test validates that heal can be stopped after it has been started - so we need to wait for some time to make sure that heal is really running at the moment when we are trying to stop it.
Without this sleep test is also working, but it tests preventive heal stop - and so it is not the thing that should be tested here.

pkg/networkservice/common/interpose/server.go
Bolodya1997: When heal fails its first Request, it cleans up active connection - it results in Close failing with `no active connection found` because for `interpose` it looks like trying to close not existing connection. `heal` stands in chain after `interpose`, so client cannot stop healing in such case.

pkg/networkservice/chains/nsmgr/heal_test.go
d-uzlov: Is there some other way we can make sure that heal has started?
[Recently](https://github.com/networkservicemesh/sdk/pull/989) we found out that on CI 100ms sometimes isn't enough for thread synchronization. And the test where we had issues with timeout is much simpler, so complex heal tests likely require even more time to be sure that another thread started working.

Bolodya1997: This can be problematic.
Waiting for `100ms` doesn't **guarantee** that heal has been started, but it does mean that **in most test runs it will be so**.

pkg/networkservice/common/interpose/server.go
denis-tingaikin: Hm... As I can see with this fix activeConnection map will not delete connections. 
Am I get it correctly that we got a leak of active connections?

Bolodya1997: You are right, missed this.
Fixed in other way with setting `clientURL = nil` an invoking `next.Close` in Close in case of no active connection exists.

denis-tingaikin (approved):

Mixaster995 (approved):

PR #996, opened 2021-06-24
[sdk#740] Rework sandbox to use tokenTimeout instead of tokenGenerator

Bolodya1997:
## Description
1. Reworks sandbox to set `tokenGeneratorSupplier` for domain:
```go
// SupplyTokenGeneratorFunc supplies token generator
type SupplyTokenGeneratorFunc func(tokenTimeout time.Duration) token.GeneratorFunc
```
2.  Starts using `tokenTimeout` instead of `tokenGenerator` in all `New*` methods:
```patch
// NewNSMgr creates a new NSMgr
func (n *Node) NewNSMgr(
	ctx context.Context,
	name string,
	serveURL *url.URL,
-	tokenGenerator token.GeneratorFunc,
+	tokenTimeout time.Duration,
	supplyNSMgr SupplyNSMgrFunc,
) *NSMgrEntry {
```
3. Adds `DefaultDialOptions` for domain:
```go
// DefaultDialOptions returns default dial options for the domain
func (d *Domain) DefaultDialOptions(tokenTimeout time.Duration) []grpc.DialOption {
	return DefaultDialOptions(d.supplyTokenGenerator(tokenTimeout))
}
```

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [x] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/nsmgr/heal_test.go
+++ b/pkg/networkservice/chains/nsmgr/heal_test.go
@@ -71,7 +71,7 @@ func testNSMGRHealEndpoint(t *testing.T, nodeNum int) {
 		SetRegistryProxySupplier(nil).
 		Build()
 
-	nsRegistryClient := domain.NewNSRegistryClient(ctx, sandbox.GenerateTestToken)
+	nsRegistryClient := domain.NewNSRegistryClient(ctx, sandbox.DefaultTokenTimeout)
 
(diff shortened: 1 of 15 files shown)


denis-tingaikin (changes requested):
I don't think that we need to replace tokenGen function with token timeout in sandbox due to it can be used for advanced OPA policy testing.

Issue #999, opened 2021-06-25
Sandbox redesign


Bolodya1997:
# Issue
There are few issues in the existing sandbox:
1. Doesn't use `Option` pattern - even the simplest test cases should setup all parameters.
2. Enables by default Registry Proxy, NSMgr Proxy - all tests should manually disable them.
3. Doesn't provide any tools for dynamically adding/deleting nodes.
4. Supports "application" stop only via context.

[1-2] affects on creating more complex tests, because they become starting too complex in code.
[3-4] affects on testing different healing scenarios.

# Proposal
![image](https://user-images.githubusercontent.com/20160689/123401411-f4afae00-d5d0-11eb-920c-e34b56b5822e.png)

## Create a cluster

```go
cluster := sandbox.NewBuilder(ctx, t).
	WithRegistry(memory.NewServer,
		sandbox.WithContext(`cluster.ctx`),
		sandbox.WithURL(`some generated URL`),
		sandbox.WithTokenTimeout(`DefaultTokenTimeout`),
		sandbox.WithTokenGenerator(`cluster.supplyTokenGenerator(tokenTimeout))`).
	WithRegistryProxy(proxydns.NewServer,
		sandbox.WithContext(`...`),
		sandbox.WithURL(`...`),
		sandbox.WithTokenTimeout(`...`),
		sandbox.WithTokenGenerator(`...`)).
	WithNSMgrProxy(nsmgrproxy.NewServer,
		sandbox.WithContext(`...`),
		sandbox.WithName(`UniqueName("nsmgr-proxy")`),
		sandbox.WithURL(`...`),
		sandbox.WithTokenTimeout(`...`),
		sandbox.WithTokenGenerator(`...`)).
	WithNodes(1). // setups nodes with the default (NSMgr + Forwarder) node builder
	WithCustomNodes(1, sandbox.NewNodeBuilder().
		WithNSMgr(nsmgr.NewServer,
			sandbox.WithContext(`...`),
			sandbox.WithName(`UniqueName("nsmgr")`),
			sandbox.WithURL(`...`),
			sandbox.WithTokenTimeout(`...`),
			sandbox.WithTokenGenerator(`...`)).
		WithForwarder(new(registry.NetworkServiceEndpoint),
			sandbox.WithContext(`...`),
			sandbox.WithServerAdditionalFunctionality(`{}`),
			sandbox.WithClientAdditionalFunctionality(`{}`),
			sandbox.WithURL(`...`),
[...]

PR #1002, opened 2021-06-29
[sdk#999] Make sandbox entries restartable

Bolodya1997:
## Description
Makes sandbox entries cancelable and restartable:
* `RegistryEntry`
* `NSMgrEntry`
* `EndpointEntry`

## Issue link
#999 

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/grpcutils/url.go
+++ b/pkg/tools/grpcutils/url.go
@@ -93,24 +93,3 @@ func TargetToNetAddr(target string) (network, addr string) {
 
 	return network, target
 }
-
-// CheckURLFree returns is given url is free for Listen
-func CheckURLFree(u *url.URL) bool {
-	ln, err := net.Listen(TargetToNetAddr(URLToTarget(u)))
-	if err == nil {
-		err = ln.Close()
-	}
-	return err == nil
-}
-
-// CloneURL clones given url
-func CloneURL(u *url.URL) *url.URL {
-	if u == nil {
-		return nil
-	}
-
-	cloned := new(url.URL)
-	*cloned = *u
-
-	return cloned
-}
--- a/pkg/tools/grpcutils/url.go
+++ b/pkg/tools/grpcutils/url.go
@@ -102,3 +102,15 @@ func CheckURLFree(u *url.URL) bool {
 	}
 	return err == nil
 }
+
+// CloneURL clones given url
+func CloneURL(u *url.URL) *url.URL {
+	if u == nil {
+		return nil
+	}
+
+	cloned := new(url.URL)
+	*cloned = *u
+
+	return cloned
+}
--- a/pkg/tools/sandbox/restartable_server.go
+++ b/pkg/tools/sandbox/restartable_server.go
@@ -1,0 +1,77 @@
+// Copyright (c) 2021 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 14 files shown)


pkg/tools/grpcutils/url.go
```diff
@@ -102,3 +102,15 @@ func CheckURLFree(u *url.URL) bool {
 	}
 	return err == nil
 }
+
+// CloneURL clones given url
+func CloneURL(u *url.URL) *url.URL {
```
glazychev-art, line 107: Could you please explain, why do we need to clone URL?

Bolodya1997: `grpcutils.ListenAndServe` alters given URL (without actual changes), it results in race conditions when entry restarts:
1. Client connects to `NSMgr.URL`.
2. NSMgr stops.
   * NSMgr restarts and modifies URL.
   * Client starts healing and reconnecting to `NSMgr.URL`.
4. Race condition.

glazychev-art (approved):

Mixaster995 (approved):

pkg/tools/sandbox/restartable_server.go
```diff
@@ -0,0 +1,77 @@
...
+
+func newRestartableServer(
+	ctx context.Context,
+	t *testing.T,
+	serveURL *url.URL,
+	startFunction func(ctx context.Context),
+) *restartableServer {
+	r := &restartableServer{
+		ctx:           ctx,
+		cancelCurrent: func() {},
```
denis-tingaikin, line 45: Why is this empty? Is it means that we'll not able to restart a new server?

denis-tingaikin (changes requested):

pkg/tools/grpcutils/url.go
```diff
@@ -102,3 +102,15 @@ func CheckURLFree(u *url.URL) bool {
 	}
 	return err == nil
 }
+
+// CloneURL clones given url
```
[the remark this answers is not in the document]
Bolodya1997: @denis-tingaikin 

> 1. I'm not sure that CloneURL should be located in grpcutils.

I agree, both `CloneURL` and `CheckURLFree` are only used in tests, but:
1. I am afraid of creating some general `testutils` package, because it can urgently turn into some very dirty package with lots of different methods.
2. We probably can add `urlutils` package for these, but I am not sure that it would be better than storing them in `grpcutils`.

Any thoughts on this?

> 2. Can we use struts instead pointers to get rid of CloneURL?

Here we need to refactor `nsmgr`, `client`, `registryclient`, `connect` packages to use URL struct as a `connectTo` - it would change API and so affect all deployments without any actual need for them (because it is a test specific case).
We cannot make it on the `sandbox` side, because in all cases this `connectTo` is passed via the non-sandbox options.
So for me it still looks like that using `CloneURL` in tests is the better solution.

Thoughts?

pkg/tools/sandbox/restartable_server.go
Bolodya1997: No, it actually means that we are **able** to restart a new server, and it wouldn't make anything on server stop (because it is not yet running).

pkg/tools/grpcutils/url.go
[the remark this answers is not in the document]
Bolodya1997: Moved to `sandbox`.

Issue #809, opened 2021-04-07
Endless loop in refresh with no actual connections


Bolodya1997:
### Env description
Packet automation cluster.

### Steps to reproduce
1. Edit [multi forwarder tests](https://github.com/networkservicemesh/deployments-k8s/tree/main/examples/multiforwarder) to use single NSM setup for running multiple times.
2. Run tests multiple times.

### Actual result
After some time NSMgr starts overusing CPU even if no test is running.
```
NAME                            CPU(cores)   MEMORY(bytes)
forwarder-sriov-hm2hc           1m           17Mi
forwarder-sriov-q6scc           5m           26Mi
forwarder-vpp-4x5tl             87m          356Mi
forwarder-vpp-m72gw             81m          340Mi
nsm-registry-6675db8489-qb98c   1m           12Mi
nsmgr-lmb5r                     1m           13Mi
nsmgr-pxfqr                     4163m        43Mi
```
Profiling for 120s shows that 102s NSMgr spends in refresh.
![image](https://user-images.githubusercontent.com/20160689/113812156-ca034580-9797-11eb-82df-4d31d052d3c8.png)
Logs tail is mostly meaningless and just shows the same: [nsmgr.log](https://github.com/networkservicemesh/sdk/files/6269078/nsmgr.log).

### Expected result
Almost zero CPU usage, time mostly spent in `Register`, `trace`.

denis-tingaikin:
@edwarnicke Do you mind if we move this into `To do`?

edwarnicke:
@denis-tingaikin Please do.  Is this still an active issue?

Bolodya1997:
Fixed with #1017.

PR #1017, opened 2021-07-13
[sdk#809] Refresh should not happen after failed request

Bolodya1997:
## Description
Disable refresh if request fails.

## Issue link
#809

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/refresh/client_test.go
+++ b/pkg/networkservice/common/refresh/client_test.go
@@ -36,6 +36,7 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/adapters"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/chain"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/inject/injecterror"
 	"github.com/networkservicemesh/sdk/pkg/tools/clock"
 	"github.com/networkservicemesh/sdk/pkg/tools/clockmock"
 	"github.com/networkservicemesh/sdk/pkg/tools/sandbox"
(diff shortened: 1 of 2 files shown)


denis-tingaikin (approved):

pkg/networkservice/common/refresh/client_test.go
```diff
@@ -267,3 +268,24 @@ func TestRefreshClient_Sandbox(t *testing.T) {
...
+		refresh.NewClient(ctx),
+		injecterror.NewClient(),
+	)
+
+	_, err := client.Request(ctx, &networkservice.NetworkServiceRequest{
+		Connection: new(networkservice.Connection),
+	})
+	require.Error(t, err)
+
+	goleak.VerifyNone(t)
```
denis-tingaikin, line 290: Why do we need this?
```suggestion
```

Bolodya1997: To validate that there is no pending refresh action.

denis-tingaikin: I don't quite follow we already have check on cleanup at line 273

denis-tingaikin: @Bolodya1997 Could you please explain this line?

pkg/networkservice/common/refresh/client_test.go
```diff
@@ -267,3 +268,24 @@ func TestRefreshClient_Sandbox(t *testing.T) {
 		return refreshSrv.getState() == testRefreshStateDoneRequest
 	}, sandboxTotalTimeout, sandboxStepDuration)
 }
+
+func TestRefreshClient_NoRefreshOnFailure(t *testing.T) {
```
denis-tingaikin, line 272: It seems to me this test is incorrect due to there are no expire time in request.
1. Please add expiration time and use `clockmock` to check that there are no new requests after getting an error on the request. 
2. Please add test for Close use-case. As I can see you've added a major return statement at pkg/networkservice/common/refresh/client.go:133. So this scenario should be covered.

denis-tingaikin (changes requested):

pkg/networkservice/common/refresh/client_test.go
denis-tingaikin: Can we use clockmock instead of an additional goleak call?

Bolodya1997: Yep, fixed.

pkg/networkservice/common/refresh/client_test.go
Bolodya1997: Yes, added.

denis-tingaikin (approved):

d-uzlov (approved):

denis-tingaikin (approved):

Issue #781, opened 2021-03-23
Use go:embed from go 1.16 to store OPA policies in separate files


denis-tingaikin:
## Motivation

With go 1.16 we can use files from the project. Look at https://golang.org/pkg/embed/

Currently, OPA policies are storing as strings. So now we can move them into separate files to make it easy to find/edit and to reduce the complexity of the go files.

denis-tingaikin:
@edwarnicke Also https://golang.org/pkg/embed/ can be used for vpp configs :)

edwarnicke:
@denis-tingaikin Yes :)

PR #1018, opened 2021-07-13
refactor: Rework OPA policies with go 'embed'

denis-tingaikin:

## Motivation
1. Closes https://github.com/networkservicemesh/sdk/issues/781
2. Removes token responsibility from `updatepath`. Now updatepath is just updating path.

## How Has This Been Tested?

- [X] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [X] Refactoring
- [ ] CI


pkg/tools/opa/current_token_signed_policy_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	token, err := jwt.New(jwt.SigningMethodES256).SignedString(cert.PrivateKey)
+	require.Nil(t, err)
+
+	validX509crt, err := x509.ParseCertificate(cert.Certificate[0])
+	require.Nil(t, err)
+
+	p := opa.WithCurrentTokenSignedPolicy()
+
+	samples := []*networkservice.Path{
```
Bolodya1997, line 49: Probably there should be some more tests? Or we probably don't actually need `samples` array of 1 element?

pkg/tools/opa/prev_token_signed_policy_test.go
```diff
@@ -44,23 +44,17 @@ func Test_LastTokenShouldBeSigned_Server(t *testing.T) {
 	validX509crt, err := x509.ParseCertificate(cert.Certificate[0])
 	require.Nil(t, err)
 
-	p := opa.WithLastTokenSignedPolicy()
+	p := opa.WithPrevTokenSignedPolicy()
 
 	samples := []*networkservice.Path{
```
Bolodya1997, line 49: So same here.

pkg/tools/opa/current_token_signed_policy_test.go
denis-tingaikin: Thanks, fixed.

pkg/tools/opa/prev_token_signed_policy_test.go
denis-tingaikin: Thanks, fixed.

pkg/networkservice/chains/endpoint/server.go
```diff
@@ -103,16 +103,16 @@ func NewServer(ctx context.Context, tokenGenerator token.GeneratorFunc, options
 	rv.NetworkServiceServer = chain.NewNamedNetworkServiceServer(
 		opts.name,
 		append([]networkservice.NetworkServiceServer{
+			serialize.NewServer(),
```
Bolodya1997, line 106: `serialize` cannot stand before `updatepath` - in such case all Request/Close events generated from `refresh`/`timeout`/`heal` will race with Request/Close coming from previous hop, because they will have different `Connection.Id`.

pkg/tools/opa/policies/curr_token_signed.opa
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2020 Cisco and/or its affiliates.
```
Bolodya1997, line 1: Please use `.rego` extension for OPA files - this extension is supported by default by Goland OPA plugin.

denis-tingaikin: Thats good catch. Will do :)

pkg/networkservice/chains/endpoint/server.go
denis-tingaikin: It seems as a problem with `serialize `.  But ok, I'll fix sequence because @edwarnicke planning to replace `serialize` to `begin`.

denis-tingaikin: @Bolodya1997 Do you have any idea why all tests working fine when `serialize` is first in the chain? Did we miss scenario coverage tests?

denis-tingaikin: CI for chain when serialize is first https://github.com/networkservicemesh/sdk/actions/runs/1040411831

Bolodya1997: We don't have sandbox tests for timeout/refresh overlapping with Close/Request coming from the Client - it is pretty hard to guarantee such case without using `clock` in sandbox.

Bolodya1997 (approved):

PR #1036, opened 2021-07-23
[qfix] Fix panic on metadata chain elements double Close

Bolodya1997:
## Description
Fixes `metadata.Map(ctx)` panic on double Close.

## Issue link
Related to kind integration tests failure - https://github.com/networkservicemesh/integration-k8s-kind/runs/3136678807?check_suite_focus=true.

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/utils/metadata/client.go
+++ b/pkg/networkservice/utils/metadata/client.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Cisco and/or its affiliates.
+// Copyright (c) 2020-2021 Cisco and/or its affiliates.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -39,11 +39,9 @@ func NewClient() networkservice.NetworkServiceClient {
 func (m *metaDataClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
 	connID := request.GetConnection().GetId()
 
-	ctx = store(ctx, connID, &m.Map)
-
-	conn, err := next.Client(ctx).Request(ctx, request, opts...)
+	conn, err := next.Client(ctx).Request(store(ctx, connID, &m.Map), request, opts...)
 	if err != nil {
-		del(ctx, connID, &m.Map)
+		m.Map.Delete(connID)
 		return nil, err
 	}
 
@@ -51,6 +49,9 @@ func (m *metaDataClient) Request(ctx context.Context, request *networkservice.Ne
 }
 
 func (m *metaDataClient) Close(ctx context.Context, conn *networkservice.Connection, opts ...grpc.CallOption) (*empty.Empty, error) {
-	ctx = del(ctx, conn.GetId(), &m.Map)
-	return next.Client(ctx).Close(ctx, conn, opts...)
+	rv, err := next.Client(ctx).Close(store(ctx, conn.GetId(), &m.Map), conn, opts...)
+
+	m.Map.Delete(conn.GetId())
+
+	return rv, err
 }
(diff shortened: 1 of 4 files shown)


pkg/networkservice/utils/metadata/client.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Cisco and/or its affiliates.
+// Copyright (c) 2020-2021 Cisco and/or its affiliates.
```
d-uzlov, line 1:
```suggestion
// Copyright (c) 2020 Cisco and/or its affiliates.
//
// Copyright (c) 2021 Doc.ai and/or its affiliates.
```

pkg/networkservice/utils/metadata/client.go
```diff
@@ -39,18 +39,19 @@ func NewClient() networkservice.NetworkServiceClient {
 func (m *metaDataClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
 	connID := request.GetConnection().GetId()
 
-	ctx = store(ctx, connID, &m.Map)
-
-	conn, err := next.Client(ctx).Request(ctx, request, opts...)
+	conn, err := next.Client(ctx).Request(store(ctx, connID, &m.Map), request, opts...)
 	if err != nil {
-		del(ctx, connID, &m.Map)
+		m.Map.Delete(connID)
```
edwarnicke, line 44: How is the Close call going to access the metadata it needs to clean up resources?  Before it was guaranteed access to that data by del()... now it will not have it at all and we should expect to leak all resources...

Bolodya1997: Close is currently making:
```
next.Server(ctx).Close(store(…), conn)
```
So it has access to metadata.
Actually the problem before was Close making:
```
next.Server(ctx).Close(del(…), conn)
```
And so on double Close there were no metadata.

pkg/networkservice/utils/metadata/context.go
```diff
@@ -34,15 +34,6 @@ func store(parent context.Context, id string, mdMap *metaDataMap) context.Contex
 	return parent
 }
 
-func del(parent context.Context, id string, mdMap *metaDataMap) context.Context {
-	_, ok := parent.Value(metaDataKey{}).(*metaData)
-	if !ok {
-		md, _ := mdMap.LoadAndDelete(id)
-		return context.WithValue(parent, metaDataKey{}, md)
-	}
-	return parent
```
edwarnicke, line 43: Its seems to me the better fix here would be:

```suggestion
return context.WithValue(parent,metaDataKey{},&metaData{})
```

edwarnicke: I don't think you quite understood my suggestion... I believe that *just* this change will solve your panic issue, no need for other changes.  Small, simple, targeted.

edwarnicke: Oh wait... I read to quickly... you did fix this :)

Bolodya1997: So actually this fix leads to the same error, because on L40 we are getting `md == nil` and so on L41 we just storing it into the context.
But this direction looks better, so:
```go
if _, ok := parent.Value(metaDataKey{}).(*metaData); !ok {
	if md, ok := mdMap.LoadAndDelete(id); ok {
		return context.WithValue(parent, metaDataKey{}, md)
	}
	return context.WithValue(parent, metaDataKey{}, new(metaData))
}
return parent
```

pkg/networkservice/utils/metadata/client.go
Bolodya1997: Fixed.

Issue #1033, opened 2021-07-20
Dynamically increase dial timeout in `connect` client


Bolodya1997:
## Issue
Currently dial timeout in `connect` client is a configurable but static value:
* Setting it too high can lead us to delays if there are some still registered but already not accessible remote NSMgrs.
* Setting it too low can lead us to the failed Requests on high load or on some lasting network problems.

## Solution
Dynamically increase/decrease dial timeout after failed/succeed dial attempts.

PR #1037, opened 2021-07-23
[qfix] Increase default dial timeout to 1 second

Bolodya1997:
## Description
Default `100ms` dial timeout is too low sometimes, increased to `1s`.

## Issue link
Related to kind integration tests failure - https://github.com/networkservicemesh/integration-k8s-kind/runs/3136678807?check_suite_focus=true.

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] ~Bug fix~ Modifying default value
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/connect/server.go
+++ b/pkg/networkservice/common/connect/server.go
@@ -72,7 +72,7 @@ func NewServer(
 	s := &connectServer{
 		ctx:               ctx,
 		clientFactory:     clientFactory,
-		clientDialTimeout: 100 * time.Millisecond,
+		clientDialTimeout: time.Second,
 	}
 
 	for _, opt := range options {


edwarnicke:
Could you explain why we are increasing this timeout?

Bolodya1997:
> Could you explain why we are increasing this timeout?

Last failure of NSE composition test case was caused by NSMgr continuously failing to connect to the passthrough-1 NSE for about 1 minute, looks like it turns into a problem if host is under high load.
There is actually an issue suggesting dynamically changing this timeout due to failed/succeed dial attempts - https://github.com/networkservicemesh/sdk/issues/1033. But this PR is just a quick fix to make tests well working.

PR #1040, opened 2021-07-26
[sdk#1039] Reorder client chain elements

### LINKED ISSUE #1039: Close generated on Request from the client chain should prevent healing start
## Expected Behavior

Close generated on Request from the client chain should prevent healing start on the same hop.

## Current Behavior

1. If authorize client fails authentication, it sends Close and so after happens heal (both for pure client, server-side client).
2. If any of additional client chain elements sends Close during the request, it also causes heal start (only for server-side client).

## Steps to Reproduce

1. Start Client and wait for the token expiration.
2. Heal starts working on Client.

1. Make VPP Forwarder fail refresh Request with further remote side Close.
2. Heal starts working on Forwarder.

Bolodya1997:
## Description
Puts `heal` client after all other chain elements, right before `networkservice.NewNetworkServiceClient(cc)` in client, client factory chains.

Depends on https://github.com/networkservicemesh/sdk/pull/1056.

## Issue link
Closes #1039.

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/client/client.go
+++ b/pkg/networkservice/chains/client/client.go
@@ -120,8 +120,8 @@ func NewClient(ctx context.Context, connectTo *url.URL, clientOpts ...Option) ne
 					return chain.NewNetworkServiceClient(
 						append(
 							opts.additionalFunctionality,
-							heal.NewClient(ctx, networkservice.NewMonitorConnectionClient(cc)),
 							opts.authorizeClient,
+							heal.NewClient(ctx, networkservice.NewMonitorConnectionClient(cc)),
 							networkservice.NewNetworkServiceClient(cc),
 						)...,
 					)
@@ -152,9 +152,9 @@ func NewClientFactory(clientOpts ...Option) connect.ClientFactory {
 					serialize.NewClient(),
 					refresh.NewClient(ctx),
 					metadata.NewClient(),
-					heal.NewClient(ctx, networkservice.NewMonitorConnectionClient(cc)),
 				}, opts.additionalFunctionality...),
 				opts.authorizeClient,
+				heal.NewClient(ctx, networkservice.NewMonitorConnectionClient(cc)),
 				networkservice.NewNetworkServiceClient(cc),
 			)...)
 		return rv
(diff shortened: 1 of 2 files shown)


Mixaster995 (approved):

pkg/networkservice/common/connect/server_test.go
```diff
@@ -262,16 +263,22 @@ func TestConnectServer_RequestParallel(t *testing.T) {
...
 
-	assert.Equal(t, int32(parallelCount), serverClient.count)
-	assert.Equal(t, int32(parallelCount), serverA.count)
-	assert.Equal(t, int32(parallelCount), serverNext.count)
+	assert.Equal(t, int32(parallelCount), atomic.LoadInt32(&serverClient.Requests))
+	assert.Equal(t, int32(0), atomic.LoadInt32(&serverClient.Closes))
+	assert.Equal(t, int32(parallelCount), atomic.LoadInt32(&serverA.Requests))
+	assert.Equal(t, int32(0), atomic.LoadInt32(&serverA.Closes))
+	assert.Equal(t, int32(parallelCount), atomic.LoadInt32(&serverNext.Requests))
+	assert.Equal(t, int32(0), atomic.LoadInt32(&serverNext.Closes))
```
d-uzlov, line 271: Why do we use `assert` here (and below), instead of `require`?

Bolodya1997: Here we need to use `assert` before `273:    barrier.Done()`, because `require` will lead us to goroutines locked on `barrier` and so lots of errors.
But here is my fault, after this we can continue using `require` and so there is no need for any `assert`.

Fixed L276-L281 to use only `require`.

d-uzlov: I think it would help to add a comment about this before these asserts, to prevent people from trying to refactor it in the future.

Bolodya1997: Added comments near `barrier` initialization and near `barrier.Done()`.

d-uzlov (approved):

glazychev-art (approved):

denis-tingaikin (commented):
Looks good. 

@Bolodya1997 Could you please move refactoring regarding to count server to separate PR.

Thanks.

denis-tingaikin (approved):

PR #1050, opened 2021-08-03
[sdk-vpp#314] Add `opa.Policies` with reading from string

Bolodya1997:
## Description
Adds `opa.Policies` implementing `TextUnmarshaller` interface to make possible adding extra OPA policies using envconfig.

## Issue link
Needed for networkservicemesh/sdk-vpp#314.

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/opa/check_name.rego
+++ b/pkg/tools/opa/check_name.rego
@@ -1,0 +1,23 @@
+# Copyright (c) 2021 Doc.ai and/or its affiliates.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package test
+
+default check_name = false
+
+check_name {
+    input.path_segments[_].name == "test-name"
+}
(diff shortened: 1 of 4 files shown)


Mixaster995 (approved):

glazychev-art (approved):

denis-tingaikin:
This PR blocked by https://github.com/networkservicemesh/sdk/pull/1050

Bolodya1997:
I am planning to use in multiforwarder tests for the forwarder selection:
https://github.com/networkservicemesh/deployments-k8s/tree/8838e72452853399c3336e0f878780660c0663dc/examples/use-cases/Kernel2Kernel%26Vfio2Noop

pkg/tools/opa/check_name.rego
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package test
+
+default check_name = false
+
+check_name {
+    input.path_segments[_].name == "test-name"
+}
```
denis-tingaikin, line 23: @Bolodya1997 Am I get it correctly that for rejecting forwarder we'll use opa policy focused on name prefix?

Bolodya1997: It is just a policy for the unit tests.
Here is a policy for rejecting forwarder - https://github.com/networkservicemesh/deployments-k8s/blob/8838e72452853399c3336e0f878780660c0663dc/examples/use-cases/Kernel2Kernel%26Vfio2Noop/policies-configmap.yaml#L7.

denis-tingaikin (commented):
I like idea of adding additional OPA policies via files, but I don't quite follow motivation at this moment. 

1. We shouldn't add new features for applications only to make test working correctly. 
2. We shouldn't create a handmade data schemes and handmade deserialisers(see at implementation of `UnmarshalText`

As I can see the main motivation of this PR is fixing the integration tests with multi forwarders which will be brorken on merge sriov and vpp forwarders.

So I would suggest the next **solutions** to solve the problem:
1. Disable test while we have not two forwarders that supports a different mechanisms.
2. Change the test to working with one forwarder that supports a few different connection types.

denis-tingaikin:
Finally, we don't need to add this.

Issue #1048, opened 2021-07-30
Registry returns incorrect result


glazychev-art:
## Current Behavior

Registry can return incorrect result, if we already registered network service with the same name but with a different payload.

## Failure Information (for bugs)
Let's take a look at `TestMemif2Wireguard2Kernel` logs. 
We need to use IP payload instead of ETHERNET.

[TestRunBasicSuite.zip](https://github.com/networkservicemesh/sdk/files/6907643/TestRunBasicSuite.zip)

## Steps to Reproduce

Apparently, we have the next steps in this test:
1. **[Test A]**: Run test
2. **[Test A]**: Register NS: `{"name":"icmp-responder","payload":"ETHERNET"}`
3. **[Test A]:** Run ping and finish test
4. **[Test B]**: Run test
5. **[Test B]**: Register NS: `{"name":"icmp-responder","payload":"IP"}`
6. **[Test B]**: NS from previous test with wrong payload is used
7. **[Test B]**: Ping works fine, because we use VxLAN mechanism instead of Wireguard

edwarnicke:
Good catch!

PR #1052, opened 2021-08-04
Return error on trying to register Network Service with same name

Mixaster995:

fixes https://github.com/networkservicemesh/sdk/issues/1048

### DIFF
--- a/pkg/registry/common/memory/ns_server.go
+++ b/pkg/registry/common/memory/ns_server.go
@@ -23,6 +23,7 @@ import (
 	"github.com/edwarnicke/serialize"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/google/uuid"
+	"github.com/pkg/errors"
 
 	"github.com/networkservicemesh/api/pkg/api/registry"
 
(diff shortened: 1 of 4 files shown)


pkg/registry/common/memory/ns_server.go
```diff
@@ -59,6 +60,11 @@ func (s *memoryNSServer) Register(ctx context.Context, ns *registry.NetworkServi
 		return nil, err
 	}
 
+	loaded, ok := s.networkServices.Load(r.Name)
```
Bolodya1997, line 63:
```suggestion
	loaded, ok := s.networkServices.LoadOrStore(r.Name, r.Clone())
	if ok && loaded.Payload != ns.Payload {
		return nil, errors.New("...")
	}
```

Mixaster995: ok, done

pkg/registry/common/memory/ns_server.go
```diff
@@ -59,6 +61,11 @@ func (s *memoryNSServer) Register(ctx context.Context, ns *registry.NetworkServi
 		return nil, err
 	}
 
+	loaded, ok := s.networkServices.LoadOrStore(r.Name, r.Clone())
+	if ok && proto.Equal(ns, loaded) {
+		return nil, errors.New("network service with this name already exist")
+	}
```
denis-tingaikin, line 67: I thought memory registry is not smart and it just replaces entries. 
I also don't quite follow why we change logic only for networkservices and don't touch networkserviceendpoints..

denis-tingaikin (changes requested):

PR #1053, opened 2021-08-04
Remove debug.Stack() from logger

glazychev-art:

## Description
It seems we don't actually need `debug.Stack()`, because `errors.New()` already [has a stack](https://github.com/pkg/errors/blob/master/errors.go#L68-L73): 
```
// New, Errorf, Wrap, and Wrapf record a stack trace at the point they are
// invoked. This information can be retrieved with the following interface:
//
//     type stackTracer interface {
//             StackTrace() errors.StackTrace
//     }
```
Even if we create `error` without  using `errors` package, `trace` chain element will wrap it to the struct with stack trace (`errors.Wrapf(...)`).

We do not need to call `debug.Stack()` for every `Errorf()` call, because we already have this information.
And we need to set `"%+v"`  if we want to print error with stack trace - [link](https://github.com/pkg/errors/blob/master/errors.go#L63-L64) 

## Issue link
https://github.com/networkservicemesh/deployments-k8s/issues/2368

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/log/spanlogger/spanlogger.go
+++ b/pkg/tools/log/spanlogger/spanlogger.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"runtime/debug"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


glazychev-art:
**There are several cases if error happen:**
1. Print stack trace for each chain element:
  **\+** full info about an error (including line number)
  **\-**  a huge number of duplicate logs
2. Print only if the error hasn't stack trace ([current behavior](https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/core/trace/client.go#L61-L67) in `trace` chain element):
  **\+** for some cases we print a stack trace only once
  **\-**  if the error already has a stack trace, we don't print it

glazychev-art:
@edwarnicke 
Do you have any thoughts on this?

denis-tingaikin (approved):

denis-tingaikin:
I think we can try this and if it will make logs not informative then we can revert this patch and consider another solution.

PR #1054, opened 2021-08-05
[deployments-k8s#2412] Fix DNS

Bolodya1997:
## Description
Fixes DNS logic:

Old logic:
```
resolv.conf
---
nameserver a.a.a.a -> nameserver <default nameserver>
...
```
```
Corefile
---
. {
    forward . a.a.a.a
    log
    reload
}
...
```

This is incorrect, because:
1. User cannot both use <default nameserver> and DNS configs passed from Endpoints, because there is no `127.0.0.1` nameserver entry in `resolv.conf` in such case.
2. In `cmd-nsc-init` + `cmd-nsc` case we have `a.a.a.a` == `127.0.0.1` at the `cmd-nsc` startup, so we have such files:
```
resolv.conf
---
nameserver 127.0.0.1
...
```
```
Corefile
---
. {
    forward . 127.0.0.1
    log
    reload
}
...
```
In such case we have no access to any OS default NS servers (we even cannot ping `google.com`) and it leads to loop DNS queries.

New logic:
```
resolv.conf
---
nameserver 127.0.0.1
nameserver a.a.a.a
...
```
```
Corefile
---
. {
    forward . <default nameserver> # <-- or just without this line if it is not set
    log
    reload
}
...
```

## Issue link
Closes networkservicemesh/deployments-k8s#2412.

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


glazychev-art (approved):

pkg/networkservice/chains/nsmgr/single_test.go
```diff
@@ -73,7 +73,7 @@ func Test_DNSUsecase(t *testing.T) {
 	err = ioutil.WriteFile(resolveConfigPath, []byte("nameserver 8.8.4.4\nsearch example.com\n"), os.ModePerm)
 	require.NoError(t, err)
 
-	const expectedCorefile = ". {\n\tforward . 8.8.4.4\n\tlog\n\treload\n}\nmy.domain1 {\n\tfanout . 8.8.4.4 8.8.8.8\n\tlog\n}"
+	const expectedCorefile = ". {\n\tlog\n\treload\n}\nmy.domain1 {\n\tfanout . 8.8.4.4 8.8.8.8\n\tlog\n}\n"
```
denis-tingaikin, line 76: This looks wrong

denis-tingaikin (changes requested):

pkg/networkservice/chains/nsmgr/single_test.go
Bolodya1997: Here are such cases:
* Do we have some nameserver entry in `resolv.conf` before running NSM client?
* Do we have some default name server set in NSM client options?

1. No (nameserver), no (default name server) - such container initially has no name server, we are adding name servers only for NSM resources, other resources like `google.com` are not accessible by domain name:
```
resolv.conf
---
nameserver 127.0.0.1
...
```
```
Corefile
---
. {
    log
    reload # to reload coredns on Corefile changes
}
my-coredns-service . {
    ...
}
...
```
2. No (nameserver), yes (default name server) - such container initially has no name server, we are adding name servers for NSM resources + default name server for other resources like `google.com`.
```
resolv.conf
---
nameserver 127.0.0.1
...
```
```
Corefile
---
. {
    forward 1.1.1.1
    log
    reload
}
my-coredns-service . {
    ...
}
...
```
3. Yes (nameserver), no (default name server) - such container initially has a name server, so it should still have access to `google.com` after all NSM configurations made.
```
resolv.conf
---
127.0.0.1  # <-- first tries this
10.96.0.10 # <-- after tries this
...
```
```
Corefile
---
. {
    log
    reload
}
my-coredns-service . {
    ...
}
...
```
4. Yes (nameserver), yes (default name server) - such container initially has a name server, but we want to additionally configure it with NSM for using another name server as default.
```
resolv.conf
---
127.0.0.1
10.96.0.10
...
```
```
Corefile
---
. {
    forward 1.1.1.1
    log
    reload
}
my-coredns-service . {
    ...
}
...
```

denis-tingaikin (approved):

PR #1065, opened 2021-08-13
[qfix] Fix client chain order

Bolodya1997:
## Description
In heal we are taking the right most connection and trying to use it for the left most chain element, so how it looks in Forwarder chain:
```
... -> mechanisms server -> ... -> mechanism translation client -> ... -> heal client -> gRPC client
```
Until #1039 it was slightly different:
```
... -> mechanisms server -> ... -> heal client -> mechanism translation client -> ... -> gRPC client
```
1. `heal` tries to close server chain using client mechanism
2. instead of closing `kerneltap` and cleaning kernel interface, it closes `vxlan` and just does nothing
3. upcoming Request fails when trying to create a new kernel interface with already existing name

Doesn't look like we can easily solve this issue without a `begin` chain element, so we probably can revert some last changes and wait for the `begin` to be created.

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/client/client.go
+++ b/pkg/networkservice/chains/client/client.go
@@ -152,9 +152,10 @@ func NewClientFactory(clientOpts ...Option) connect.ClientFactory {
 					serialize.NewClient(),
 					refresh.NewClient(ctx),
 					metadata.NewClient(),
+					// TODO: move back to the end of the chain when `begin` chain element will be ready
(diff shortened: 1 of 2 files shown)


glazychev-art (approved):

denis-tingaikin (approved):

edwarnicke:
Good catch :)

PR #1081, opened 2021-09-20
[qfix] Add refresh retries on failure

Bolodya1997:
## Description
Add refresh retries on failure.

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/refresh/client.go
+++ b/pkg/networkservice/common/refresh/client.go
@@ -29,14 +29,13 @@ import (
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 
-	"github.com/networkservicemesh/sdk/pkg/networkservice/common/begin"
-	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/metadata"
-	"github.com/networkservicemesh/sdk/pkg/tools/log"
-
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 
+	"github.com/networkservicemesh/sdk/pkg/networkservice/common/begin"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/metadata"
 	"github.com/networkservicemesh/sdk/pkg/tools/clock"
+	"github.com/networkservicemesh/sdk/pkg/tools/log"
 )
 
 type refreshClient struct {
@@ -52,6 +51,8 @@ func NewClient(ctx context.Context) networkservice.NetworkServiceClient {
 }
 
 func (t *refreshClient) Request(ctx context.Context, request *networkservice.NetworkServiceRequest, opts ...grpc.CallOption) (*networkservice.Connection, error) {
+	logger := log.FromContext(ctx).WithField("refreshClient", "Request")
+
 	conn, err := next.Client(ctx).Request(ctx, request, opts...)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


denis-tingaikin (approved):

pkg/networkservice/common/refresh/client.go
```diff
@@ -74,17 +75,25 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
...
+			select {
+			case <-cancelCtx.Done():
+				return
+			case <-afterTicker.C():
+				if err := <-eventFactory.Request(begin.CancelContext(cancelCtx)); err != nil {
+					logger.Warnf("refresh failed: %s", err.Error())
+					continue
+				}
+				return
+			}
```
edwarnicke, line 94: Why retry instead of letting the next refresh catch it?

edwarnicke: So... perhaps I'm simply confused... but this looks to me like it will call the refresh *once*, that will cancel the context (see line 72 above) and then it will return based on line 86 ... meaning... I don't see what this does for us at all.  I can see some of the other fixes in the PR... but this just introduces complexity I don't think will ever be used...

pkg/networkservice/common/refresh/client.go
```diff
@@ -74,17 +75,25 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
...
 		}
-	}(cancelCtx, afterCh)
+	}()
```
edwarnicke, line 96: Please pass the variables to the go routine explicitly, rather than by closure when possible, it precludes a variety of potential issues and bugs.

pkg/networkservice/common/refresh/client.go
```diff
@@ -74,17 +75,25 @@ func (t *refreshClient) Request(ctx context.Context, request *networkservice.Net
 	store(ctx, metadata.IsClient(t), cancel)
 
 	eventFactory := begin.FromContext(ctx)
-	timeClock := clock.FromContext(ctx)
+	clockTime := clock.FromContext(ctx)
```
edwarnicke, line 78: I do like 'clockTime' better than 'timeClock'

denis-tingaikin:
@Bolodya1997 Could you also remind why is this critical to have retries for refresh?

pkg/networkservice/common/refresh/client.go
edwarnicke: Ah wait... I see.. the context isn't canceled if our call to the downstream gets an error.

edwarnicke: OK... this is clever... it will tick at the roughly 1/3 marks even if we get an error... so not overwhelming... but keeps trying.

edwarnicke (approved):

edwarnicke:
@Bolodya1997 One question... and maybe I'm just tired, but how does the Refresh terminate?

Refresh is only for running in the real clients (not the passthroughs)... so the real clients never hit a 'timeout'... it feels like we need a termination clause in your loop that *isn't* just the cancel, something where we don't try past the point the client side runs out of runway on its own timeout... thoughts?

Bolodya1997:
> @Bolodya1997 One question... and maybe I'm just tired, but how does the Refresh terminate?
> 
> Refresh is only for running in the real clients (not the passthroughs)... so the real clients never hit a 'timeout'... it feels like we need a termination clause in your loop that _isn't_ just the cancel, something where we don't try past the point the client side runs out of runway on its own timeout... thoughts?

It is terminated only on passthrough clients with timeout. Probably we can add some timeout on client so it will close the chain if it fails to send a Request during some time?

edwarnicke:
Totally get the termination on timeout for passthrough

> Probably we can add some timeout on client so it will close the chain if it fails to send a Request during some time?

Bolodya1997:
@edwarnicke 
Please take a look at this issue - https://github.com/networkservicemesh/sdk/issues/1092.
I suppose it should fix the infinite refresh loop.

PR #1082, opened 2021-09-20
[qfix] Add WithEndpointChange heal option

Bolodya1997:
## Description
Adds `WithEndpointChange` heal option to be enabled only on NSC.

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/heal/option.go
+++ b/pkg/networkservice/common/heal/option.go
@@ -67,8 +67,16 @@ func WithRestoreTimeout(restoreTimeout time.Duration) Option {
 	}
 }
 
+// WithEndpointChange sets if Connection.EndpointName can be changed with monitor updates
+func WithEndpointChange() Option {
+	return func(healOpts *healOptions) {
+		healOpts.changeEndpoint = true
+	}
+}
+
 type healOptions struct {
 	onHeal         *networkservice.NetworkServiceClient
 	onRestore      OnRestore
 	restoreTimeout time.Duration
+	changeEndpoint bool
 }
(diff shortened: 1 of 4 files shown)


pkg/networkservice/common/heal/option.go
```diff
@@ -67,8 +67,16 @@ func WithRestoreTimeout(restoreTimeout time.Duration) Option {
 	}
 }
 
+// WithEndpointChange sets if Connection.EndpointName can be changed with monitor updates
+func WithEndpointChange() Option {
```
denis-tingaikin, line 71: @Bolodya1997 Could you please explain in what cases we plan do not use this option?

Bolodya1997: It is NSC-only option. Do you mean adding such comment in code?

denis-tingaikin: I don't quite follow why do we need the option only for nscs.

denis-tingaikin:
I don't like adding more complexity into heal, but we are planning to remove and replace exist heal logic.


## Motviation

1. https://github.com/networkservicemesh/sdk/pull/1082#discussion_r713012114
2. This PR is required for https://github.com/networkservicemesh/deployments-k8s/issues/2381

So merging this as a part of solution for https://github.com/networkservicemesh/deployments-k8s/issues/2381

PR #1088, opened 2021-09-28
[vlan multiplexing] add vlan tag for vlan supported kernel mechanism

pperiyasamy:
Signed-off-by: Periyasamy Palanisamy <periyasamy.palanisamy@est.tech>

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


JanScheurich:
Does that mean the VLAN-multiplexing will only work with local kernel mechanism? That would exclude VLAN multiplexing on SRIOV vfio devices in NSE pods. For Mellanox SmartVFs it's OK as they are kernel mechanisms as well as DPDK devices.

Bolodya1997:
@pperiyasamy 
Please add some `WithVLAN` option and use VLAN tags only if it is selected. I suppose that there are cases when we don't need VLAN tags for kernel mechanism.

edwarnicke:
@pperiyasamy Would it possibly make more sense to have a separate 'vlan' chain element that handles these details? (similar to how vni is handled)  This would give nice modularity, and allow easy use of vlan multiplexing across different mechanisms.

pperiyasamy:
@edwarnicke @JanScheurich where do you want 'vlan' chain element to be placed ? currently only kernel mechanism has [vlan parameter](https://github.com/networkservicemesh/api/blob/main/pkg/api/networkservice/mechanisms/kernel/helpers.go#L121-L154) support, so can we place it under sdk/pkg/networkservice/common/mechanisms/kernel/vlan for now ?

edwarnicke:
@pperiyasamy Sounds about right :)

edwarnicke:
@pperiyasamy This looks good.

@JanScheurich A brief comment here on 'working with other non kernel mechanisms'.  As written, the implementation @pperiyasamy has provided only works for kernel mechanism... *but* it has the right architectural whitespace to be easily extended to any Vlan supporting mechanism.

I'll talk through how briefly, and you can all decide when you'd like to make those extensions :)

Right now the constants 'SupportsVLAN' and 'VLAN' are only defined in [api/networkservice/mechanisms/kernel/constants.go](https://github.com/networkservicemesh/api/blob/9a36433d7d6e99699bbdd553249ae65d0ddd57cd/pkg/api/networkservice/mechanisms/kernel/constants.go#L53-L57).  As you can see just above there, we've in the past defined some constants by reference to 'common' because those constants get used in multiple mechanisms.

vlan multiplexing is seen as a cross-cutting concern... you could treat it as such by introducing `pkg/api/networkservice/mechanisms/vlanmultiplexing/constants.go` (to act as a home for the constants, similar to what commons does) and `pkg/api/networkservice/mechanisms/vlanmultiplexing/helpers.go` to introduce its own `ToMechanism` and accessors (which could be reused in the kernel mechanism).  This way other mechanisms could do likewise, and the vlan chain element could support whatever mechanism provides SupportsVlan.

Thoughts?

JanScheurich:
> Right now the constants 'SupportsVLAN' and 'VLAN' are only defined in [api/networkservice/mechanisms/kernel/constants.go](https://github.com/networkservicemesh/api/blob/9a36433d7d6e99699bbdd553249ae65d0ddd57cd/pkg/api/networkservice/mechanisms/kernel/constants.go#L53-L57). As you can see just above there, we've in the past defined some constants by reference to 'common' because those constants get used in multiple mechanisms.
> 
> vlan multiplexing is seen as a cross-cutting concern... you could treat it as such by introducing `pkg/api/networkservice/mechanisms/vlanmultiplexing/constants.go` (to act as a home for the constants, similar to what commons does) and `pkg/api/networkservice/mechanisms/vlanmultiplexing/helpers.go` to introduce its own `ToMechanism` and accessors (which could be reused in the kernel mechanism). This way other mechanisms could do likewise, and the vlan chain element could support whatever mechanism provides SupportsVlan.
> 
> Thoughts?

Makes sense. The current implementation in the kernel mechanism is sufficient for our OVS forwarder + SmartVF use case. We can leave the generalization to a future contributor.

edwarnicke:
The good news is its actually pretty each and non-impactful to generalize from @pperiyasamy 's implementation here :)

PR #1107, opened 2021-10-14
Add begin.WithReselect option

edwarnicke:
Signed-off-by: Ed Warnicke <hagbard@gmail.com>

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/client/client.go
+++ b/pkg/networkservice/chains/client/client.go
@@ -23,6 +23,8 @@ import (
 	"github.com/google/uuid"
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 
+	"github.com/networkservicemesh/sdk/pkg/networkservice/common/trimpath"
+
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/begin"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/clientconn"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/clienturl"
@@ -64,6 +66,7 @@ func NewClient(ctx context.Context, clientOpts ...Option) networkservice.Network
 					dial.WithDialTimeout(opts.dialTimeout),
 				),
 				opts.authorizeClient,
+				trimpath.NewClient(),
 				connect.NewClient(),
 			)...,
 		)...,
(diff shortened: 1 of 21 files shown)


edwarnicke:
@Bolodya1997 First: an apology.  You were correct that we needed to do something about 'outside requests' and how to handle them.  You'd thought of it in terms of 'diffing' things.  I thought about it in https://github.com/networkservicemesh/sdk/pull/1107/commits/c076175fbb57b147ef18040654749a26a0c6ff60 as 'merging'.  Either way, something needed to be done.  We've got a place holder in https://github.com/networkservicemesh/sdk/pull/1107/commits/c076175fbb57b147ef18040654749a26a0c6ff60 to hash out the details (if any) beyond what's there.

I pulled forward the chain/nsmgr/heal_test.go tests.  Currently all are passing except for TestNSMGR_HealForwarder.  I believe this is because the interpose chain element gets stuck on trying to reach the same old Forwarder, even if its gone.  I haven't fully gotten to the bottom of it, and that might be a good place for you to start @Bolodya1997

edwarnicke:
@Bolodya1997 Figured out two things contributing to the TestNSMGR_HealForwarder failure:

1. We are using in sandbox.DialOptions() we are using

https://github.com/networkservicemesh/sdk/blob/f264fec6da05add58321efc8a679d979c64d8c29/pkg/tools/sandbox/dial_options.go#L58-L58

https://github.com/networkservicemesh/sdk/blob/f264fec6da05add58321efc8a679d979c64d8c29/pkg/tools/sandbox/dial_options.go#L60-L60

If we don't use grpc.WithBlock() and grpc.WaitForReady() ... not only do the tests pass (with the note below) but the tests also run 5-6x faster.

 Why are we using these options when they slow things down a lot?

2.  We are looking for unique connections on heal... but the heal is instead cleanly healing the existing connection:  
 https://github.com/networkservicemesh/sdk/blob/4807755c335d44ddf0ff9685e5fe4694cb8be292/pkg/networkservice/chains/nsmgr/heal_test.go#L97-L99

https://github.com/networkservicemesh/sdk/blob/4807755c335d44ddf0ff9685e5fe4694cb8be292/pkg/networkservice/chains/nsmgr/heal_test.go#L110-L111

If we change those to checking for Requests() and Closes() with the correct expectations... that helps.

Issue #1094, opened 2021-10-03
Add support for point-to-multipoint forwarders in NSM


denis-tingaikin:
## Requires

1. https://github.com/networkservicemesh/sdk/pull/1084
2. https://github.com/networkservicemesh/sdk/pull/1083

## Overview

See at  https://docs.google.com/presentation/d/1b_jtSFygaYPQa6Xd0pRR50enieaydu6yK8fxprp7bno/edit?usp=sharing

## Link to the Google Docs document

See at https://viewer.diagrams.net/?highlight=0000ff&edit=_blank&layers=1&nav=1&page-id=D2ctnZZM9BhLXnHZu3kN&title=Forwarder%20as%20NSE.drawio#Uhttps%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D14HHm-3sY1Zkl3jkV4CRgDNhIgMvNTTkU%26export%3Ddownload

## Tasks

- [x]  1. Remove loop `nsc->nsmgr->forwarder->nmsgr->nse` to `nsc->nsmgr->forwarder->nse`. 
Total estimation: `4.3d`

- [x] 1.1. Update discover chain element for nsmgr and forwarder and test forwarder selection (interpose could be removed).
Estimation: `0.8d`

- [x] 1.2. Add support `Find` for `interpose` RegistryServer.
Estimation: `0.5d`

- [x] 1.3. Update `filtermechanisms` for forwarders .
Estimation: `1d`

- [ ] 1.4 Rework a few unit tests with the new architecture and add new ones to cover p.1 (i.e. add examples on `nsc->nsmgr->forwarder->nse1...nseN`).
Estimation: `2d`

- [ ] 2. Update dependent repositories (i.e. update forwarders).
Estimation: `0.5d`

- [ ] 3.  Add an example on forwarder selection by capabilities(labels).
Estimation: `0.5d`

- [ ] 4. Verify that each integration test is stable with the new architecture.
Estimation: `3d`

Total Estimation: `8.3d`

denis-tingaikin:
@edwarnicke Question on 4. 

How should forwarders provide their labels?

Currently, I see two options:
1. Just pass labels into env config for each forwarder and use it on registartion
2. Cluster admin should register cross-NSE such as it does for NSE.
3. Your option.


For me option 1. is looking good.

denis-tingaikin:
[...]

Issue #1118, opened 2021-10-26
`sendfd` doesn't send fd to all connections


Bolodya1997:
## Related issues
1. Continues https://github.com/networkservicemesh/sdk/issues/1111.
2. Needed for https://github.com/networkservicemesh/sdk/issues/1106.

## Expected Behavior
In such case:
1. `sendfd.Client` sends `fd` on `Close` or `Request`.
2. `connect.Client` makes `Close` or `Request` to `URL-1`.
3. ...
4. `connect.Client` makes `Close` or `Request` to `URL-n`.

We want `fd` to be sent to all `URL-1`, ..., `URL-n` receivers.

## Current Behavior
`sendfd.Client` sends `fd` only to `URL-1`.

## Cause
1. `sendfd.Client` uses [`grpcfd.(*wrapPerRPCCredentials)`](https://github.com/edwarnicke/grpcfd/blob/main/per_rpc_transport_credentials.go) for sending `fd` before the gRPC interaction.
2. `grpcfd.(*wrapPerRPCCredentials)` captures `FDTransceiver` from the first gRPC interaction (it would be `URL-1` in our case) and sends all fds only to this transceiver:
   1. https://github.com/edwarnicke/grpcfd/blob/39c35952c7c38397177cf36ed7b8026ab88c61ec/per_rpc_transport_credentials.go#L39-L48
   2. https://github.com/edwarnicke/grpcfd/blob/39c35952c7c38397177cf36ed7b8026ab88c61ec/per_rpc_transport_credentials.go#L66-L72

### Analysis
We have 4 different points where we need to use `FDTransceiver` in terms of linear execution:
1. Client requests Server and sends fd (`sendfd.Client`).
2. Server processes request and receives fd (`recvfd.Server`).
3. Server responses to the Client and sends fd (`sendfd.Server`).
4. Client processes response and receives fd (`recvfd.Client`).

We have the following:
1. At [1] we can request `grpc.Peer`, but we will get it only on [4].
2. At [2, 3] we can access `grpc.Peer` directly from the request context.

So in [2, 3, 4] cases we can simply use `grpcfd.FromPeer` to get `FDTransceiver`.

In [1] case we don't actually need `FDTransceiver`, we only need `FDSender`.

### Edit `grpcfd.(*wrapPerRPCCredentials)`
1. Implement only `FDSender` interface.
2. Stop capturing `FDTransceiver`.
```go
type wrapPerRPCCredentials struct {
[...]

PR #1122, opened 2021-10-27
feat: Add support for point-to-multipoint forwarders in NSM 

denis-tingaikin:

## Description
This PR adds a part of p2mp rework related to SDK

## Issue link
https://github.com/networkservicemesh/sdk/issues/1094

## How Has This Been Tested?

- [X] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [X] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


denis-tingaikin:
TODO:

1. Rework `crossnsediscover` due to `discover` was simplified
2. Enable all disables tests(currently 2 unix tests in SDK are not working)

pkg/networkservice/chains/nsmgr/single_test.go
```diff
@@ -253,3 +254,78 @@ func Test_ShouldParseNetworkServiceLabelsTemplate(t *testing.T) {
 	_, err = nse.Unregister(ctx, nseReg)
 	require.NoError(t, err)
 }
+
+func Test_UsecasePoint2MultiPoint(t *testing.T) {
```
denis-tingaikin, line 258: @edwarnicke Is this test look valid?

pkg/networkservice/chains/nsmgr/server.go
```diff
@@ -150,55 +149,53 @@ func NewServer(ctx context.Context, tokenGenerator token.GeneratorFunc, options
...
 	}
 
-	var nseRegistry registryapi.NetworkServiceEndpointRegistryServer
+	nsRegistry = registrychain.NewNetworkServiceRegistryServer(
+		registryserialize.NewNetworkServiceRegistryServer(),
+		setlogoption.NewNetworkServiceRegistryServer(map[string]string{"name": "NetworkServiceRegistryServer." + opts.name}),
+		nsRegistry,
+	)
+
+	var nseInMemoryRegistry = registrychain.NewNetworkServiceEndpointRegistryServer(
```
denis-tingaikin, line 167: Note: now nsmgr has its own memory registry by default.

pkg/networkservice/chains/nsmgr/single_test.go
```diff
@@ -119,7 +119,8 @@ func Test_ShouldCorrectlyAddForwardersWithSameNames(t *testing.T) {
 	require.NoError(t, err)
 
 	forwarderReg := &registry.NetworkServiceEndpoint{
-		Name: "forwarder",
+		Name:                "forwarder",
+		NetworkServiceNames: []string{"forwarder"},
```
denis-tingaikin, line 123: @edwarnicke 
1. Should we add const for forwarder networkservicename into api? 
2. Should we create a registry network service endpoint chain element for registration forwarders?

pkg/networkservice/chains/nsmgr/unix_test.go
```diff
@@ -39,6 +39,7 @@ import (
 )
 
 func Test_Local_NoURLUsecase(t *testing.T) {
+	t.Skip("https://github.com/networkservicemesh/sdk/issues/1118")
```
denis-tingaikin, line 42: Note: This is disabled because with p2mp rework forwarder should recv fd from the nsmgr. Currently it is not working in these unit tests by https://github.com/networkservicemesh/sdk/issues/1118 but these scenarios work well in integration tests.

pkg/networkservice/chains/nsmgr/unix_test.go
```diff
@@ -84,6 +85,7 @@ func Test_Local_NoURLUsecase(t *testing.T) {
 }
 
 func Test_MultiForwarderSendfd(t *testing.T) {
+	t.Skip("https://github.com/networkservicemesh/sdk/issues/1118")
```
denis-tingaikin, line 88: Note: This is disabled because with p2mp rework forwarder should recv fd from the nsmgr. Currently, it is not working in these unit tests by https://github.com/networkservicemesh/sdk/issues/1118 but these scenarios work well in integration tests.

pkg/networkservice/common/discover/match_selector.go
```diff
@@ -56,15 +36,19 @@ func matchEndpoint(clockTime clock.Clock, nsLabels map[string]string, ns *regist
...
-		if !isSubset(nsLabels, match.GetSourceSelector(), nsLabels) {
+		if !matchutils.IsSubset(nsLabels, match.GetSourceSelector(), nsLabels) {
 			continue
 		}
 		nseCandidates := make([]*registry.NetworkServiceEndpoint, 0)
 		// Check all Destinations in that match
 		for _, destination := range match.GetRoutes() {
 			// Each NSE should be matched against that destination
 			for _, nse := range validNetworkServiceEndpoints {
-				if isSubset(nse.GetNetworkServiceLabels()[ns.Name].Labels, destination.GetDestinationSelector(), nsLabels) {
```
denis-tingaikin, line 67: potential nil panic.

nse.GetNetworkServiceLabels()[ns.Name] can be nil. 

Fixed.

pkg/networkservice/common/discovercrossnse/metadata.go
```diff
@@ -14,17 +14,24 @@
...
+	if !ok {
+		return ""
+	}
+	return v.(string)
 }
 
-// Is returns true if passed name contains interpose identity
-func Is(name string) bool {
-	return strings.HasSuffix(name, nameSuffix)
+func storeForwarderName(ctx context.Context, v string) {
```
denis-tingaikin, line 35: @edwarnicke What if we'll load forwarder name from the path segments instead of metadata?

pkg/networkservice/common/interpose/server.go
```diff
@@ -1,166 +0,0 @@
...
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-// Package interpose provides a NetworkServiceServer chain element that tracks local Cross connect Endpoints and call them first
-// their unix file socket as the clienturl.ClientURL(ctx) used to connect to them.
-package interpose
```
denis-tingaikin, line 19: Note: interpose totally deleted.

pkg/registry/common/recvfd/client.go
```diff
@@ -0,0 +1,95 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build linux
+
+package recvfd
```
denis-tingaikin, line 21: Note: forwarder should be able to get fd from the local nse registration

pkg/registry/common/sendfd/server_linux.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build linux
+
+package sendfd
```
denis-tingaikin, line 21: Note: nsmgr should be able to send fd from the local nse registration

pkg/tools/sandbox/node.go
```diff
@@ -116,13 +119,22 @@ func (n *Node) NewForwarder(
...
 	entry.restartableServer = newRestartableServer(ctx, n.t, entry.URL, func(ctx context.Context) {
 		entry.Endpoint = endpoint.NewServer(ctx, generatorFunc,
 			endpoint.WithName(entry.Name),
 			endpoint.WithAdditionalFunctionality(
 				append(
-					additionalFunctionality,
-					clienturl.NewServer(CloneURL(n.NSMgr.URL)),
+					append([]networkservice.NetworkServiceServer{
+						discover.NewServer(nsClient, nseClient),
+						roundrobin.NewServer(),
```
denis-tingaikin, line 136: @edwarnicke Should we add chains/forwarder with these two chain elements?

pkg/networkservice/chains/nsmgr/single_test.go
edwarnicke: > Should we create a registry network service endpoint client chain element for registration forwarders?

I don't quite follow this, could you say more?

pkg/networkservice/chains/nsmgr/single_test.go
edwarnicke: Yes

pkg/networkservice/common/discovercrossnse/metadata.go
edwarnicke: How would that work on initial request?

pkg/networkservice/common/discovercrossnse/server.go
```diff
@@ -0,0 +1,172 @@
...
+			return nil, errors.New("forwarders not found")
+		}
+
+		nses = matchForwarders(request.Connection.GetLabels(), nses)
+
+		if len(nses) == 0 {
+			return nil, errors.New("no candidates found")
+		}
+
+		u, err := url.Parse(nses[0].Url) // TODO: Should we consider about load balancing?
```
edwarnicke, line 80: Yes, we should consider load balancing... open an issue :)

pkg/registry/common/recvfd/client.go
edwarnicke: Got it.

pkg/registry/common/sendfd/server_linux.go
edwarnicke: Got it.

pkg/networkservice/common/discovercrossnse/metadata.go
denis-tingaikin: >How would that work on initial request?

For the initial request, we do have not the next segment. 
So simply go to discover a forwarder (see at branch where `forwarderName == ""`)

pkg/tools/sandbox/node.go
edwarnicke: > Should we add chains/forwarder with these two chain elements?

Could you say more... I don't quite follow...

edwarnicke:
Do we have corresponding PRs for the Forwarders to have them do selection?

pkg/networkservice/chains/nsmgr/single_test.go
denis-tingaikin: Sure, we can add something like:

```go
type setForwarderNetworkServiceNameNSEClient struct{} 

func (n *setNetworkServiceNamesNSEClient) Register(ctx context.Context, in *registry.NetworkServiceEndpoint, opts ...grpc.CallOption) (*registry.NetworkServiceEndpoint, error) {
	in.NetworkServiceNames = []string{"forwarder"}
...
```

This client can be used by forwarders on registration.

pkg/networkservice/common/discovercrossnse/server.go
denis-tingaikin: I think this can be resolved in this PR if we go with https://github.com/networkservicemesh/sdk/pull/1122#discussion_r744229781

pkg/networkservice/chains/nsmgr/single_test.go
edwarnicke: Ah... that's interesting... but how does it handle the case of that being overriden for some reason by env variables in the forwarder?

Better question... where is the 'forwarder' string encoded in the NSMgr ?

pkg/networkservice/common/discovercrossnse/metadata.go
edwarnicke: And what if the next segment is no longer valid?

pkg/networkservice/common/discovercrossnse/server.go
edwarnicke: How would it be resolved?

pkg/tools/sandbox/node.go
denis-tingaikin: we can add a new pkg `forwarder` into `sdk/pkg/networkservice/chains/`.

So currently, it seems to me we have motivation because many forwarders may be interested in default SDK chain elements such as `roundrobin`, `disocver`, `recvfd`, `sendfd`. So we can add a default preset for them.

pkg/networkservice/common/discovercrossnse/server.go
denis-tingaikin: 1. cut this line https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/roundrobin/server.go#L64 
2. add into` sdk/pkg/networkservice/setednpoint` chain element that will set this
https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/roundrobin/server.go#L64 based on candidates and cleinturl from the context only if https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/roundrobin/server.go#L64 is empty.
3.`discovercrossnse` should pass into context candates.

Finally, we can have chains like this:

nsmgr:
```go
...
discovercrossnse.NewServer(...) // works based on pass segments. If it is presented into connection just get it otherwise add canidates for the next load balancing chain element
roundrobind.NewServer()
```

forwarder

```go
discover.NewServer(...)
roundrobind.NewServer()
setednpoint.NewServer() //sets nse name into connection.
```

edwarnicke: Ah... I see... you are separating the setting of the value in the connection from the selection of where to go next.

Would it be possible to open an issue to come back to this and go ahead and land the P2MP now?

pkg/networkservice/chains/nsmgr/single_test.go
denis-tingaikin: > Better question... where is the 'forwarder' string encoded in the NSMgr ?

See at discovercrossnse chain element.

pkg/networkservice/common/discovercrossnse/server.go
denis-tingaikin: > Would it be possible to open an issue to come back to this and go ahead and land the P2MP now?

Yes.

pkg/networkservice/common/discovercrossnse/metadata.go
denis-tingaikin: @edwarnicke  That's a good question. It looks like in this case we should select another forwawrder.

pkg/networkservice/chains/nsmgr/single_test.go
denis-tingaikin: @edwarnicke What about the first question?

pkg/networkservice/common/discovercrossnse/server.go
edwarnicke: Let's do that then :)

pkg/tools/sandbox/node.go
edwarnicke: Hmm... lets try getting the existing forwarders going first

edwarnicke: Because discover/roundrobin are kind of bespoke to our p2p forwarders... p2mp forwarders may chose to do something quite different.

denis-tingaikin: @edwarnicke Load balancing sound like option :)

pkg/networkservice/chains/nsmgr/single_test.go
edwarnicke: I'd hold off on adding a constant to the API currently.  I'm still curious how:

'forwarder' gets set as the magic string in NSMgr... I'd be more inclined to make it a settable parameter via an option and have it flow through from the top of the cmd-nsmgr ...

denis-tingaikin: >'forwarder' gets set as the magic string in NSMgr... I'd be more inclined to make it a settable parameter via an option and have it flow through from the top of the cmd-nsmgr ...

Also all forwarders should know about the constant to be correctly registered.

denis-tingaikin: if you are fine with this, then I'll add the option as you suggested.

edwarnicke: Forwarders should already be able to overload the NS name via ENV variable... do they not?

denis-tingaikin: Yes, but all forwarders and nsmgr on the cluster should be synchronized with each other.

denis-tingaikin: Mostly, I like your suggestion, but just shared another possible way

denis-tingaikin: Fixed in https://github.com/networkservicemesh/sdk/pull/1122/commits/1a0be07e86e600742b488cdc1531b329d8025d00

Issue #1123, opened 2021-10-27
Add `retry` wrapper for network service client and network service monitor


denis-tingaikin:
We should use `retry` logic since to we get rid of grpc.WithBlock and grpc.WaitForReady(true). Also with it we can simplify discover chain element.

PR #1129, opened 2021-10-31
feat: Add "retry" networkservice.Client wrapper for nscs

denis-tingaikin:

## Issue link
https://github.com/networkservicemesh/sdk/issues/1123

## How Has This Been Tested?

- [X] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [X] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/retry/client.go
+++ b/pkg/networkservice/common/retry/client.go
@@ -55,9 +55,9 @@ func WithSettings(s Settings) Option {
 func NewClient(client networkservice.NetworkServiceClient, opts ...Option) networkservice.NetworkServiceClient {
 	var result = &retryClient{
 		Settings: Settings{
-			Interval:   time.Millisecond * 100,
-			TryTimeout: time.Second,
-			Timeout:    (time.Second + time.Millisecond*100) * 15,
+			Interval:   time.Millisecond * 200,
+			TryTimeout: time.Second * 15,
+			Timeout:    (time.Second*15 + time.Millisecond*200) * 5,
 		},
 		client: client,
 	}
--- a/pkg/networkservice/common/retry/client.go
+++ b/pkg/networkservice/common/retry/client.go
@@ -1,0 +1,135 @@
+// Copyright (c) 2021 Cisco and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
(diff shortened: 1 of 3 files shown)


pkg/networkservice/common/retry/client.go
```diff
@@ -0,0 +1,135 @@
...
+type retryClient struct {
+	Settings
+	client networkservice.NetworkServiceClient
+}
+
+// Option configuress retry.Client instance.
+type Option func(*retryClient)
+
+// WithSettings sets retry policy settings for the retry.Client instance.
+func WithSettings(s Settings) Option {
```
edwarnicke, line 48: Why not do this as standard options like 

retry.WithInterval
retry.WithTryTimeout

instead of 'Settings'

pkg/networkservice/common/retry/client.go
```diff
@@ -0,0 +1,135 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/clock"
+	"github.com/networkservicemesh/sdk/pkg/tools/log"
+)
+
+// Settings represents retry policy settings.
+type Settings struct {
+	Timeout    time.Duration
```
edwarnicke, line 34: Why have timeout here instead of simply using the ctx passes into the RPC calls as normal?

pkg/networkservice/common/retry/client.go
```diff
@@ -0,0 +1,135 @@
...
+	if r.Timeout > 0 {
+		var cancel func()
+		ctx, cancel = c.WithTimeout(ctx, r.Timeout)
+		defer cancel()
+	}
+
+	for ctx.Err() == nil {
+		requestCtx, cancel := c.WithTimeout(ctx, r.TryTimeout)
+		resp, err := r.client.Request(requestCtx, request, opts...)
+		cancel()
```
edwarnicke, line 85: Shouldn't this be

```suggestion
defer cancel()
```

pkg/networkservice/common/retry/client.go
denis-tingaikin: Thanks, fixed.

pkg/networkservice/common/retry/client.go
denis-tingaikin: Thanks, fixed.

pkg/networkservice/common/retry/client.go
denis-tingaikin: Thanks, fixed.

Issue #1152, opened 2021-11-10
Use in memory logging for each sandbox test


denis-tingaikin:
## Problem statement

If test failed on ci we can get logs from nsm components.

## Solution

Always write logs into memory logger.

If `t.Failed()` push all logs into stdout.

NikitaSkrynnik:
Blocked by https://github.com/networkservicemesh/sdk/issues/1182

PR #1162, opened 2021-11-12
Memory logger for sandbox tests

NikitaSkrynnik:
Signed-off-by: Nikita Skrynnik <nikita.skrynnik@xored.com>

## Issue link
https://github.com/networkservicemesh/sdk/issues/1152

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


edwarnicke:
I'm pretty sure this is the root of our problem:

https://github.com/networkservicemesh/sdk/blob/e33b665d2e98196dd3b1a49d875b528c2f78bfce/pkg/tools/log/spanlogger/spanlogger.go#L43-L43

Its effectively a *global* lock on the entire system via the spanlogger... we should solve our race condition problem *without* such a global lock.

PR #1171, opened 2021-11-23
Made group logger creation public

Mixaster995:
Needed for https://github.com/networkservicemesh/cmd-nsmgr/pull/422

### DIFF
--- a/pkg/tools/log/grouplogger.go
+++ b/pkg/tools/log/grouplogger.go
@@ -23,8 +23,8 @@ type groupLogger struct {
 	loggers []Logger
 }
 
-// New creates grouped logger from all provided loggers
-func New(s ...Logger) Logger {
+// Combine creates grouped logger from all provided loggers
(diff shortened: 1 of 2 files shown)


pkg/tools/log/grouplogger.go
```diff
@@ -23,7 +23,8 @@ type groupLogger struct {
 	loggers []Logger
 }
 
-func newGroup(s ...Logger) Logger {
+// New creates grouped logger from all provided loggers
+func New(s ...Logger) Logger {
```
denis-tingaikin, line 27:
```suggestion
func Combine(s ...Logger) Logger {
```

denis-tingaikin (commented):
Added one comment, other things look correct.

pkg/tools/log/grouplogger.go
Mixaster995: ok, fixed

denis-tingaikin (approved):

Issue #1179, opened 2021-11-29
Exclude-prefixes not updated correctly


Mixaster995:
## Expected Behavior
Excluded-prefixes should exactly match prefixes from source file.

## Current Behavior
Excluded-prefixes only appended never removed

## Scenario
1.  Initial request  starts
2. Excluded prefixes server reads prefixes from file and appending it to request's `excluded prefixes` 
3. After initial request `refresh` mechanism uses the same request object(with already added excluded prefixes)
4. Excluded prefixes server reads prefixes from file(which might be modified at this moment) and if it's different set of prefixes than on initial request just appends it to request.
5. Every refresh it repeats, so if file with `excluded prefixes` changes often, it'll just grow excluded prefixes

## Possible solution
Add `client` chain element, and on request it's going to save excluded prefixes added to request by client, and on request response change possibly modified request prefixes back to saved ones.
Also, update client's `cmd`-s to use this chain element

Mixaster995:
@edwarnicke could you, please, take a look to see if you okay with suggested solution 
It prepared it in https://github.com/networkservicemesh/sdk/pull/1180

PR #1180, opened 2021-11-29
fixed updates of excluded prefixes

Mixaster995:

## Description
Excluded prefixes never removed, only appended. It's not desired behaviour.

## Issue link
closes https://github.com/networkservicemesh/sdk/issues/1179

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/networkservice/common/excludedprefixes/server.go
```diff
@@ -37,10 +37,11 @@ import (
...
 type excludedPrefixesServer struct {
-	ctx        context.Context
-	prefixPool atomic.Value
-	once       sync.Once
-	configPath string
+	ctx              context.Context
+	prefixPool       atomic.Value
+	once             sync.Once
+	configPath       string
+	previousPrefixes map[string]struct{}
```
denis-tingaikin, line 44: Could you explain why is this not per connection?

Mixaster995: ok, fixed

glazychev-art (approved):

denis-tingaikin (changes requested):
as far as I can see these changes are still **not** resistant to the following scenarios:

1. Client changed his prefixes on refresh.
2. Client's and server's prefixes have a collision on the first request and do not have a collision on the next request. request1: {clientPrefixes: a, serverPrefixes:a, finalPrefixes: a}, Actual: request2: {clientPrefixes: a, serverPrefixes:b, finalPrefixes: b}. 
Expected: request2: {clientPrefixes: a, serverPrefixes:b, finalPrefixes: ab}.

Mixaster995:
@denis-tingaikin reworked some logic, added unit test for describe scenario - please, take a look

pkg/networkservice/common/excludedprefixes/server.go
```diff
@@ -87,15 +88,56 @@ func (eps *excludedPrefixesServer) Request(ctx context.Context, request *network
...
+
+	prefixesInfo.previousDiff = Subtract(clientPrefixes, prefixesFromFile)
+
+	log.FromContext(ctx).
+		WithField("excludedPrefixes", "server").
+		WithField("prefixesFromFile", prefixesFromFile).
+		WithField("previousPrefixesInfo", prefixesInfo).
+		Debugf("adding excluded prefixes to connection")
+
+	store(ctx, prefixesInfo)
```
denis-tingaikin, line 135: it starts to look more and more complicated.

I think we should consider the problem on the client(nsc) side within exludeprefix.Client chain element that will simply translate response prefixes to client prefixes. In this case all applications in the chain of request can always append prefixes and don't care about diff/conflict cases.

denis-tingaikin (changes requested):

pkg/networkservice/common/excludedprefixes/server.go
Mixaster995: ok, it sounds like a good solution, added it

denis-tingaikin:
@Mixaster995 Could you rebase this?

Mixaster995:
@denis-tingaikin ok, done

pkg/networkservice/common/excludedprefixes/utils.go
```diff
@@ -50,3 +55,17 @@ func exclude(source, exclude []string) []string {
...
+	s2copy := make([]string, len(s2))
+
+	copy(s1copy, s1)
+	copy(s2copy, s2)
+
+	sort.Strings(s1copy)
+	sort.Strings(s2copy)
+
+	return reflect.DeepEqual(s1copy, s2copy)
+}
```
denis-tingaikin, line 71: Consider using github.com/google/go-cmp/cmp

We already have dep on it https://github.com/networkservicemesh/sdk/blob/main/go.mod#L17

```suggestion
```

denis-tingaikin (approved):
Looks good. Only one comment to fix and this can be merged.

pkg/networkservice/common/excludedprefixes/utils.go
Mixaster995: @denis-tingaikin ok, done

denis-tingaikin (approved):

Issue #1196, opened 2021-12-09
Advanced healing: complex cases


glazychev-art:
## Overview

Currently healing only works if one of the components has been broken.

We need to consider scenarios when we have 2 or more deaths.
For example:

**2 death cases:**
1. Local NSMgr +
   a.  Local Forwarder
   b.  Local Endpoint
   c.  Remote NSMgr
   d.  Remote Forwarder
   e.  Remote Endpoint
   f.  Registry

denis-tingaikin:
Is this duplicate of https://github.com/networkservicemesh/sdk/issues/899?

@glazychev-art Can we create a root issue for _Advanced healing_?

glazychev-art:
For `sandbox` tests we can iterate over all combinations.
But for `integration tests` it would be too wasteful. I would suggest using some of the most illustrative ones.

**Integration tests list:**
1. Registry + Local Endpoint - basic case for `Registry` and one component
2. Registry + Remote NSMgr - the `path` is longer than in the local case
3. Registry + Remote Forwarder - the `path` is longer than in the local case
4. Local NSMgr + Local Forwarder - if main management resources restarted on the one node 
5. Local NSMgr + Remote NSMgr - all managers restarted at the same time
6. Local Forwarder + Remote Forwarder - all forwarders restarted at the same time

glazychev-art:
@edwarnicke 
What do you think about this list? https://github.com/networkservicemesh/sdk/issues/1196#issuecomment-993238554

edwarnicke:
@glazychev-art This list looks good.  Could we add to it simply deleting the nsm-system namespace and then recreating it? (meaning the clients and endpoints stay around, but everything else restarts?).  This also somewhat simulates an upgrade.

PR #1206, opened 2021-12-28
[qfix] Clear mechanism preferences on heal

glazychev-art:

## Description
We don't need to use old mechanism preferences on heal request

## Issue link
https://github.com/networkservicemesh/sdk/issues/1196

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/begin/event_factory.go
+++ b/pkg/networkservice/common/begin/event_factory.go
@@ -97,6 +97,7 @@ func (f *eventFactoryClient) Request(opts ...Option) <-chan error {
 				defer cancel()
 				_, _ = f.client.Close(ctx, request.GetConnection(), f.opts...)
 				if request.GetConnection() != nil {
+					request.MechanismPreferences = nil
 					request.GetConnection().Mechanism = nil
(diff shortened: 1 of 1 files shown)


pkg/networkservice/common/begin/event_factory.go
```diff
@@ -97,6 +97,7 @@ func (f *eventFactoryClient) Request(opts ...Option) <-chan error {
 				defer cancel()
 				_, _ = f.client.Close(ctx, request.GetConnection(), f.opts...)
 				if request.GetConnection() != nil {
+					request.MechanismPreferences = nil
```
denis-tingaikin, line 100: As I can see we'll miss the [client's mechanism](https://github.com/networkservicemesh/cmd-nsc-init/blob/main/main.go#L186-L188) with this patch.

Am I missing something?

glazychev-art: You are right
This PR is incorrect

PR #1216, opened 2022-01-20
Add policyroute chain element

glazychev-art:

## Description
This PR adds `policyroute` chain element to implement Policy Based Routing.
It will put policies in connection context on the NSE side.

## Issue link
https://github.com/networkservicemesh/api/issues/119

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/policyroute/server.go
+++ b/pkg/networkservice/common/policyroute/server.go
@@ -1,0 +1,79 @@
+// Copyright (c) 2022 Doc.ai and/or its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package policyroute
+
+import (
+	"context"
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+// PolicyRoutesFunc - method for the new policyRoutes getting
+type PolicyRoutesFunc func() []*networkservice.PolicyRoute
+
+type policyrouteServer struct {
+	getPolicies PolicyRoutesFunc
+}
+
+// NewServer creates a NetworkServiceServer that will put the routing policies to connection context
+func NewServer(policyRouteGetter PolicyRoutesFunc) networkservice.NetworkServiceServer {
+	return &policyrouteServer{
+		getPolicies: policyRouteGetter,
+	}
+}
+
+func (p *policyrouteServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
+	conn := request.GetConnection()
+	if conn.GetContext() == nil {
+		conn.Context = &networkservice.ConnectionContext{}
+	}
+	if conn.GetContext().GetIpContext() == nil {
+		conn.GetContext().IpContext = &networkservice.IPContext{}
+	}
+	ipContext := conn.GetContext().GetIpContext()
+
+	// Update policies
+	// Remove old IP addresses
+	policies := p.getPolicies()
+	for _, p := range ipContext.Policies {
+		if p.From != "" {
+			for s := range ipContext.SrcIpAddrs {
+				if ipContext.SrcIpAddrs[s] == p.From {
+					ipContext.SrcIpAddrs = append(ipContext.SrcIpAddrs[:s], ipContext.SrcIpAddrs[s+1:]...)
+					break
+				}
+			}
+		}
+	}
+	// Use new policies
+	ipContext.Policies = policies
+
+	// Add new IP addresses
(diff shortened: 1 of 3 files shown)


pkg/networkservice/common/policyroute/server_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/common/policyroute"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/chain"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/checks/checkrequest"
+	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/metadata"
+)
+
+const defaultPrefixesFileName = "policies.yaml"
+
+func TestNewPoliciesServer(t *testing.T) {
```
denis-tingaikin, line 41: Can we add goleak check?

pkg/networkservice/common/policyroute/server_test.go
```diff
@@ -0,0 +1,169 @@
...
+				IpContext: &networkservice.IPContext{},
+			},
+		},
+	}
+
+	_, err := server.Request(context.Background(), req)
+	require.NoError(t, err)
+}
+
+func TestCheckReloadedPolicies(t *testing.T) {
```
denis-tingaikin, line 84: Can we add goleak check?

pkg/networkservice/common/policyroute/server.go
```diff
@@ -0,0 +1,119 @@
...
+	ipContext := conn.GetContext().GetIpContext()
+
+	// Update policies
+	// Remove old IP addresses
+	policies := p.policies.Load().([]*networkservice.PolicyRoute)
+	for _, p := range ipContext.Policies {
+		if p.From != "" {
+			for s := range ipContext.SrcIpAddrs {
+				if ipContext.SrcIpAddrs[s] == p.From {
+					ipContext.SrcIpAddrs = append(ipContext.SrcIpAddrs[:s], ipContext.SrcIpAddrs[s+1:]...)
```
denis-tingaikin, line 99: Could you explain why do we need to delete the src ip address in this if branch?

pkg/networkservice/common/policyroute/server.go
```diff
@@ -0,0 +1,119 @@
...
+		for update := range updateCh {
+			updatePrefixes(update)
+		}
+	}()
+}
+
+// NewServer creates a NetworkServiceServer that will put the routing policies to connection context
+func NewServer(ctx context.Context, setters ...ServerOption) networkservice.NetworkServiceServer {
+	server := &policyrouteServer{
+		configPath: "/etc/route-policies/config.yaml",
```
edwarnicke, line 68: why are we reading these from a file in the chain element rather than passing them in as an object as an argument to NewServer?

pkg/networkservice/common/policyroute/server.go
glazychev-art: So, we assume that policies can be dynamic. If we on the refresh cycle, we need to update policies and SrcIPAddreses according to new values.  On the previous `Request` we already set SrcIpAddress - https://github.com/networkservicemesh/sdk/pull/1216/files#diff-7db704fbded091b2c2a44d017b420651c29c4c949321bd5903825fc80fde77a2R116
And on `refresh` we need to delete previous IP addresses from the context and then add new ones. Because we have no idea what policy values ​​we have now (they are dynamic).

pkg/networkservice/common/policyroute/server.go
glazychev-art: Because we assume that we can use `ConfigMap` and it can be changed. In this case we `watch` the config file to receive updates in a timely manner.

pkg/networkservice/common/policyroute/server_test.go
glazychev-art: Thanks, fixed

pkg/networkservice/common/policyroute/server.go
edwarnicke: @glazychev-art ConfigMaps should be read in main.go, not in a chain element.  I'm fine with main.go reading a config file... but a chain element shouldn't.

denis-tingaikin: @edwarnicke Configmap can be changed dynamically by the user. So the question, will we ignore the user's config map update?

edwarnicke: @denis-tingaikin If we want to support dynamic updating (which I'm fine with) we should still handle it via an object passed in, embedding config file watching logic inside a chain element ties that chain element to choices around the file system being the source of the config.  Best to avoid that.

denis-tingaikin: Agreed. Can we then pass _channel_ with policies into the chain element?

edwarnicke: @denis-tingaikin That's certainly one option... but we need to be careful to correctly handle cases where we don't receive policyies etc... blocking isn't OK in those cases, but errors would be.

denis-tingaikin: @edwarnicke Hmm, I think it makes sense, working with a channel can produce a complex logic in the chain element.

So as an alternative option we can provide some interface for getting actual PBRs

```go
type Provider interface {
 Get() []*networkservice.PolicyRoute
}
```

Thoughts?

edwarnicke: Or even simpler:

```go
type PolicyRoutesFunc func() []*networkpolicy.PolicyRoute
```

denis-tingaikin (approved):

denis-tingaikin:
@edwarnicke Let us know if you still have any additional comments then we'll consider them in separate PR.

PR #1224, opened 2022-02-04
Add extension for sync.Once

glazychev-art:

## Description
Adds `once UntilSuccsess` , that we can call until the first success 

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


denis-tingaikin:
@edwarnicke PR is not ready to merge, but be free to add comments

pkg/tools/once/once.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Do will be callable until the first success
+func (u *UntilSuccess) Do(f func() bool) {
+	if atomic.LoadUint32(&u.done) > 0 {
+		return
+	}
+	u.m.Lock()
+	defer u.m.Unlock()
+	if u.done == 0 && f() {
```
ljkiraly, line 38: Very interesting and good initiative. I had similar doubt when I checked this ['Do' function ](https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/sync/once.go;l=48), it doesn't check the return value of the init function.
I suggest using atomic call to check the 'u.done' value:
`if atomic.LoadUint32(&u.done) == 0 && f() {`

Issue #1231, opened 2022-02-21
Begin merge connection strategy


LionelJouin:
## Overview

Together with the new [source routing feature](https://github.com/networkservicemesh/api/issues/119), we would like our NSCs to have the ability to re-configure their connections on the fly, so without having to close, and re-request (mostly to update the IP context). But currently, the begin client chain component is preventing this due a merge function supporting only the extracontext for now. 

Here is the [link to the function](https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/begin/merge.go#L39), my proposal would be to not merge the ip contexts and only keep the requested one.

## References

https://github.com/networkservicemesh/api/issues/119
https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/begin/merge.go#L39

denis-tingaikin:
Hmm. Can this be related to https://github.com/networkservicemesh/cmd-forwarder-vpp/issues/441?

LionelJouin:
yes, from what I understood it seems related. So far, the only blocking element is the begin chain component. I tried with the changes in my environment and it worked fine (my NSC re-request with updated source IPs and policies)

LionelJouin:
Solved: https://github.com/networkservicemesh/sdk/pull/1238

PR #1234, opened 2022-02-23
IPContext merged in begin chain element (#1231)

LionelJouin:
## Description
Using the Client chain, the network service client cannot update the IPContext without the `mergeConnectionContext` function implemented. A NSC might want to update it if it needs to add, update or delete routes, src IP addresses (VIPs...), policies...

## Issue link
https://github.com/networkservicemesh/sdk/issues/1231

## How Has This Been Tested?
- [x] Tested manually
- [ ] Have not tested

I am not sure what would be the full impact in NSM, but in my environment, I managed to update IPs and policies without any issue.

## Types of changes
- [x] Bug fix
- [x] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/begin/merge.go
+++ b/pkg/networkservice/common/begin/merge.go
@@ -1,4 +1,5 @@
-// Copyright (c) 2021 Cisco and/or its affiliates.
+// Copyright (c) 2021-2022 Cisco and/or its affiliates.
+// Copyright (c) 2022 Nordix Foundation
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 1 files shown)


denis-tingaikin:
Also, I feel we need to add a unit test for this case. See at unit tests for begin pkg https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/begin/serialize_both_test.go

pkg/networkservice/common/begin/merge.go
```diff
@@ -36,7 +37,8 @@ func mergeConnection(returnedConnection, requestedConnection, connection *networ
 func mergeConnectionContext(returnedConnectionContext, requestedConnectionContext, connectioncontext *networkservice.ConnectionContext) *networkservice.ConnectionContext {
 	rv := proto.Clone(connectioncontext).(*networkservice.ConnectionContext)
 	if !proto.Equal(returnedConnectionContext, requestedConnectionContext) {
-		// TODO: IPContext, DNSContext, EthernetContext, do we need to do MTU?
+		// TODO: DNSContext, EthernetContext, do we need to do MTU?
+		rv.IpContext = requestedConnectionContext.IpContext
```
denis-tingaikin, line 41: Hmm, I think we also need to do it for other context as you pointed in the comment.


@edwarnicke Thoughts?

edwarnicke: @LionelJouin Apologies for this taking so long to get back to you on. The net net is that what you did here was fine.  I also realized why it took me so long to sort it out, and produced a better documented version here in #1237 that handles the merge more or less correctly (which you were doing) and also in a way that's hopefully clearer.  I also provided testing.  Could you have a look?

LionelJouin: no problem, I will try it today, thank you

Issue #1187, opened 2021-12-02
Advanced healing: datapath was down


glazychev-art:
## Overview

Let's consider a case when our control plane works fine - we established a `connection`, requests from `nsc` to `nse` pass with no errors. With `refreshes`, we just maintain this connection. 

But what if the _**data plane**_ was down - for example, parameters for routing were changed accidentally, or the interface is frozen or dead.
The current healing will not be able to fix this, because the `connection` has been established and there are no problems with further `requests`.

Requests:
```
┌─────────┐         ┌─────────┐         ┌─────────┐        ┌─────────┐
│         ├────────►│         ├────────►│         ├───────►│         │
│   NSC   │         │  NSMGR  │         │   FWD   │        │   NSE   │
│         │◄────────┤         │◄────────┤         │◄───────┤         │
└─────────┘         └─────────┘         └─────────┘        └─────────┘
```
Ping:
```
 ┌─────────┐        ┌─────────┐        ┌─────────┐
 │         │        │         │        │         │
 │   NSC   ├───────►│   FWD   ├──►X    │   NSE   │
 │         │        │         │        │         │
 └─────────┘        └─────────┘        └─────────┘
```

### 1. Use `WithLivelinessCheck` option for `healClient`

https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/heal/options.go#L34

On the client side we already have all of the parameters (`ip-addresses`, `routes`), with which we can check the status of the connection.
With this option we can create a goroutine, that will periodically send an `icmp` packet to the endpoint. 
For example, if `nsc` has:
- `kernel` interface - we can try to use `icmp` package https://pkg.go.dev/golang.org/x/net/icmp or https://github.com/go-ping/ping
- `memif` interface - we need to implement a new `Checker`, that will run `ping` from `vpp`.

If ping fail - send a faulty connection to the channel, that will be processed by `healClient` like an event from `MonitorConnections` ([Link](https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/heal/eventloop.go#L95)).
`healClient` will start its work and call `Close` and `reRequest`.

**Example:**
Let's take a look at local `Kernel2Kernel` scenario.
**_Steps:_**
1. Create NSC with `healClient` option:
```
WithLivelinessCheck(func(conn *networkservice.Connection) bool {
 // ping(conn.IpContext.DstIPAddr)
 // if !ping {
 //      return false
 // }
 // return true
})
```
[...]

PR #1259, opened 2022-04-06
call livenessCheck periodically

xzfc:
## Description

Check data-plane aliveness periodically, not only after a control-plane event happens

Continuation of PR #1252.

## Issue link

#1187

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/networkservice/common/heal/client.go
```diff
@@ -31,21 +32,26 @@ import (
...
+	livenessCheckInterval time.Duration
 }
 
 // NewClient - returns a new heal client chain element
 func NewClient(chainCtx context.Context, opts ...Option) networkservice.NetworkServiceClient {
-	o := &option{
-		livelinessCheck: func(_ *networkservice.Connection) bool { return false },
+	o := &options{
+		attemptAfter:          time.Millisecond * 200,
+		livenessCheckInterval: time.Millisecond * 500,
```
edwarnicke, line 45: Out of curiosity, why these default values?

xzfc: * `attemptAfter` is the same as before this PR.

* About `livenessCheckInterval`. In the latest revision of the PR, this value was removed; the user should provide its own timer instead in the `LivenessChecker` implementation.

  In the previous version of this PR (the one that you commented), `livenessCheckInterval` had roughly the same order of magnitude as the default `ping` utility interval. Also, it was tied to [these values](https://github.com/Chunosov/cmd-nsc/blob/6fca644fa41fe5a8d533375c088a1b010d8b7ee6/main.go#L74).

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -32,18 +31,19 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/tools/log"
 )
 
-const attemptAfter = time.Millisecond * 200
+const maxAttemptsWithoutReselect = 3
```
denis-tingaikin, line 34: @xzfc Could you say more why do we need this?

xzfc: I've added it as a workaround to [this issue](https://github.com/networkservicemesh/sdk/issues/1187#issuecomment-1002990951) (NSE pod was deleted, but it still was responding to pings), not knowing that this was fixed in https://github.com/networkservicemesh/sdk-vpp/pull/537.

However, I think it still makes sense to perform a forceful reselect after a few failed requests to the same endpoint, even if the data plane is still alive.

denis-tingaikin: We should keep requests without reselect while datapath is fine.

xzfc: Removed `maxAttemptsWithoutReselect`.

pkg/networkservice/common/heal/client.go
edwarnicke: @xzfc Are we using the 'ping' command, or are we sending the icmp echo messages directly from Go?

xzfc: We're sending ICMP messages directly, see https://github.com/networkservicemesh/sdk-kernel/pull/447.

edwarnicke: @xzfc Woot!

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -78,55 +76,113 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
+	return res
+}
+
+func (cev *eventLoop) eventLoop() {
+	reselect := false
+
+	ctrlPlaneCh := cev.waitCtrlPlaneEvent()
+
+	livenessCheckerCtx, livenessCheckerCancel := context.WithCancel(context.Background())
+	dataPlaneCh := make(chan struct{}, 1)
```
denis-tingaikin, line 136:
```suggestion
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -78,55 +76,113 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
+	defer livenessCheckerCancel()
+	if cev.heal.livenessChecker != nil {
+		go func() {
+			cev.heal.livenessChecker(livenessCheckerCtx, cev.conn)
+			dataPlaneCh <- struct{}{}
```
denis-tingaikin, line 141:
```suggestion
			livenessCheckerCancel()
```

pkg/networkservice/common/heal/client.go
```diff
@@ -32,20 +33,22 @@ import (
 
 type healClient struct {
 	chainCtx        context.Context
-	livelinessCheck LivelinessCheck
+	livenessChecker LivenessChecker
+	attemptAfter    time.Duration
```
denis-tingaikin, line 37: Why do we need this?

pkg/networkservice/common/heal/client.go
```diff
@@ -32,20 +33,22 @@ import (
...
 }
 
 // NewClient - returns a new heal client chain element
 func NewClient(chainCtx context.Context, opts ...Option) networkservice.NetworkServiceClient {
-	o := &option{
-		livelinessCheck: func(_ *networkservice.Connection) bool { return false },
+	o := &options{
+		attemptAfter: time.Millisecond * 200,
```
denis-tingaikin, line 43:
```suggestion
```

pkg/networkservice/common/heal/client.go
```diff
@@ -32,20 +33,22 @@ import (
...
 	}
 	for _, opt := range opts {
 		opt(o)
 	}
 	return &healClient{
 		chainCtx:        chainCtx,
-		livelinessCheck: o.livelinessCheck,
+		livenessChecker: o.livenessChecker,
+		attemptAfter:    o.attemptAfter,
```
denis-tingaikin, line 51:
```suggestion
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -78,55 +76,113 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
+
 	/* Attempts to heal the connection */
-	afterTicker := clock.FromContext(cev.eventLoopCtx).Ticker(attemptAfter)
+	afterTicker := clock.FromContext(cev.eventLoopCtx).Ticker(cev.heal.attemptAfter)
 	defer afterTicker.Stop()
 	for {
 		select {
 		case <-cev.chainCtx.Done():
 			return
+		case <-dataPlaneCh:
```
denis-tingaikin, line 173: This confuses. Sometimes it calls dataPlane, sometimes it calls as livenesscheker. Can we follow one style naming?

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -78,55 +76,113 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
+			cev.logger.Warnf("Data plane is down")
+			reselect = true
 		case <-afterTicker.C():
```
denis-tingaikin, line 176: What if request try was too long and the ticker channel is not empty?

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -78,55 +76,113 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
 }
 
-func (cev *eventLoop) eventLoop() {
-	/* Receive monitor events */
-	for {
-		eventIn, err := cev.client.Recv()
-		if cev.eventLoopCtx.Err() != nil {
-			return
-		}
+func (cev *eventLoop) waitCtrlPlaneEvent() <-chan struct{} {
```
denis-tingaikin, line 95: consider better naming

pkg/networkservice/common/heal/options.go
```diff
@@ -17,22 +17,33 @@
...
+	"context"
+	"time"
+
 	"github.com/networkservicemesh/api/pkg/api/networkservice"
 )
 
-// LivelinessCheck - function that returns true of conn is 'live' and false otherwise
-type LivelinessCheck func(conn *networkservice.Connection) bool
+// LivenessChecker blocks until the connection is alive or ctx is done.
+type LivenessChecker func(ctx context.Context, conn *networkservice.Connection)
```
denis-tingaikin, line 27: @edwarnicke Is this API looking good for you?

denis-tingaikin (commented):
In general looks OK. Added a few comments related to style and pointed at potential attemptAfter leak

pkg/networkservice/common/heal/client.go
xzfc: To make sandbox tests run faster. In the latest revision of this PR replaced with the `BackOff` interface.

pkg/networkservice/common/heal/eventloop.go
xzfc: Renamed `livenessChecker` to `dataPlaneLivenessChecker`.

pkg/networkservice/common/heal/eventloop.go
xzfc: Added `cev.chainCtx.Err() != nil` check. Also replaced `case <-dataPlaneCh` with `dataPlaneCtx.Err() != nil` check.

pkg/networkservice/common/heal/eventloop.go
xzfc: Renamed to `monitorCtrlPlane()`.

glazychev-art:
See - https://github.com/networkservicemesh/sdk/pull/1271

PR #1260, opened 2022-04-06
Run golangci-lint with GOOS=linux/darwin/windows

xzfc:
## Description
Currently, our CI doesn't check non-linux platform-specific code. Checking these files may be useful for those who use non-Linux OS as their dev environment.

This PR enables checks for Windows-only and macOS-only files by running `GOOS=windows golangci-lint` and `GOOS=darwin golangci-lint` on Ubuntu.

An alternative option is to run this action on `windows-latest` and `macos-lates` natively as proposed in the [doc](https://github.com/golangci/golangci-lint-action#multiple-os-support), but it fails on Windows due to the path separators difference in `exclude-rules` in `.golangci.yml` (see https://github.com/golangci/golangci-lint/issues/1398).

So, this PR solves the issue completely for darwin (i.e. you could run `golangci-lint` on macOS without errors), and partially for Windows (i.e. Windows-specific files like [grpc_utils_windows.go](https://github.com/networkservicemesh/sdk/blob/main/pkg/tools/sandbox/grpc_utils_windows.go) are linted, but `golangci-lint` is still not usable on Windows).

---

Also, this PR removes an unused function `GetInode()`. The reason is that the linter complains about the Windows implementation of this function.
```
pkg/tools/fs/inode_windows_amd64.go:38:17: G103: Use of unsafe calls should be audited (gosec)
	ptr := uintptr(unsafe.Pointer(stat))
```

As I can see in the code, it casts a pointer to some internal data structure into an integer and treats it as an inode number. I don't think this would produce any meaningful result. And I am not sure how to properly fix this function on Windows since there are no inode numbers in Windows. But since this function isn't used anywhere except an archived repo `cmd-nsc-proxy`, I've deleted it.

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [x] CI


denis-tingaikin (approved):

denis-tingaikin:
@xzfc Should we consider non-linux linter running?

xzfc:
I'd think it would be an overkill.

An option to lint all files regardless of the platform was proposed in https://github.com/golangci/golangci-lint/issues/1646, but not implemented.
Other options are to run golangci-lint multiple times with different flags (e.g. `golangci-lint --build-tags linux && golangci-lint --build-tags windows`) or to introduce an additional `all` tag to all platform-specific code (e.g. `// +build linux all`).

denis-tingaikin:
>(e.g. golangci-lint --build-tags linux && golangci-lint --build-tags windows) or to introduce an additional all tag to all platform-specific code (e.g. // +build linux all).

Could you please add a step for golangci job with non linux tag?

xzfc:
> Could you please add a step for golangci job with non linux tag?

Done, see the updated PR description.

PR #1278, opened 2022-05-05
Reusable workflows

NikitaSkrynnik:
## Issue link

https://github.com/networkservicemesh/.github/issues/4

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [x] CI


denis-tingaikin (approved):
LGTM 🚀

.github/workflows/automerge.yaml
```diff
@@ -8,23 +8,5 @@ on:
...
-          find . -type f ! -name 'go.mod' ! -name 'go.sum' -exec git diff --exit-code origin/main -- {} +
-      - name: Merge PR
-        uses: ridedott/merge-me-action@master
-        with:
-          GITHUB_LOGIN: nsmbot
-          ENABLED_FOR_MANUAL_CHANGES: true
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          MAXIMUM_RETRIES: 25
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.actor == 'nsmbot'}}
+    uses: networkservicemesh/.github/.github/workflows/automerge.yaml@main
```
edwarnicke, line 12: Do we want to depend on a branch here rather than a specific point?

I can see tradeoffs either way... so the real question is 'have we thought it through' ?

edwarnicke:
Overall other than my branch vs commit comment above, I'm good with this :)

.github/workflows/automerge.yaml
denis-tingaikin: I see next pros/cons

pros:
- we're always up to date.

cons:
- ci for previous could be unstable.

In my mind the best option is: 
 1. use @main version by default.
 2. change @main version to point on the release. 
 
  think we can address this in the v1.5.0.

Thoughts?

edwarnicke: Sounds good :)

denis-tingaikin: @edwarnicke I've created a ticked for v1.5.0 https://github.com/networkservicemesh/.github/issues/6

PR #1292, opened 2022-05-15
Fix backward compatibility for NewNetworkServiceEndpointRegistryClient and NewNetworkServiceRegistryClient

edwarnicke:
NewNetworkServiceEndpointRegistryClient and NewNetworkServiceRegistryClient where recently enhanced.
In that process, the leading "connectTo *url.URL" parameter was removed.
The functionality continues to be supported via the WithClientURL option.

In order to avoid breaking current downstream users, the leading "connectTo *url.URL"
has been restored in such a way that it has no effect if it is nil and otherwise
utilize the WithClient option (to insure the code path is exercised).

Signed-off-by: Ed Warnicke <hagbard@gmail.com>

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


denis-tingaikin (commented):
@edwarnicke We've already updated most of our applications, at this moment URL is **_optional_** parameter because can be use dns url resolution.


Can we discuss this change on community call? I think it would be good option if we can update upstream repositories with new option `WithClientUrl(...)`

edwarnicke:
@denis-tingaikin Totally fine to discuss on the community call :)  My concern is wrt folks who have their own NSEs and NSCs and the change this would mean for them.

Issue #1311, opened 2022-06-15
Concurrency issue causing the NSMgr to crash


LionelJouin:
## Current Behavior

The NSMgr crashes during the deployment of NSCs and NSEs, not sure what is the exact cause and how we could reproduce, but I believe from the logs it is due to this part: https://github.com/networkservicemesh/sdk/blob/v1.4.0/pkg/networkservice/common/metrics/server.go#L79.  
I included logs below.

## Failure Information (for bugs)

```
2022-06-15T12:46:13.045665873Z stderr F fatal error: concurrent map read and map write
2022-06-15T12:46:13.048873743Z stderr F 
2022-06-15T12:46:13.048940244Z stderr F goroutine 886 [running]:
2022-06-15T12:46:13.048945407Z stderr F runtime.throw({0xd2ef0d?, 0x4af46f?})
2022-06-15T12:46:13.048960268Z stderr F 	/usr/local/go/src/runtime/panic.go:992 +0x71 fp=0xc0008f5ea8 sp=0xc0008f5e78 pc=0x4359b1
2022-06-15T12:46:13.048971173Z stderr F runtime.mapaccess2_faststr(0x133eb00?, 0x1?, {0xc0003da7f8, 0x11})
2022-06-15T12:46:13.048986036Z stderr F 	/usr/local/go/src/runtime/map_faststr.go:117 +0x3d4 fp=0xc0008f5f10 sp=0xc0008f5ea8 pc=0x412d54
2022-06-15T12:46:13.04899616Z stderr F github.com/networkservicemesh/sdk/pkg/networkservice/common/metrics.(*metricServer).writeMetrics(0xc00000ddd0, {0xe44880, 0xc0004b9a40}, 0xc00075b090)
2022-06-15T12:46:13.049000884Z stderr F 	/go/pkg/mod/github.com/networkservicemesh/sdk@v1.4.0/pkg/networkservice/common/metrics/server.go:79 +0x199 fp=0xc0008f60d0 sp=0xc0008f5f10 pc=0xb1a339
2022-06-15T12:46:13.049004563Z stderr F github.com/networkservicemesh/sdk/pkg/networkservice/common/metrics.(*metricServer).Request(0xe44880?, {0xe44880, 0xc0004b9a40}, 0xc0007dc410?)
[...]

PR #1312, opened 2022-06-16
Fix metrics concurrency

glazychev-art:

## Description

The metrics map should be unique for each connection and shouldn't cause concurrency issues.
In this PR the map has been moved from the chain element to the metadata.

## Issue link
Closes: https://github.com/networkservicemesh/sdk/issues/1311

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/metrics/metadata.go
+++ b/pkg/networkservice/common/metrics/metadata.go
@@ -25,7 +25,7 @@ import (
 )
 
 type keyType struct{}
-type metricsMap map[string]metric.Int64Histogram
+type metricsMap = map[string]metric.Int64Histogram
 
 func loadOrStore(ctx context.Context, metrics metricsMap) (value metricsMap, ok bool) {
 	rawValue, ok := metadata.Map(ctx, false).LoadOrStore(keyType{}, metrics)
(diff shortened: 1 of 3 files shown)


denis-tingaikin (commented):
Can we somehow cover the problyem by unit test?

pkg/networkservice/common/metrics/metadata.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/metadata"
+)
+
+type keyType struct{}
+type metricsMap map[string]metric.Int64Histogram
```
denis-tingaikin, line 28: I think we could use this as alias
```suggestion
type metricsMap = map[string]metric.Int64Histogram
```

glazychev-art: thanks, fixed

PR #1318, opened 2022-07-05
Remove duplicate dns configs in response connection in DNSContextClient

NikitaSkrynnik:
Signed-off-by: Nikita Skrynnik <nikita.skrynnik@xored.com>

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/connectioncontext/dnscontext/client.go
+++ b/pkg/networkservice/connectioncontext/dnscontext/client.go
@@ -172,8 +172,10 @@ func removeDuplicates(elements []*networkservice.DNSConfig) []*networkservice.DN
 	for _, e := range elements {
 		exists := false
 		for _, v := range result {
-			if equal(e.DnsServerIps, v.DnsServerIps) && equal(e.SearchDomains, v.SearchDomains) {
+			if intersect(e.DnsServerIps, v.DnsServerIps) && intersect(e.SearchDomains, v.SearchDomains) {
 				exists = true
+				v.DnsServerIps = merge(v.DnsServerIps, e.DnsServerIps)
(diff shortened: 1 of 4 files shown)


denis-tingaikin (approved):

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -169,3 +165,45 @@ func (c *dnsContextClient) initialize() {
...
+				break
+			}
+		}
+		if !exists {
+			result = append(result, e)
+		}
+	}
+
+	return result
+}
```
denis-tingaikin, line 186: What if we have two dns configs with different len but have that have intersection? Please handle correctly corner cases.

denis-tingaikin (changes requested):
Please fix the comment above.

anastasia-malysheva (approved):

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -169,3 +165,56 @@ func (c *dnsContextClient) initialize() {
...
+	}
+
+	for _, v := range b {
+		if _, ok := common[v]; !ok {
+			result = append(result, v)
+		}
+	}
+
+	return result
+}
```
denis-tingaikin, line 220: We are going a bit over-complicated here. Moreover, am not sure about merging dns configs because it could produce a ton of new unexpected states. Thus I think we need to consider about using metadata for server dnscontext.

denis-tingaikin (changes requested):

PR #1328, opened 2022-07-19
Add replaceNSEName chain element

glazychev-art:

## Description
Split `replacelabels` chain element to `replacelabels` and `replaceNSEName`
We need to use already discovered NSE on refresh

## Issue link
https://github.com/networkservicemesh/deployments-k8s/issues/6305

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/networkservice/common/replacelabels/client.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Doc.ai and/or its affiliates.
```
denis-tingaikin, line 1: Should we return this?

pkg/networkservice/chains/nsmgr/suite_test.go
```diff
@@ -795,6 +796,7 @@ func additionalFunctionalityChain(ctx context.Context, clientURL *url.URL, clien
 					client.WithAdditionalFunctionality(
 						mechanismtranslation.NewClient(),
 						replacelabels.NewClient(labels),
+						replacensename.NewClient(),
```
denis-tingaikin, line 799: Im mostly sure that we could face with the similar issue in future (when we need to replace something in pass through requests)

Should we create a pkg like `passthrough` and concatenate all needed  replace elements?

Something like

```go
package passthrough

func NewClient(labels map[string]string) {
return chain.NewClient(
replacelabels.NewClient(labels),
replacensename.NewClient(),
)

}
``` 

And use `passthrough` chain in firewall nses?


/cc @edwarnicke

glazychev-art: It's a good idea, thanks! 
Done

pkg/networkservice/common/replacelabels/client.go
glazychev-art: Yes, done

denis-tingaikin (approved):

Issue #1339, opened 2022-08-15
dns: dnscontext doesn't restore `resolv.conf`


glazychev-art:
## Current Behavior

dnscontext doesn't restore `resolv.conf` after restarts

## Failure Information (for bugs)

If the NSE restarts dnscontext cannot find previous `resolv.conf` (/etc/resolv.conf.restore). But the current `resolv.conf` is also invalid - it contains already modified values (`nameserver 127.0.0.1 ...`)

## Steps to Reproduce

Integration tests:

1. Add logs to an NSC to print the current `resolv.conf` - https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/connectioncontext/dnscontext/client.go#L123
2. Delete `nsmgr` from from the nsm-system kustomization file
3. Run any of integration tests (Kernel2Kernel)
4. `kubectl get logs {nsc_name}`
5. See that the first output is fine, but the second (after restart) contains invalid values (already modified).

## Context

* Kubernetes Version:
* etc.

denis-tingaikin:
Seems like fixed.

PR #1342, opened 2022-08-17
DNS: dnscontext doesn't restore resolv.conf

NikitaSkrynnik:
## Issue link

https://github.com/networkservicemesh/sdk/issues/1339

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/connectioncontext/dnscontext/client.go
+++ b/pkg/networkservice/connectioncontext/dnscontext/client.go
@@ -20,7 +20,6 @@ package dnscontext
 
 import (
 	"context"
-	"errors"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -33,13 +33,16 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/tools/log"
 )
 
+const (
+	modifiedResolvConfKey = "# modified"
```
denis-tingaikin, line 37: Can we simply try to parse resolv.conf from the comments? 

If the comments can be parsed as resolv.conf then we are using it. Otherwise we may think that the resolv.conf was not stored and we may do store operation.

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -96,22 +96,51 @@ func (c *dnsContextClient) Close(ctx context.Context, conn *networkservice.Conne
...
+		return
+	}
+
+	_ = os.WriteFile(c.resolveConfigPath, []byte(commentedPart), os.ModePerm)
 }
 
 func (c *dnsContextClient) storeOriginalResolvConf() {
-	if _, err := os.Stat(c.storedResolvConfigPath); err == nil {
+	bytes, err := os.ReadFile(c.resolveConfigPath)
+	originalResolvConfig := string(bytes)
```
glazychev-art, line 123: It will be better to check for error before `originalResolvConfig := string(bytes)`

pkg/networkservice/connectioncontext/dnscontext/resolvconf_test.go
```diff
@@ -71,11 +70,8 @@ options ndots:5`
 	config.SetValue(searchProperty, "default.svc.cluster.local", "svc.cluster.local", "cluster.local")
 	config.SetValue(optionsProperty, "ndots:5")
 	config.SetValue("my_property")
-	err = config.Save()
-	require.Nil(t, err)
-	bytes, err := os.ReadFile(filepath.Clean(p))
-	require.Nil(t, err)
-	actual := string(bytes)
+	serialized := config.Serialize()
+	actual := serialized
```
glazychev-art, line 74: I think we can use here only one variable: `serialized` or `actual`

pkg/networkservice/connectioncontext/dnscontext/client.go
```diff
@@ -96,22 +96,51 @@ func (c *dnsContextClient) Close(ctx context.Context, conn *networkservice.Conne
...
+	for i := range lines {
+		lines[i] = "#" + lines[i]
+	}
+
+	_ = os.WriteFile(c.resolveConfigPath, []byte(strings.Join(lines, "\n")+"\n\n"), os.ModePerm)
+}
+
+func (c *dnsContextClient) appendResolvConf(resolvConf string) error {
+	bytes, err := os.ReadFile(c.resolveConfigPath)
+	originalResolvConfig := string(bytes)
```
glazychev-art, line 138: It will be better to check for error before `originalResolvConfig := string(bytes)`

pkg/networkservice/connectioncontext/dnscontext/resolvconf.go
```diff
@@ -69,8 +71,8 @@ func (r *resolveConfig) SetValue(k string, values ...string) {
 	}
 }
 
-// Save saves resolve config file
-func (r *resolveConfig) Save() error {
+// Serialize serializes resolve config
+func (r *resolveConfig) Serialize() string {
```
glazychev-art, line 75: Perhaps this method should be called `String()`?

denis-tingaikin (approved):

Issue #1272, opened 2022-04-27
NSM logging system is complicated and buggy


denis-tingaikin:
## Description

During fixing issue reported by @richardstone23 I found a few issues in our current NSM logging system

1. We are using TracingEnabled(true) in each cmd application, when we can check is the tracing enabe by the grpc options.
2. Even if loglevel != TRACE && TracingEnabled(false) then trace chain element calculates diff and uses CPU for nothing.  
3. TracingEnabled(false)/TracingEnabled(true) is complicated and confuses.
4. Don’t depend on logrus
5. https://github.com/networkservicemesh/sdk/issues/1364

## Solution 

Rework and simplify the logging system.

uablrek:
And remove any hard-coded reference to logrus in the interface.

It should be possible to implement a custom logger from the [Logger interface](https://github.com/networkservicemesh/sdk/blob/main/pkg/tools/log/logger.go) that isn't logrus.

Now that is possible (it seems) except that you must set the `logrus.TraceLevel` even if logrus isn't used. It is awkward but no big deal really. It seems that the proposed "simplifications" makes even more hard-coded references to logrus.

Logrus is in maintenace mode for the reason below. I want to use [Zap](https://github.com/uber-go/zap) as logger.

> I believe Logrus' biggest contribution is to have played a part in today's widespread use of structured logging in Golang. There doesn't seem to be a reason to do a major, breaking iteration into Logrus V2, since the fantastic Go community has built those independently. Many fantastic alternatives have sprung up. Logrus would look like those, had it been re-designed with what we know about structured logging in Go today. Check out, for example, [Zerolog](https://github.com/rs/zerolog), [Zap](https://github.com/uber-go/zap), and [Apex](https://github.com/apex/log).

uablrek:
> Now that is possible (it seems) ...

Nope. NSM doesn't care about the logger set in the context with 
```
log.WithLog(ctx, log.NSMLogger(logger)) // allow NSM logs
```

uablrek:
I guess the problem above is addressed by the first items in the proposed solutio;
> 1.Use logger from the chain context by trace server chain elements.
2.Remove Default logger

uablrek:
IMO a good test (necessary actually) to verify that an abstract log interface really is usable for may implementations is to try at least two, e.g. logrus and Zap. I include my Zap based logger below. We use [go-logr/logr](https://github.com/go-logr/logr) with Zap as backend.

[...]

PR #1348, opened 2022-09-08
Settable global logger

uablrek:
## Description

It should be possible to set an own customizwd Logger to be used by NSM. Now the logger is hard-coded in some places, most notable the `logruslogger.FromSpan()` used for trace logging. The intention for this PR is to fix the most urgent needs in a backward compatible way with as little changes as possible.

This PR adds a global logger reachable with `log.L()` (same as [zap](https://github.com/uber-go/zap)) and settable with `log.SetGlobalLogger(l Logger)`.

`logruslogger.FromSpan()` has been altered to become generic by using the `log.Logger` interface rather than a variation of `logruslogger`. It uses the global logger as base (calls `log.L()`) and thus allows users to specify the logger used for trace. ~For backward compatibility the global Logger is set to a `logruslogger` on init.~ If no logger is specified (logger is log.Default()) a backward compatible logger (logrus) is used.

The global logger should be set early in `main()`. Example;
```go
	if config.LogLevel == "TRACE" {
		nsmlog.EnableTracing(true)
		// Work-around for hard-coded logrus dependency in NSM
		logrus.SetLevel(logrus.TraceLevel)
	}
	logger.Info("NSM trace", "enabled", nsmlog.IsTracingEnabled())
	nsmlogger := log.NSMLogger(logger)
	nsmlog.SetGlobalLogger(nsmlogger)
	ctx = nsmlog.WithLog(ctx, nsmlogger)
```

#### Tasks that should be done but are not backward compatible

* Move `FromSpan()` from package `logruslogger` to `log` (since it's not logrus specific any more)
* Rename `FromSpan()` -> `NewTraceLogger()`

I will not add these things since it affects code all over the place I guess. Also `FromSpan()` is a horrible hack IMHO and should be re-written.

## Issue link

https://github.com/networkservicemesh/sdk/issues/1272

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

I have used this PR in our NSM application with trace and it works fine. The example above is from real code. I have also used a `logruslogger/logruslogger_test.go` unit test to exercise the code in a convenient way. I have not included it since it doesn't really make any automatic tests, it just emits logs in various ways to be eyeballed by the developer.

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/log/defaultlogger.go
+++ b/pkg/tools/log/defaultlogger.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Doc.ai and/or its affiliates.
+// Copyright (c) 2021-2022 Doc.ai and/or its affiliates.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -30,6 +30,12 @@ type defaultLogger struct {
 // Default - provides a default logger
 func Default() Logger {
 	return &defaultLogger{}
+}
+
+// IsDefault - true if the logger is a default logger
+func IsDefault(logger Logger) bool {
+	_, ok := logger.(*defaultLogger)
+	return ok
 }
 
 func (l *defaultLogger) Info(v ...interface{}) {
--- a/pkg/tools/log/logger.go
+++ b/pkg/tools/log/logger.go
@@ -31,7 +31,8 @@ const (
 )
 
 var (
-	isTracingEnabled int32 = 0
+	isTracingEnabled int32  = 0
+	globalLogger     Logger = Default()
 )
 
 // Logger - unified interface for logging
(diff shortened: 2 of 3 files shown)


uablrek:
Example with a custom logger;
```json
{
  "severity": "debug",
  "timestamp": "2022-09-08T18:13:49.429+00:00",
  "service_id": "Meridio-LB",
  "message": "(7)       ⎆ sdk/pkg/registry/core/trace/traceNetworkServiceEndpointRegistryFindClient.Recv()",
  "version": "1.0.0",
  "extra_data": {
    "subsystem": "NSM",
    "type": "registry"
  }
}
```

uablrek:
About failed CI;
* I can't remove the `init()` function without breaking backward compatibility
* The `FromContext()` now returns `log.L()` if there is no log in the contexts. IMHO this is the right way, but apparetly it breaks CI. I can change it back

The others seem to be minor updates. I can fix those.

edwarnicke:
@denis-tingaikin @edwarnicke

denis-tingaikin:
@uablrek There is also failed build/tests. Could you have a look?

denis-tingaikin:
I'm just interested in why we can't just declare a global context and put there a logger. 

@uablrek Could you remind why we can't use a global context?

uablrek:
> Could you remind me why we can't use a global context?

Not sure, but nobody else does that AFAIK so I suppose there are reasons. It is recommended to pass context as the first parameter (not using a global context) and context's are more often derived than a logger.

edwarnicke (approved):

PR #1353, opened 2022-09-13
Don't trace gRPC "Canceled" errors

caboteria:
## Description

We were getting a *lot* of log chaff at TRACE level during normal operation because gRPC returns a "context canceled" error periodically but it's not really an error for us because we immediately loop around again.

This commit uses the gRPC status.Code() method to identify these non-errors. It also uses errors.Is() instead of == to compare error objects. This handles wrapped errors.

https://pkg.go.dev/google.golang.org/grpc/status#Code

https://github.com/golang/go/wiki/ErrorValueFAQ#how-should-i-change-my-error-handling-code-to-work-with-the-new-features

## Issue link

https://github.com/networkservicemesh/sdk/issues/1068

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

Deploy cmd-nse-icmp-responder, see periodic "context cancelled" errors, deploy build with patched SDK, verify that errors aren't present.

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

Only relevant when running at TRACE log level.

### DIFF
--- a/pkg/registry/core/trace/ns_registry.go
+++ b/pkg/registry/core/trace/ns_registry.go
@@ -19,6 +19,7 @@ package trace
 
 import (
 	"context"
+	"errors"
 	"io"
 
 	"github.com/networkservicemesh/sdk/pkg/registry/core/streamcontext"
@@ -26,6 +27,8 @@ import (
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/networkservicemesh/api/pkg/api/registry"
 )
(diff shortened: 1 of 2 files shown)


denis-tingaikin (approved):
Looks good.

Could you sign the commit?

caboteria:
I *think* that I signed the commit but please let me know if something's wrong as I'm new to git signatures.

denis-tingaikin:
@caboteria Seems like it still requires to sign the commit.  See at DCO check https://github.com/networkservicemesh/sdk/pull/1353/checks?check_run_id=8351758816

In short, you can just do something simple like

```
git reset --soft HEAD^
git commit -m 'TODO' **-s**
```

acnodal-tc:
Ah, I get it. Unfortunately confusing git terminology! I was signing the commit (-S) but it sounds like you want me to signoff (-s) the commit. It should be all set now, I did *both*.

denis-tingaikin:
@acnodal-tc It's a useful fix. Thanks!

acnodal-tc:
Thank you for your patience in helping me get the patch in shape to merge!

PR #1356, opened 2022-09-21
Don't restore eventFactory after Close/Unregister

glazychev-art:

## Description
Don't restore eventFactory in case the connection has already been closed/unregistered (it was removed from the map)
We ran into this when we started using values ​​from the passed context (for example spiffeID)

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


denis-tingaikin:
@glazychev-art Can we add a test to cover the issue?

Issue #1359, opened 2022-09-27
Datapath healing is not working properly


glazychev-art:
## Expected Behavior

All heal attempts should be without `reselect` if dataplane is healthy

## Current Behavior

If controlplane is down - only the **_first_** heal attempt will be without `reselect`. For all subsequent attempts, `reselect` will be chosen, even if the dataplane is normal.

## Root cause

We're handling the return from the `monitorDataPlane` function incorrectly.
Let's take a look at - https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/heal/eventloop.go#L204
This is just a sign of the end of the previous `eventLoop`, not an indication that the dataplan is dead. But now we treat this as a dataplan error and subsequent healing attempts are made with reselection.

## Possible solution
Change `monitorDataPlane` [function](https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/heal/eventloop.go#L190) signature as `monitorCtrlPlane` [function](https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/heal/eventloop.go#L95). In this case, we will be able to distinguish between a real problem with the dataplan and a simple closing of `eventLoop`.

glazychev-art:
@edwarnicke @denis-tingaikin

Perhaps we will see instabilities with tests after the fix, and it will take time to sort it out

PR #1362, opened 2022-09-28
heal: fix dataplane healing

glazychev-art:

## Issue link
https://github.com/networkservicemesh/sdk/issues/1359

## How Has This Been Tested?

- [x] ~~Added~~ Fixed unit test
- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/nsmgr/heal_test.go
+++ b/pkg/networkservice/chains/nsmgr/heal_test.go
@@ -278,7 +278,7 @@ func TestNSMGRHealEndpoint_DatapathHealthy_CtrlPlaneBroken(t *testing.T) {
 	domain.Nodes[0].NewEndpoint(ctx, nseReg2, sandbox.GenerateTestToken, counter)
 
 	// Should not connect to new NSE
-	require.Eventually(t, func() bool { return counter.UniqueRequests() == 1 }, timeout, tick)
+	require.Never(t, func() bool { return counter.UniqueRequests() > 1 }, time.Second*2, tick)
 	require.Equal(t, 1, counter.UniqueRequests())
 }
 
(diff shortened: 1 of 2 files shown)


pkg/networkservice/chains/nsmgr/heal_test.go
```diff
@@ -278,7 +278,7 @@ func TestNSMGRHealEndpoint_DatapathHealthy_CtrlPlaneBroken(t *testing.T) {
 	domain.Nodes[0].NewEndpoint(ctx, nseReg2, sandbox.GenerateTestToken, counter)
 
 	// Should not connect to new NSE
-	require.Eventually(t, func() bool { return counter.UniqueRequests() == 1 }, timeout, tick)
+	require.Never(t, func() bool { return counter.UniqueRequests() > 1 }, time.Second*2, tick)
```
denis-tingaikin, line 281: Why do we need to change this?

denis-tingaikin (commented):
I very much like the channel approach for the datapath healing.

Can we add a test for this scenario to avoid regressions?

pkg/networkservice/chains/nsmgr/heal_test.go
glazychev-art: This is an invalid condition in the test. It only worked on ci because of the timeout settings.

This test checks that if the dataplan is alive, we will not select another endpoint. Therefore, the number of unique requests to endpoints should never exceed 1.

In the previous implementation, the exit from `Eventually` happened almost instantly, just without waiting for the second request.

glazychev-art:
@denis-tingaikin 
In fact, we already have tests that cover this. But one of them didn't work correctly (fixed from Eventually to Never)

denis-tingaikin (approved):

PR #1368, opened 2022-10-12
qfix: Dont exclude subnets from routes

denis-tingaikin:
Signed-off-by: denis-tingaikin <denis.tingajkin@xored.com>

## Description
We shouldn't exclude whole subnets on refresh

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [X] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/excludedprefixes/client.go
+++ b/pkg/networkservice/common/excludedprefixes/client.go
@@ -190,6 +190,10 @@ func (epc *excludedPrefixesClient) Close(ctx context.Context, conn *networkservi
 func getRoutePrefixes(routes []*networkservice.Route) []string {
 	var rv []string
 	for _, route := range routes {
+		var o, b = route.GetPrefixIPNet().Mask.Size()
+		if o != b {
+			continue
+		}
 		rv = append(rv, route.GetPrefix())
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/excludedprefixes/client.go
```diff
@@ -190,6 +190,10 @@ func (epc *excludedPrefixesClient) Close(ctx context.Context, conn *networkservi
 func getRoutePrefixes(routes []*networkservice.Route) []string {
 	var rv []string
 	for _, route := range routes {
+		var o, b = route.GetPrefixIPNet().Mask.Size()
+		if o != b {
+			continue
+		}
```
edwarnicke, line 196: This seems to only compare the mask sizes between routes... I don't quite follow how this precludes us from including a Connection's routes/srcAddr/dstAddr in the exclude_prefixes for Refresh requests.

denis-tingaikin: as I can see srcAddr/dstAddr is already supported we only without a reason exclude a whole subnet. This patch prevents excluding a subnet.

edwarnicke: That will work for *now*.  But we need to be more specific in in the medium term because the routes from *other* vWires should be included :)

edwarnicke (approved):

PR #1369, opened 2022-10-13
ipv6 urls

acnodal-tc:
## Description

V6 addresses in URLs need to be wrapped in square brackets. The Resolver implementation didn't do this so URLs with IPV6 addresses were invalid.

## Issue link

n/a

## How Has This Been Tested?

- [X] Tested manually
- [ ] Have not tested

The unit tests needed more changes than the code under test. The tests used only 127.0.0.1 so the first commit in this PR enhances the tests to handle any address.

## Types of changes

- [X] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/tools/sandbox/fake_resolver.go
```diff
@@ -55,35 +62,31 @@ func (f *FakeDNSResolver) LookupSRV(ctx context.Context, service, proto, name st
 }
 
 // LookupIPAddr lookups IP address by host
-func (f *FakeDNSResolver) LookupIPAddr(_ context.Context, host string) ([]net.IPAddr, error) {
+func (f FakeDNSResolver) LookupIPAddr(_ context.Context, host string) ([]net.IPAddr, error) {
```
denis-tingaikin, line 65: Interesting. Why do we need to make the receiver a struct here?

denis-tingaikin:
@acnodal-tc resolve problems with red CI and it can be merged

pkg/tools/sandbox/fake_resolver.go
acnodal-tc: No need, I'll modify it to use pointers instead.

glazychev-art (approved):

denis-tingaikin (approved):

PR #1375, opened 2022-11-07
begin: update eventFactory context based on the initial one

glazychev-art:

## Description
In the current implementation, we update the `eventFactory` context [after a successful request](https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/begin/client.go#L85).
This is not quite right, because in addition to updating the context values, we also update **_the request timeout_**. It is calculated on the wayback.
For example, the request came with a timeout of 15s and took 5s. For the event factory, we calculate 10s, but we need 15s.

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/begin/event_factory_client_test.go
+++ b/pkg/networkservice/common/begin/event_factory_client_test.go
@@ -33,11 +33,13 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/common/begin"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/chain"
 	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+	"github.com/networkservicemesh/sdk/pkg/tools/clock"
+	"github.com/networkservicemesh/sdk/pkg/tools/clockmock"
 )
 
 // This test reproduces the situation when refresh changes the eventFactory context
 // nolint:dupl
-func TestRefresh_Client(t *testing.T) {
+func TestContextValues_Client(t *testing.T) {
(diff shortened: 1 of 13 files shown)


pkg/networkservice/common/begin/event_factory_client_test.go
```diff
@@ -130,6 +132,40 @@ func TestRefreshDuringClose_Client(t *testing.T) {
...
+	client := chain.NewNetworkServiceClient(
+		begin.NewClient(),
+		eventFactoryCl,
+		&delayedNSEClient{t: t, clock: clockMock},
+	)
+
+	// Do Request
+	request := testRequest("1")
+	conn, err := client.Request(ctx, request.Clone())
+	assert.NotNil(t, t, conn)
```
NikitaSkrynnik, line 159: Why don't we use `require` package here?

glazychev-art: Replaced with `require`

NikitaSkrynnik (approved):

denis-tingaikin (approved):

denis-tingaikin:
@glazychev-art Could you rebase this on the latest main and check that tests are working?

glazychev-art:
@denis-tingaikin
 It is already rebased (I did it yesterday) - https://github.com/networkservicemesh/sdk/pull/1375#event-8013060968

PR #1382, opened 2022-11-17
singlepointipam can work with IP ranges

zolug:
Related issue: cmd-nse-remote-vlan/110

Signed-off-by: Lugossy Zoltan <zoltan.lugossy@est.tech>

## Description
ippool extended so that IP addresses from an IP pool could be added to the pool.
singlepointipam modified to also support allocating IPs based on IP ranges.

## Issue link
https://github.com/networkservicemesh/cmd-nse-remote-vlan/issues/110

## How Has This Been Tested?
- [x] Tested manually
- [ ] Have not tested

## Types of changes
- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/networkservice/ipam/singlepointipam/range.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// IpamNet - IP network with optional IP ranges to limit address scope
+type IpamNet struct {
+	Network *net.IPNet
+	Ranges  []*IpamRange // optional: limit IpamNet by specifying valid range(s) within Network
+}
+
+// NewIpamNet - creates a new IpamNet
+func NewIpamNet() *IpamNet {
```
denis-tingaikin, line 38: @zolug 

Seems like it can be replaced with https://github.com/networkservicemesh/sdk/blob/main/pkg/tools/ippool/ippool.go#L37

IPPool already supports ranges/contains operation. The data structure is based on a red-black tree and should be extremely fast.

Am I missing something?

zolug: Hi!

For context here's the Remote VLAN NSE PR with the configuration changes that we would like to achieve:
https://github.com/networkservicemesh/cmd-nse-remote-vlan/pull/111

IpamNet is mostly aimed to be used verifying input in the Remote Vlan NSE (already relies on net pkg).

Correct me, if i'm wrong but IMHO ippool could be only used partially (without maybe major changes). For example to replace `IpamRange` slice in `IpamNet`.
That's because in our case the IPAM must be able to handle IPs of multiple ranges that all belong to the same pre-configured subnet, and not just a "random" set of IP addresses.
Meaning that for example the pre-configured network prefix had to be still stored some place, in order to execute admittance checks of ranges. (But the returned IPs are also assigned using the pre-configured network prefix.)

Thus for me it seems that e.g. an additional ippool could be needed for admittance control, that would contain the whole pre-configured subnet, but it wouldn't be used to allocate IPs from. (And as it seems the ippool package would also require a `ContainsRange` function but that's a minor thing.) And I still might struggle with some config checks.

On the other hand, I guess the allowed IP ranges could be mapped to a net + exclude ranges but IMHO that would add unnecessary complexity.

So my point is that IMHO a new struct still would be required even if built on top of ippool.

Maybe sg like this:
```
type IpamNet struct {
	Network *net.IPNet
	Ranges *ippool.IPPool
}
```

Or did I misundertand sg?

pkg/networkservice/ipam/singlepointipam/server_test.go
```diff
@@ -34,14 +34,29 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/networkservice/utils/metadata"
 )
 
-func newIpamServer(prefixes ...*net.IPNet) networkservice.NetworkServiceServer {
+func newIpamServer(networks ...*net.IPNet) networkservice.NetworkServiceServer {
```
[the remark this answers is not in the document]
denis-tingaikin:
```
singlepointipam % go test ./... -test.v
=== RUN   TestServer
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:9ed725e7-cd56-4df8-8bb4-714efb88e752] metadata deleted
--- PASS: TestServer (0.00s)
=== RUN   TestServerIPv6
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:bceb5088-abcb-4eb4-a27a-abbbf1e9f02f] metadata deleted
--- PASS: TestServerIPv6 (0.00s)
=== RUN   TestNilPrefixes
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:4e193851-b3dd-428a-9a47-77b4539a713f] metadata deleted
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:dc47c410-b56b-42c8-b8b0-96f3aaa1c334] metadata deleted
--- PASS: TestNilPrefixes (0.00s)
=== RUN   TestNilPrefixesIPv6
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:16dea0b5-01a1-4619-8e67-014b52dd44b7] metadata deleted
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:a7bcf515-e511-40e7-8624-e2997c88c2b2] metadata deleted
--- PASS: TestNilPrefixesIPv6 (0.00s)
=== RUN   TestExclude32Prefix
--- PASS: TestExclude32Prefix (0.00s)
=== RUN   TestExclude128PrefixIPv6
--- PASS: TestExclude128PrefixIPv6 (0.00s)
=== RUN   TestOutOfIPs
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:96131849-5601-45dd-ab2e-f3c0194167ed] metadata deleted
--- PASS: TestOutOfIPs (0.00s)
=== RUN   TestOutOfIPsIPv6
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:6d7c1d39-1084-437e-845b-18527d281613] metadata deleted
--- PASS: TestOutOfIPsIPv6 (0.00s)
=== RUN   TestAllIPsExcluded
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:8c60cb9a-44c9-43d6-bc15-0bec3f869b49] metadata deleted
--- PASS: TestAllIPsExcluded (0.00s)
=== RUN   TestAllIPsExcludedIPv6
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:3a16d9e8-967d-4fda-872d-12eac6181966] metadata deleted
--- PASS: TestAllIPsExcludedIPv6 (0.00s)
=== RUN   TestRefreshRequest
--- PASS: TestRefreshRequest (0.00s)
=== RUN   TestRefreshRequestIPv6
--- PASS: TestRefreshRequestIPv6 (0.00s)
=== RUN   TestNextError
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:16d973c4-a413-499d-b382-abcfc0c4f0cf] metadata deleted
--- PASS: TestNextError (0.00s)
=== RUN   TestRefreshNextError
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:272eb656-481b-4503-ac4b-c3e4f1c3620b] metadata deleted
--- PASS: TestRefreshNextError (0.00s)
=== RUN   TestServers
--- PASS: TestServers (0.00s)
=== RUN   TestRefreshRequestMultiServer
--- PASS: TestRefreshRequestMultiServer (0.00s)
=== RUN   TestIPv4Range
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:a7987da0-62f3-4184-a46d-0007009a0491] metadata deleted
--- PASS: TestIPv4Range (0.00s)
=== RUN   TestIPv6Range
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:a58a1ad8-ae3c-4606-8881-8659ec908212] metadata deleted
--- PASS: TestIPv6Range (0.00s)
=== RUN   TestIPv4MissingRange
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:a9f8aa97-09b0-4935-840f-0913ef556b62] metadata deleted
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:317915f5-3b10-4f84-9c0a-385eeccb5ffd] metadata deleted
--- PASS: TestIPv4MissingRange (0.00s)
=== RUN   TestIPv6MissingRange
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:7988ae1c-e1e0-478c-895d-acaa70cb660e] metadata deleted
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:f93b57c7-b5ca-4362-925f-89859f689dd0] metadata deleted
--- PASS: TestIPv6MissingRange (0.00s)
=== RUN   TestIPv4RangeOutOfBounds
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:22bd3e53-fbb0-4fcd-8058-744b0db7c0bb] metadata deleted
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:32fd90e8-cab4-4dd0-959b-dc995888d42c] metadata deleted
--- PASS: TestIPv4RangeOutOfBounds (0.00s)
=== RUN   TestIPv6RangeOutOfBounds
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:6ed49c86-3f7e-450a-b9db-154b735c66db] metadata deleted
2022/12/12 15:18:43 [DEBUG] [metadata:server] [connID:64396bdd-b4b9-486d-86ba-cc27f0d8be68] metadata deleted
--- PASS: TestIPv6RangeOutOfBounds (0.00s)
PASS
ok      github.com/networkservicemesh/sdk/pkg/networkservice/ipam/singlepointipam       0.153s
```

pkg/networkservice/ipam/singlepointipam/range.go
denis-tingaikin: Hello @zolug !

I started looking deeply into the patch and found that the new functionality is not covered by the unit test for the chain element.

Can we add a test? I think it could explain the necessity of the new `IpamNet`

denis-tingaikin (commented):
@zolug if I read the code correctly, basically these changes are needed to pre-exclude some of the IP addresses from the ippool for the connection. 

Does it look correct?

denis-tingaikin:
Question: Is it possible to use https://github.com/networkservicemesh/api/blob/main/pkg/api/networkservice/connectioncontext.proto#L69 to solve the problem? 

See at chain element https://github.com/networkservicemesh/sdk/tree/main/pkg/networkservice/common/excludedprefixes

zolug:
> @zolug if I read the code correctly, basically these changes are needed to pre-exclude some of the IP addresses from the ippool for the connection.
> 
> Does it look correct?

@denis-tingaikin Basically yes, but the goal is to split up a particular network between multiple independently operating singlepointipams.

denis-tingaikin:
@zolug OK, now I see the motivation. 

So, now let's figure out if is it possible to re-use `https://github.com/networkservicemesh/sdk/tree/main/pkg/networkservice/common/excludedprefixes` chain element.

zolug:
> @zolug OK, now I see the motivation.
> 
> So, now let's figure out if is it possible to re-use https://github.com/networkservicemesh/sdk/tree/main/pkg/networkservice/common/excludedprefixes chain element.

I'll try to list my points, but I don't think that's really suitable. Please feel free to weigh in and correct me.

-It's not just about excluding a few IP addresses but splitting up networks into ranges to be used by multiple independent singlepointipams or VLAN Remote NSEs. This could mean many-many excluded IPs for example (unless cleverly describing them with multiple networks).
-Also, for me it seems that the code dealing with excluded prefixes neither suport ranges.
And adding excluded ranges support seems way more intrusive. And could also imply that compared to the proposed PR a simple NSE uplift would not be sufficient to introduce the functionality (this might not be a priority for us though, but nsm core is handled separately and in general we try to avoid upgrade requirements towards it).
-A common excluded prefixes config in nsmgr is not feasible as you can see. Which IMHO leaves room for an NSC or NSE based excluded prefixes approach.
-It would be also preferable to contain the ipam configuration related information in one place. Which the excluded prefixes concept seems to contradict if these prefixes are passed by the NSCs or sg else (other than the NSE) on the path.
-I guess some NSE based excluded prefixes solution could provide config collocation. Where for example a new chain would "amend" the incoming request. Then once the ipam chain had finished it would remove the additional excluded prexies it previously added. Thus basically limiting the scope of the changes required in NSM (e.g. avoid impact on nsmgr merge). But this approach feels hackis and not generic either.
-We would like to offer a clean and simple configuration option for our users. As the goal is to split up a specific subnetwork among multiple independent IPAMs, the allowed ranges approach looks more natural. And I have a feeling that it could be overwhelming and error prone to require users to configure excluded prefixes instead of allowed ranges. But I could be completely wrong here :)
Of course doing the conversion internally in our product could be also an option. We do have an Operator as the official deployment option. However in case of helm based deployments it could be tricky/cumbersome to implement such allowed ranges to exluded IPs mapping.

So basically it felt more natural to modify the IPAM prefix configuration itself, both because it seems to have limited impact on NSM and the configuration feels more straightforward.

denis-tingaikin (approved):
I think it can be merged.

Would you like to add an encoder for IP ranges for `envoyconfig` to the sdk/pkg/tools?

edwarnicke:
Specifically, do you want to add something like [encoding.TextUnMarshaller](https://pkg.go.dev/encoding#TextUnmarshaler) so envconfig can unmarshal for you automatically:

https://github.com/kelseyhightower/envconfig#supported-struct-field-types

denis-tingaikin:
@zolug Do you have any updates on this?

PR #1384, opened 2022-11-23
Add priorityMechanisms chain element

glazychev-art:

## Description
This PR adds `priorityMechanisms` chain element.
The input is a list of mechanisms by priority. On the Request, the chain element sorts the mechanism preferences according to these priorities.

This element will be used on the forwarder.

## Issue link
https://github.com/networkservicemesh/sdk-vpp/issues/638

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/prioritymechanisms/client.go
+++ b/pkg/networkservice/common/prioritymechanisms/client.go
@@ -35,7 +35,7 @@ type priorityMechanismsClient struct {
 }
 
 // NewClient - returns a new client chain element that prioritize mechanisms according to the list
-func NewClient(priorities []string) networkservice.NetworkServiceClient {
+func NewClient(priorities ...string) networkservice.NetworkServiceClient {
 	c := &priorityMechanismsClient{
 		priorities: map[string]int{},
 	}
@@ -55,18 +55,23 @@ func (p *priorityMechanismsClient) Close(ctx context.Context, conn *networkservi
 	return next.Client(ctx).Close(ctx, conn, opts...)
 }
 
+// At the top of the list are the mechanisms to be sorted by priority. The tail contains mechanisms without priorities.
 func prioritizeMechanismsByType(mechanisms []*networkservice.Mechanism, priorities map[string]int) []*networkservice.Mechanism {
-	var head, tail []*networkservice.Mechanism
-	for _, mechanism := range mechanisms {
-		if _, ok := priorities[mechanism.GetType()]; ok {
-			head = append(head, mechanism)
+	tailIndex := 0
+	passedElements := 0
+	for ; passedElements < len(mechanisms); passedElements++ {
+		if _, ok := priorities[mechanisms[tailIndex].GetType()]; ok {
+			tailIndex++
 		} else {
-			tail = append(tail, mechanism)
+			value := mechanisms[tailIndex]
+			mechanisms = append(append(mechanisms[:tailIndex], mechanisms[tailIndex+1:]...), value)
 		}
 	}
-	sort.Slice(head, func(i, j int) bool {
-		return priorities[head[i].GetType()] < priorities[head[j].GetType()]
+
+	// Sort by priority
+	sort.SliceStable(mechanisms[:tailIndex], func(i, j int) bool {
+		return priorities[mechanisms[i].GetType()] < priorities[mechanisms[j].GetType()]
 	})
 
-	return append(head, tail...)
+	return mechanisms
 }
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/prioritymechanisms/client.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"google.golang.org/grpc"
+
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+type priorityMechanismsClient struct {
```
denis-tingaikin, line 33: How do we plan to use this in forwarder-vpp app?

glazychev-art: We add it to the forwarder's chain, for example after `filtermechanisms` - https://github.com/networkservicemesh/sdk-vpp/blob/main/pkg/networkservice/chains/forwarder/server.go#L152
To pass values, we will use `NSM_MECHANISM_PRIORITY` env that is a prioritized list (`[]string`) - the first value has the highest priority.

For example, if we have
```
            - name: NSM_MECHANISM_PRIORITY
              value: ipsec
```
this means that the mechanismPreferences order will be: `{ IPSec, Wireguard }`

pkg/networkservice/common/prioritymechanisms/client.go
```diff
@@ -0,0 +1,72 @@
...
+		} else {
+			tail = append(tail, mechanism)
+		}
+	}
+	sort.Slice(head, func(i, j int) bool {
+		return priorities[head[i].GetType()] < priorities[head[j].GetType()]
+	})
+
+	return append(head, tail...)
+}
```
denis-tingaikin, line 72: 1. What about a sequence of the same mechanism types? Say like MEMIF, MEMIF with different parameters.

I think we should use `sort.SliceStable`

2. Can we don't create a copy of the mechanism slice here?

pkg/networkservice/common/prioritymechanisms/client_test.go
```diff
@@ -0,0 +1,101 @@
...
+					Type: kernel.MECHANISM,
+				},
+				{
+					Cls:  cls.LOCAL,
+					Type: memif.MECHANISM,
+				},
+			},
+		}
+	}
+	samples := []struct {
```
denis-tingaikin, line 63: Consider to cover corner cases for nil and empty cases

pkg/networkservice/common/prioritymechanisms/client.go
denis-tingaikin: Looks good to me.

/cc @edwarnicke

pkg/networkservice/common/prioritymechanisms/client.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+type priorityMechanismsClient struct {
+	priorities map[string]int
+}
+
+// NewClient - returns a new client chain element that prioritize mechanisms according to the list
+func NewClient(priorities []string) networkservice.NetworkServiceClient {
```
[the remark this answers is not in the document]
glazychev-art: Fixed

pkg/networkservice/common/prioritymechanisms/client.go
glazychev-art: 1. I'm not sure about the same mechanism types. If we have MEMIF and MEMIF, we need additional diff parameters. What if the user needs the second MEMIF? Also, sort.Slice is a bit faster :)

2. Yes, we can. Fixed

pkg/networkservice/common/prioritymechanisms/client_test.go
glazychev-art: Fixed

pkg/networkservice/common/prioritymechanisms/client.go
denis-tingaikin: >  If we have MEMIF and MEMIF, we need additional diff parameters. What if the user needs the second MEMIF? Also, sort.Slice is a bit faster :)

Note: We shouldn't change the order of MEMIF and MEMIF sequence. So we need to use a stable sort.

glazychev-art: Fixed

denis-tingaikin (approved):

Issue #1398, opened 2022-12-16
Authorize on timeout


glazychev-art:
## Current Behavior

Now there may be a problem with closing the connection on `timeout`, since the timeout mostly happens when the token expires. 
After that, `Close` is called, but it fails on `authorize`, because token is already expired.

## Steps to Reproduce

1. Step 1
2. Step 2
3. You get the idea...

## Context

* Kubernetes Version:
* etc.

glazychev-art:
The general idea is to call `timeout`/`expire` in advance, considering the request timeout.

Steps:
- [ ] **networkservice**: consider request timeout before calling [Close](https://github.com/networkservicemesh/sdk/blob/main/pkg/networkservice/common/timeout/server.go#L71)
- [ ] **networkservice**: sandbox test
- [ ] **registry**: `updatepath` should update nse.ExpirationTime (skipped values [here](https://github.com/networkservicemesh/sdk/blob/main/pkg/registry/common/updatepath/nse_server.go#L48-L52))
- [ ] **registry**: `expire` updates expirationTime to default value only if it was not set. Call `Unregister` in advance
- [ ] **registry**: sandbox test
- [ ] manual integration testing

denis-tingaikin:
Fixed by https://github.com/networkservicemesh/sdk/pull/1404

PR #1404, opened 2022-12-23
Fix timeout/expire chain elements

glazychev-art:

## Description
Main changes:
1. **networkservice timeout**: at the time of the call `Close()` on `timeout`, the token may already be expired. Therefore, we need to call `Close()` in advance - `(tokenTimeout - requestTimeout)`
2. **registry expire**:  since we now have tokens in the registry, we must follow the same logic as in the **networkservice** (see above). Expiration time selection has been extended. Now it is  - `min(peertoken.ExpirationTime, token.ExpirationTime, nse.ExpirationTIme) - registerTimeout`.
3. **grpcmetadata.NewNetworkServiceEndpointRegistryClient** - `grpc.Header` was added to `Unregister()` to process an updated `path` on the way back (like here - https://github.com/networkservicemesh/sdk/blob/main/pkg/registry/common/grpcmetadata/nse_client.go#L47-L48)
4. **registry authorize and updatepath** were placed before `begin`, because in case of an event that triggers an `eventFactory`, we don't need to re-authorize.
5. **registry begin** - when updating the context, we need to keep a copy of `path`, since we access it by reference (we should not be able to accidentally change it after `begin`)

## Issue link
https://github.com/networkservicemesh/sdk/issues/1398

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionallity
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/nsmgr/heal_test.go
+++ b/pkg/networkservice/chains/nsmgr/heal_test.go
@@ -538,10 +538,6 @@ func testNSMGRCloseHeal(t *testing.T, withNSEExpiration bool) {
 		SetNSMgrProxySupplier(nil).
 		SetRegistryProxySupplier(nil)
 
-	if withNSEExpiration {
-		builder = builder.SetRegistryExpiryDuration(time.Second / 2)
-	}
-
 	domain := builder.Build()
 
 	nsRegistryClient := domain.NewNSRegistryClient(ctx, sandbox.GenerateTestToken)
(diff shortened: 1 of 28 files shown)


pkg/networkservice/chains/nsmgr/heal_test.go
```diff
@@ -538,20 +538,20 @@ func testNSMGRCloseHeal(t *testing.T, withNSEExpiration bool) {
 		SetNSMgrProxySupplier(nil).
 		SetRegistryProxySupplier(nil)
 
-	if withNSEExpiration {
-		builder = builder.SetRegistryExpiryDuration(time.Second / 2)
-	}
-
```
denis-tingaikin, line 544: Why do we change this?

pkg/registry/common/expire/nse_server.go
```diff
@@ -32,35 +32,38 @@ import (
 )
 
 type expireNSEServer struct {
-	nseExpiration time.Duration
-	ctx           context.Context
+	defaultNseExpiration time.Duration
```
denis-tingaikin, line 35: Should we get rid of default expiration? Seems like the update path is going to always put the expiration time for nses after these changes.

```suggestion
```

For unit tests, we can add `injectexpirationtime` chain element.

glazychev-art: Totally agree
But should we consider the situation where the user hasn't `updatepath`?

pkg/networkservice/chains/nsmgr/heal_test.go
glazychev-art: We now rely mostly on the token rather than registryExpirationDuration. Therefore, below we create a token with the expiration we need.

denis-tingaikin: Got it

denis-tingaikin (approved):

PR #1439, opened 2023-03-23
timeout: take the minimum expiration time of pathSegments

glazychev-art:
## Issue link
https://github.com/networkservicemesh/deployments-k8s/issues/8882

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/networkservice/common/timeout/server.go
```diff
@@ -97,3 +92,34 @@ func (s *timeoutServer) Close(ctx context.Context, conn *networkservice.Connecti
 	}
 	return &empty.Empty{}, err
 }
+
+func after(ctx context.Context, conn *networkservice.Connection) time.Duration {
+	clockTime := clock.FromContext(ctx)
+
+	var minTimeout *time.Duration
```
denis-tingaikin, line 99:
```suggestion
	var minTimeout time.Duration = 1
```

pkg/networkservice/common/timeout/server.go
```diff
@@ -97,3 +92,34 @@ func (s *timeoutServer) Close(ctx context.Context, conn *networkservice.Connecti
...
+			expireTime = expTime
+		}
+	}
+
+	if minTimeout == nil || *minTimeout <= 0 {
+		return 1
+	}
+	log.FromContext(ctx).Infof("expiration after %s at %s", minTimeout.String(), expireTime.UTC())
+
+	return *minTimeout
```
denis-tingaikin, line 124:
```suggestion
	return minTimeout
```

pkg/networkservice/common/timeout/server.go
```diff
@@ -97,3 +92,34 @@ func (s *timeoutServer) Close(ctx context.Context, conn *networkservice.Connecti
 	}
 	return &empty.Empty{}, err
 }
+
+func after(ctx context.Context, conn *networkservice.Connection) time.Duration {
```
denis-tingaikin, line 96: I think we need to rename it. 

For example,

```suggestion
func minTokenTimeout(ctx context.Context, conn *networkservice.Connection) time.Duration {
```

denis-tingaikin (commented):
I think these changes make sense.

@edwarnicke What do you think?

pkg/networkservice/common/timeout/server.go
glazychev-art: we can't set 1 here because we won't find a value less than that in Path segments

d-uzlov (approved):

NikitaSkrynnik (approved):

Issue #1442, opened 2023-03-28
Memory DNS Handler handles requests incorrectly if it doesn't have the required DNS record.


NikitaSkrynnik:
## Expected Behavior

Memory DNS handler handles requests with empty answers correctly.

## Current Behavior

Memory DNS handler handler requests for non-existent records incorrectly. Some utilities (ping, for example) don't work because of this.

## Steps to Reproduce

1. Run vl3-dns example
2. Ping one alpine from another alpine using its dns name without `-4` flag

## Failure Logs
[logs.txt](https://github.com/networkservicemesh/sdk/files/11092655/logs.txt)

denis-tingaikin:
Seems like fixed by https://github.com/networkservicemesh/sdk/pull/1443

PR #1443, opened 2023-03-29
Rework request handling in memory handler

NikitaSkrynnik:
## Issue link
[](https://github.com/networkservicemesh/sdk/issues/1442)

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/dnsutils/memory/handler.go
+++ b/pkg/tools/dnsutils/memory/handler.go
@@ -54,10 +54,6 @@ func (f *memoryHandler) ServeDNS(ctx context.Context, rw dns.ResponseWriter, msg
 		return
 	}
 
-	rwWrapper := &responseWriter{
-		ResponseWriter: rw,
-	}
-
 	var name = dns.Name(msg.Question[0].Name).String()
 	var resp = new(dns.Msg)
 	resp.SetReply(msg)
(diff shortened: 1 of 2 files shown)


pkg/tools/dnsutils/memory/handler.go
```diff
@@ -72,18 +68,23 @@ func (f *memoryHandler) ServeDNS(ctx context.Context, rw dns.ResponseWriter, msg
...
 		return
 	}
 
-	if err := rw.WriteMsg(resp); err != nil {
-		dns.HandleFailed(rw, msg)
+	if _, ok := f.records.Load(name); ok {
+		m := new(dns.Msg)
+		_ = rw.WriteMsg(m.SetRcode(msg, dns.RcodeSuccess))
+	} else {
+		rwWrapper := &responseWriter{ResponseWriter: rw}
```
denis-tingaikin, line 82: 1. please remove `responseWriter.`

`fanout` should handler the server fail.

2. add regression unit test

denis-tingaikin (changes requested):

pkg/tools/dnsutils/memory/handler.go
NikitaSkrynnik: `responseWriter` doesn't check only fanout, but any handler that goes after memory handler. If we remove `responseWriter` we can get an error, because in some cases nothing will be written to `rw`.

denis-tingaikin (approved):
@glazychev-art Do you have any comments?

glazychev-art (approved):

PR #1456, opened 2023-05-10
Fix InitialRegistrationTime in case of reregistration

glazychev-art:
## Description
We should not change the `InitialRegistrationTime` field if we have already registered the NSE before.

Added registry `metadata` chain element

## Issue link
https://github.com/networkservicemesh/deployments-k8s/issues/9063

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/registry/common/setregistrationtime/nse_server.go
+++ b/pkg/registry/common/setregistrationtime/nse_server.go
@@ -1,5 +1,7 @@
 // Copyright (c) 2021 Doc.ai and/or its affiliates.
 //
+// Copyright (c) 2023 Cisco and/or its affiliates.
+//
 // SPDX-License-Identifier: Apache-2.0
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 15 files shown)


pkg/registry/common/setregistrationtime/nse_server.go
```diff
@@ -19,15 +21,19 @@ package setregistrationtime
...
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/networkservicemesh/sdk/pkg/registry/core/next"
 	"github.com/networkservicemesh/sdk/pkg/tools/clock"
 )
 
-type setregtimeNSEServer struct{}
+type setregtimeNSEServer struct {
+	genericsync.Map[string, *timestamppb.Timestamp]
```
denis-tingaikin, line 35: Can we use metadata instead?

glazychev-art: @denis-tingaikin 
I thought about it, but unfortunately we don't have `metadata` for registry

denis-tingaikin: It might be a good time to add it ;)

glazychev-art: Fixed :)

glazychev-art:
@denis-tingaikin 
Updated - https://github.com/networkservicemesh/sdk-k8s/pull/439

Issue #1457, opened 2023-05-10
Healing after failed refresh


glazychev-art:
## Expected Behavior

Healing works as expected

## Current Behavior

Healing doesn't work if refresh fails

## Failure Information (for bugs)

This happens because we cancel the previous healing monitoring on refresh, but do nothing in case of error: 
https://github.com/networkservicemesh/sdk/blob/6fa2f68d5532ace9f0d7c4edccb1e069d35c15df/pkg/networkservice/common/heal/client.go#L61-L68

Perhaps we need to consider re-refresh right after a failure, or try to restore the previous monitoring

## Steps to Reproduce

1. NSC connects to NSE
2. Refresh Request from NSC fails
3. Kill NSE to start healing
4. Healing doesn't start

## Context

* Kubernetes Version:
* etc.

d-uzlov:
There are 2 separate problems:
1. Refresh doesn't do anything if its request fails
2. Heal stops working after a refresh failed

--- 
> Refresh doesn't do anything if its request fails

I see 2 strategies:
1. Use `retry` strategy for refreshes:
  On error repeat requests every `interval` milliseconds until one request succeeds
2. On error reduce refresh interval in half:
  For example: `refresh after 3 minutes` -> `error` -> `refresh after 1.5 minutes` -> `error` -> `refresh after 0.75 minutes` -> `success` -> `refresh after 3 minutes`

---
> Heal stops working after a refresh failed

Current `heal` behavior when it receives a request:
- Stop current monitor thread.
- On success create a new monitor thread.
- On error do nothing.

This works because of the monitor thread behavior:
- Monitor can be cancelled while it waits for "down" events
- Monitor can never be cancelled after it received a "down" event and started healing
- Monitor retries requests until one succeeds
- Monitor checks data plane in the healing loop

Reasons heal must stop monitoring before request:
1. Control plane monitoring relies on gRPC connection to next server, which is closed and recreated during request, to update timeouts and tokens.
2. Data plane monitoring may give you a false-negative result if connection was interrupted during request

Reasons heal does not start monitoring after error:
1. Control plane monitoring doesn't work because gRPC connection is closed on request error.
2. Usually failed requests are caused by healing loop. If we start a new one while the old is still running, we will have several parallel healing loops.
3. If error happens before first success for current connection, there is no reason to heal it.

But monitoring after error is required in case the fail was caused be refresh or by user.
We need to solve the reasons `heal` doesn't start a new monitor on error:
[...]

PR #1465, opened 2023-06-06
Fix healing after failed refresh

d-uzlov:
## Description

Restart connection monitoring after a failed refresh.

## Issue link

https://github.com/networkservicemesh/sdk/issues/1457

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/chains/client/options.go
+++ b/pkg/networkservice/chains/client/options.go
@@ -109,3 +109,13 @@ func WithoutRefresh() Option {
 		c.refreshClient = null.NewClient()
 	}
 }
+
+// WithRefresh sets a custom refreshClient for the client chain
+func WithRefresh(refreshClient networkservice.NetworkServiceClient) Option {
+	if refreshClient == nil {
+		panic("refreshClient cannot be nil")
+	}
+	return Option(func(c *clientOptions) {
+		c.refreshClient = refreshClient
+	})
+}
--- a/pkg/networkservice/common/heal/eventloop.go
+++ b/pkg/networkservice/common/heal/eventloop.go
@@ -94,6 +94,80 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
 	return eventLoopCancel, cev.monitorFinishedCh, nil
 }
 
+func (cev *eventLoop) eventLoop() {
+	needToHeal, reselect := cev.waitForEvents()
+
+	if !needToHeal {
+		return
+	}
+
+	for {
+		select {
+		case <-cev.chainCtx.Done():
+			return
+		default:
+			if cev.chainCtx.Err() != nil {
+				return
+			}
+
+			// We need to force check the DataPlane if a down event was received from the ControlPlane
+			if !reselect {
+				deadlineCtx, deadlineCancel := context.WithDeadline(cev.chainCtx, time.Now().Add(cev.heal.livenessCheckTimeout))
+				if !cev.heal.livenessCheck(deadlineCtx, cev.conn) {
+					cev.logger.Warnf("Data plane is down")
+					reselect = true
+				}
+				deadlineCancel()
+			}
+
+			var options []begin.Option
+			if reselect {
+				cev.logger.Debugf("Reconnect with reselect")
+				options = append(options, begin.WithReselect())
+			}
+			err := <-cev.eventFactory.Request(options...)
+			if err == nil {
+				return
+			}
+		}
+	}
+}
+
+func (cev *eventLoop) waitForEvents() (needToHeal bool, reselect bool) {
+	defer close(cev.monitorFinishedCh)
+
+	ctrlPlaneCh := cev.monitorCtrlPlane()
+	dataPlaneCh := cev.monitorDataPlane()
+
+	if dataPlaneCh == nil {
+		// Since we don't know about data path status - always use reselect
+		reselect = true
+	}
+
+	select {
+	case _, ok := <-ctrlPlaneCh:
+		if ok {
+			// Connection closed
+			return
+		}
+		cev.logger.Warnf("Control plane is down")
+		needToHeal = true
+		cev.monitorFinishedCh <- needToHeal
+	case _, ok := <-dataPlaneCh:
+		if ok {
+			// Connection closed
+			return
+		}
+		cev.logger.Warnf("Data plane is down")
+		reselect = true
+		needToHeal = true
+		cev.monitorFinishedCh <- needToHeal
+	case <-cev.chainCtx.Done():
+	case <-cev.eventLoopCtx.Done():
+	}
+	return
+}
+
 func (cev *eventLoop) monitorCtrlPlane() <-chan struct{} {
 	res := make(chan struct{}, 1)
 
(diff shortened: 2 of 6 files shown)


pkg/networkservice/chains/client/options.go
```diff
@@ -109,3 +109,13 @@ func WithoutRefresh() Option {
...
+
+// WithRefresh sets a custom refreshClient for the client chain
+func WithRefresh(refreshClient networkservice.NetworkServiceClient) Option {
+	if refreshClient == nil {
+		panic("refreshClient cannot be nil")
+	}
+	return Option(func(c *clientOptions) {
+		c.refreshClient = refreshClient
+	})
+}
```
denis-tingaikin, line 121: Do not add a new api for testing goals.

pkg/networkservice/chains/nsmgr/select_forwarder_test.go
```diff
@@ -219,8 +219,6 @@ func Test_DiscoverForwarder_ChangeForwarderOnDeath_LostHeal(t *testing.T) {
 	conn, err = nsc.Request(ctx, request.Clone())
 	require.NoError(t, err)
 	require.Equal(t, 1, counter.UniqueRequests())
-	require.Equal(t, 2, counter.Requests())
-	require.Equal(t, 0, counter.Closes())
```
denis-tingaikin, line 223: Could you explain the change?

pkg/networkservice/chains/nsmgr/select_forwarder_test.go
```diff
@@ -288,7 +286,5 @@ func Test_DiscoverForwarder_ChangeRemoteForwarderOnDeath(t *testing.T) {
 	conn, err = nsc.Request(ctx, request.Clone())
 	require.NoError(t, err)
 	require.Equal(t, 1, counter.UniqueRequests())
-	require.Equal(t, 3, counter.Requests())
-	require.Equal(t, 0, counter.Closes())
```
denis-tingaikin, line 292: Could you explain the change?

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
...
 		cev.logger.Warnf("Data plane is down")
 		reselect = true
+		needToHeal = true
+		cev.monitorFinishedCh <- needToHeal
 	case <-cev.chainCtx.Done():
-		return
 	case <-cev.eventLoopCtx.Done():
-		return
 	}
+	return needToHeal, reselect
```
denis-tingaikin, line 173:
```suggestion
	return false, false
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
...
 		if ok {
 			// Connection closed
-			return
+			return needToHeal, reselect
 		}
-		// Start healing
 		cev.logger.Warnf("Data plane is down")
 		reselect = true
+		needToHeal = true
+		cev.monitorFinishedCh <- needToHeal
```
denis-tingaikin, line 169:
```suggestion
		cev.monitorFinishedCh <- needToHeal
		return true, true
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
...
 	case _, ok := <-dataPlaneCh:
 		if ok {
 			// Connection closed
-			return
+			return needToHeal, reselect
 		}
-		// Start healing
 		cev.logger.Warnf("Data plane is down")
 		reselect = true
+		needToHeal = true
```
denis-tingaikin, line 168:
```suggestion
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
 	case _, ok := <-ctrlPlaneCh:
 		if ok {
 			// Connection closed
-			return
+			return needToHeal, reselect
```
denis-tingaikin, line 156:
```suggestion
			return false, false
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
...
 		}
-		// Start healing
 		cev.logger.Warnf("Control plane is down")
+		needToHeal = true
+		cev.monitorFinishedCh <- needToHeal
 	case _, ok := <-dataPlaneCh:
 		if ok {
 			// Connection closed
-			return
+			return needToHeal, reselect
```
denis-tingaikin, line 164:
```suggestion
			return false, false
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
...
-			if !reselect {
-				deadlineCtx, deadlineCancel := context.WithDeadline(cev.chainCtx, time.Now().Add(cev.heal.livenessCheckTimeout))
-				if !cev.heal.livenessCheck(deadlineCtx, cev.conn) {
-					cev.logger.Warnf("Data plane is down")
-					reselect = true
+			// Handle error
+			if err != nil {
+				s, _ := status.FromError(err)
+				// This condition means, that the client closed the connection. Stop healing
+				if s.Code() == codes.Canceled {
```
denis-tingaikin, line 194: Why are we looking at grpc status?

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
 	case _, ok := <-ctrlPlaneCh:
 		if ok {
 			// Connection closed
-			return
+			return needToHeal, reselect
 		}
-		// Start healing
 		cev.logger.Warnf("Control plane is down")
+		needToHeal = true
+		cev.monitorFinishedCh <- needToHeal
```
denis-tingaikin, line 160:
```suggestion
		cev.monitorFinishedCh <- true
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -142,54 +153,59 @@ func (cev *eventLoop) eventLoop() {
 	case _, ok := <-ctrlPlaneCh:
 		if ok {
 			// Connection closed
-			return
+			return needToHeal, reselect
 		}
-		// Start healing
 		cev.logger.Warnf("Control plane is down")
+		needToHeal = true
```
denis-tingaikin, line 159:
```suggestion
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -66,69 +68,78 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
-	res := make(chan struct{}, 1)
-
-	go func() {
-		defer close(res)
+func (cev *eventLoop) eventLoop() {
+	needToHeal, reselect := cev.waitForEvents()
 
-		for {
-			eventIn, err := cev.client.Recv()
+	if !needToHeal {
```
denis-tingaikin, line 102: Probably name inversion could be more clear

```suggestion
	if healthy {
```

denis-tingaikin: What if needToHead=false and reselct=true?

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -66,69 +68,78 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
-				// Otherwise - Start healing
-				return
+				deadlineCancel()
 			}
 
-			// Handle event. Start healing
-			if eventIn.GetConnections()[cev.conn.GetId()].GetState() == networkservice.State_DOWN {
+			var options []begin.Option
+			if reselect {
+				cev.logger.Debugf("Reconnect with reselect")
```
denis-tingaikin, line 127:
```suggestion
			var options []begin.Option
			if reselect {
			...
			} else {
				deadlineCtx, deadlineCancel := context.WithDeadline(cev.chainCtx, time.Now().Add(cev.heal.livenessCheckTimeout))
				if !cev.heal.livenessCheck(deadlineCtx, cev.conn) {
					cev.logger.Warnf("Data plane is down")
					reselect = true
				}
				deadlineCancel()
			}
```

pkg/networkservice/common/heal/eventloop.go
```diff
@@ -66,69 +68,78 @@ func newEventLoop(ctx context.Context, cc grpc.ClientConnInterface, conn *networ
...
 		}
-	}()
-
-	return res
+	}
 }
 
-func (cev *eventLoop) eventLoop() {
-	reselect := false
+func (cev *eventLoop) waitForEvents() (needToHeal, reselect bool) {
```
denis-tingaikin, line 139: As I can see always if needToHeal is true then reselect is also true. 

Consider to merge to one bool response.

denis-tingaikin (changes requested):

pkg/networkservice/common/heal/eventloop.go
d-uzlov: This is the contract of `MonitorConnection_MonitorConnectionsClient` interface which is generated by GRPC.
This is an old code, I didn't modify it, I just moved to around to make it easier to read.
I moved it back to reduce number of changes in PR.

pkg/networkservice/chains/client/options.go
d-uzlov: Fixed

pkg/networkservice/chains/nsmgr/select_forwarder_test.go
d-uzlov: I modified it because this test was unstable, and yesterday the CI was constantly failing.
There was somewhat of a race between the test code and healing.
I figured out that this check isn't really important for the purpose of this test.
Now I pushed a better solution which makes the test more deterministic.

pkg/networkservice/common/heal/eventloop.go
d-uzlov: In the event monitoring there are three states:
1. Context was cancelled, `needToHeal == false`
2. Control plane is down, `needToHeal == true`, `reselect == false`
3. Data plane is down, `needToHeal == true`, `reselect == true`

Maybe we could add a special enum for these three states, but I think the code would only become more complex, because these 2 variables are checked and changed in different places.

d-uzlov: > Probably name inversion could be more clear

Fixed.
I modified it to use `canceled`.

pkg/networkservice/common/heal/eventloop.go
d-uzlov: This is also old code that was moved. Now I reverted this to make it clear that I didn't modify it.
Here `reselect` is not a constant flag. It is used as a state that could be switched, so it doesn't make sense to use `if-else` pattern.

denis-tingaikin (approved):

denis-tingaikin:
@glazychev-art Could you have a look?

glazychev-art (approved):

PR #1509, opened 2023-09-11
Add prometheus as metric exporter

LionelJouin:
## Description

An init method supporting prometheus for the metric exporter has been added for the opentelemetry pkg. NSMgr or any metric collector could now export the metrics via prometheus.
span trace and metrics can now be individually disabled by passing nil value to the optl init function.

## Issue link
/

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/tools/opentelemetry/exporter.go
```diff
@@ -52,8 +54,8 @@ func InitSpanExporter(ctx context.Context, exporterURL string) trace.SpanExporte
 	return exporter
 }
 
-// InitMetricExporter - returns an instance of OpenTelemetry Metric Exporter.
-func InitMetricExporter(ctx context.Context, exporterURL string) *sdkmetric.Exporter {
+// InitOPTLMetricExporter - returns an instance of OpenTelemetry Metric Exporter.
+func InitOPTLMetricExporter(ctx context.Context, exporterURL string) sdkmetric.Reader {
```
denis-tingaikin, line 58: I'm not sure about the change because it breaks compatibility with cmd apps.

Could you explain how is critical to change signature?

bellycat77 (approved):

pkg/tools/opentelemetry/exporter.go
LionelJouin: The open telemetry SDK returns a metric.Reader for Prometheus for the prometheus exporter. 
https://github.com/open-telemetry/opentelemetry-go/blob/v1.16.0/exporters/prometheus/exporter.go#L53 
From what I understood with the open telemetry SDK, `metric.Reader` is more generic way to be used as way of collecting data than `sdkmetric.Exporter`

denis-tingaikin: OK, got it.
Do you need the change in release v1.11.0?

LionelJouin: @szvincze ?

szvincze: Right now I don't think it is critical to have it in 1.11.0.

pkg/tools/opentelemetry/exporter.go
```diff
@@ -73,5 +75,22 @@ func InitMetricExporter(ctx context.Context, exporterURL string) *sdkmetric.Expo
 		return nil
 	}
 
-	return &exporter
+	return sdkmetric.NewPeriodicReader(
+		exporter,
+		sdkmetric.WithInterval(10*time.Second),
```
NikitaSkrynnik, line 80: Isn't it better to make it configurable?

LionelJouin: do you mean the 10 seconds?

pkg/tools/opentelemetry/exporter.go
denis-tingaikin: @LionelJouin , @szvincze OK, let's target this for release 1.12.0 because this change will require updating all cmd apps due to the changing contract of function InitMetricExporter -> InitOPTLMetricExporter

szvincze: > @LionelJouin , @szvincze OK, let's target this for release 1.12.0 because this change will require updating all cmd apps due to the changing contract of function InitMetricExporter -> InitOPTLMetricExporter

@denis-tingaikin: We had some internal discussions about this topic. It seems this change would save quite significant effort for us if we could get it in 1.11.0. Is it possible to consider it in 1.11.0? Is there other alternative to get it before 1.12.0?

denis-tingaikin: Sure, but we'll need a help with updating cmd-repos.

Also, we supposed to be ready for delaying in release if something went wrong.

@edwarnicke Don't you mind if we consider this PR in this release (v.1.11.0)?

denis-tingaikin:
@LionelJouin a you may run `go mod tidy` to make it green.

denis-tingaikin:
@LionelJouin I think we may start with v1.20.8 https://github.com/networkservicemesh/.github/blob/main/.github/workflows/build-and-test.yaml#L22
Could you open PR?

LionelJouin:
Sure: https://github.com/networkservicemesh/.github/pull/45

denis-tingaikin (approved):
LGTM

@glazychev-art , @NikitaSkrynnik Could you also have a look?

NikitaSkrynnik (approved):

denis-tingaikin (approved):

Issue #1512, opened 2023-09-19
tokens_expired policy in case of partial expiration


glazychev-art:
## Current Behavior

If we run tokens_expired policy on timeout (on Close), we can run into the problem, that the left side of the Path is already expired, but the right side (including the current element) is still fine.

Suggestion: we can check only the right side of the Path

PR #1517, opened 2023-09-28
opa: token_expired policy should check only the right side of Path

glazychev-art:
## Issue link
Closes: https://github.com/networkservicemesh/sdk/issues/1512

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


denis-tingaikin (commented):
I see the problem that we're trying to fix with this patch, but I'm not sure that it will work for all cases. For example, the right part could also be expired or we may have a custom policy from the user.

Seems to me that  closes from timeout should be always authorized.
 Thoughts?

PR #1519, opened 2023-10-02
[begin]: event_factory should cancel Close context in case of reselect

glazychev-art:
## Description
Chain elements in Closes can rely on the context and its timeout. Therefore, after the operation is completed, the context must be canceled.

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/common/begin/event_factory.go
+++ b/pkg/networkservice/common/begin/event_factory.go
@@ -101,13 +101,13 @@ func (f *eventFactoryClient) Request(opts ...Option) <-chan error {
 			request := f.request.Clone()
 			if o.reselect {
 				ctx, cancel := f.ctxFunc()
-				defer cancel()
 				_, _ = f.client.Close(ctx, request.GetConnection(), f.opts...)
 				if request.GetConnection() != nil {
 					request.GetConnection().Mechanism = nil
 					request.GetConnection().NetworkServiceEndpointName = ""
 					request.GetConnection().State = networkservice.State_RESELECT_REQUESTED
 				}
+				cancel()
 			}
 			ctx, cancel := f.ctxFunc()
 			defer cancel()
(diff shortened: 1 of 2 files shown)


pkg/networkservice/common/begin/event_factory.go
```diff
@@ -101,13 +101,13 @@ func (f *eventFactoryClient) Request(opts ...Option) <-chan error {
 			request := f.request.Clone()
 			if o.reselect {
 				ctx, cancel := f.ctxFunc()
-				defer cancel()
 				_, _ = f.client.Close(ctx, request.GetConnection(), f.opts...)
```
dualBreath, line 104: LGTM,
ran the test on my side, works fine

Just a small question about the source of the issue: Am I correct in understanding that the original problem occurs because the `cancel()` for `Request` was run before the `cancel()` for `Close`, because they both use the `defer` directive?

dualBreath (approved):

pkg/networkservice/common/begin/event_factory.go
glazychev-art: Not really. Close can use the context somewhere in its goroutine.
And this goroutine will be alive as long as this context is alive. But after the parent Close is completed, this is unnecessary and even dangerous - the goroutine can use the same data as the subsequent Request.
Therefore, before the Request, we need to complete Close call (including its goroutines)

dualBreath: Great, thanks for explanation!

d-uzlov (approved):

denis-tingaikin (approved):

PR #1529, opened 2023-10-13
`begin` should let NSE unregister it self even if it doesn't have an event factory

NikitaSkrynnik:
## Description
`begin` should let NSE unregister it self even if it doesn't have an event factory. Also change `expire` a bit to distinguish between events coming from `begin` and NSE.

## Issue link
https://github.com/networkservicemesh/sdk-k8s/issues/456

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/registry/common/begin/nse_server.go
+++ b/pkg/registry/common/begin/nse_server.go
@@ -18,7 +18,7 @@ package begin
 
 import (
 	"context"
-	"sync/atomic"
+	"sync"
 
 	"github.com/edwarnicke/genericsync"
 	"github.com/golang/protobuf/ptypes/empty"
@@ -32,8 +32,8 @@ import (
 )
 
 type queue struct {
-	eventCount *atomic.Int64
-	queueChan  chan func()
+	eventCount int
+	sync.Mutex
 }
 
 type beginNSEServer struct {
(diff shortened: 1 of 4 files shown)


pkg/registry/common/begin/nse_server.go
```diff
@@ -92,8 +99,37 @@ func (b *beginNSEServer) Unregister(ctx context.Context, in *registry.NetworkSer
...
+		}
+
+		q.Lock()
+		_, err := next.NetworkServiceEndpointRegistryServer(ctx).Unregister(ctx, in)
+		q.eventCount--
+		if q.eventCount == 0 {
+			b.queueMap.Delete(id)
+		}
+		q.Unlock()
+		return &emptypb.Empty{}, err
```
d-uzlov, line 132: Is there a reason we hold the lock for the whole `next.Unregister` call?
I would think that we only need it for `q.eventCount` manipulations and `b.queueMap.Delete`.

NikitaSkrynnik: Yes, there is a reason for this. Next chain elements can manipulate some data and if we have several concurrent Unregister events it can lead to data races. We need mutex for the whole Unregister call to let only one event go further.

d-uzlov: Oh, I have just noticed that you don't use event factory here.
But now I have a different question: what will happen if we get `Register` call with the same ID as this Unregister call?

NikitaSkrynnik:
Fixed in another PR: https://github.com/networkservicemesh/sdk/pull/1542

PR #1582, opened 2024-02-01
qfix: Do not reset IPContext configuration on the reselect

denis-tingaikin:
## Description
As we'd discussed on the latest meeting, we decided not to clear the IP context on the NSC side.  See at https://github.com/networkservicemesh/deployments-k8s/issues/9888#issuecomment-1906520918

## Issue link
Closes  https://github.com/networkservicemesh/deployments-k8s/issues/9888

## How Has This Been Tested?

- [X] Tested manually
- [ ] Have not tested

## Types of changes

- [X] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/networkservice/ipam/strictipam/server.go
+++ b/pkg/networkservice/ipam/strictipam/server.go
@@ -1,0 +1,69 @@
+// Copyright (c) 2024 Cisco and its affiliates.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


pkg/networkservice/ipam/strictipam/server.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, p := range prefixes {
+		ipPool.AddNet(p)
+	}
+	return next.NewNetworkServiceServer(
+		&strictIPAMServer{ipPool: ipPool},
+		newIPAMServer(prefixes...),
+	)
+}
+
+func (n *strictIPAMServer) areAddressesValid(addresses []string) bool {
```
LionelJouin, line 50: It depends on each NSE and IPAM needs, but here nothing prevent a NSC to request the connection with an IP that is used by another connection.

LionelJouin (approved):

pkg/networkservice/ipam/strictipam/server.go
denis-tingaikin: Yes, it's right. With this patch, we're shifting the responsibility of managing IP contexts to NSEs. I also added a new IPAM wrapper chain element that can be used for clearing unepected addresses (if someone will need this functionality in the future). At this moment, it's used only in unit tests, but it can also be used for common cases, for example, as described in the original issue: https://github.com/networkservicemesh/deployments-k8s/issues/9888 ;)

Issue #1565, opened 2023-12-11
Improve the recovery of NSM clients


denis-tingaikin:

glazychev-art:
Another problem scenario:
**Scenario 5:**
1. Create a cluster.
2. Deploy NSM
3. Deploy NSC and NSE
4. Deploy NetSvc
5. Kill NSE after cmd-nsc-init finished but before cmd-nsc finished
    Actual: connection is not restored by cmd-nsc

**Description:**
After a connection is established by the init-container, it remains unattended for some time until cmd-nsc sends a request.
During this short time, any event that disrupts the connection can occur.
Apparently, this is the responsibility of the `heal` chain element, to monitor and restore such connections.

glazychev-art:
We found that for [begin.Request()](https://github.com/networkservicemesh/sdk/blob/v1.12.0/pkg/networkservice/common/begin/client.go#L70-L86) and for [eventFactory.Request()](https://github.com/networkservicemesh/sdk/blob/v1.12.0/pkg/networkservice/common/begin/event_factory.go#L101-L120) we generally use a very similar logic. 
Over time, the amount of code increases in two places and eventually we support 2 versions.
We got an idea to use `eventFactory.Request()` from `begin.Request()`, rather than using [eventFactory.executor](https://github.com/networkservicemesh/sdk/blob/v1.12.0/pkg/networkservice/common/begin/client.go#L60) directly.

Most likely, this can also help with modifying the **retry** functionality - we can use it from the `eventFactory` (see **Changes** step 2 in the Description)

**Question:**
May we call `Close()` for **_the same connection_**  multiple times?
Or do we consider `Close()` only as an addition to `Request()` (that is the `begin` chain element skips `Close()` if the connection is not ready)?

glazychev-art:
### retry problems
I've considered adding retry after the begin chain element. 
And I found a few problems:
1. `retry` calls `next.Request()` on error
Problem case:
![retry-problem1](https://github.com/networkservicemesh/sdk/assets/17748275/55842a7d-2330-4a15-bc9b-48e8a9dfb483)
If refresh failed, it starts retries until success or context timeout. So, other calls eventFactory calls will be blocked by executor. What if we need to heal the connection?
2. `retry` calls eventFactory on error
Let's imagine that the first client Request failed.
![retry-Page-3](https://github.com/networkservicemesh/sdk/assets/17748275/b4d99450-8559-4a7f-87bd-cd4b27e296f2)
[...]

PR #1583, opened 2024-02-01
Add retry attempts to refresh

glazychev-art:
## Description
This PR adds a refresh retry functionality.
If the refresh fails, it is attempted again after a few ms (200 ms by default) 

## Issue link
https://github.com/networkservicemesh/sdk/issues/1565

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


pkg/networkservice/utils/count/client_backward.go
```diff
@@ -0,0 +1,72 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/networkservicemesh/api/pkg/api/networkservice"
+
+	"github.com/networkservicemesh/sdk/pkg/networkservice/core/next"
+)
+
+// ClientBackward checks the connection on the way back
+type ClientBackward struct {
```
denis-tingaikin, line 32: Instead of creating a new chain element, I'd suggest updating the count chain element to support the backward counters

denis-tingaikin:
@edwarnicke I think we need to go with this patch; it solves most of the scenarios that I described here: https://github.com/networkservicemesh/sdk/issues/1565. and this patch allows us to quickly schedule a check for https://github.com/networkservicemesh/cmd-forwarder-vpp/issues/664.

denis-tingaikin (approved):
Added one comment related to testing, other things look good.

pkg/networkservice/utils/count/client_backward.go
glazychev-art: Fixed

PR #1588, opened 2024-02-27
A new chain element for vl3 endpoints that resets IPContext configuration depending on dst and src addresses

NikitaSkrynnik:
## Description
Added a new chain element that resets client's IPContext configuration if its dst and src addresses are from another vl3 subnetwork

## Issue link
https://github.com/networkservicemesh/cmd-nse-vl3-vpp/issues/289

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [x] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/ipam/strictvl3ipam/server.go
+++ b/pkg/ipam/strictvl3ipam/server.go
@@ -14,8 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package strictipam provides a networkservice.NetworkService Server chain element for building an IPAM server that prevents IP context configuration out of the settings scope
-package strictipam
+// Package strictvl3ipam provides a networkservice.NetworkService Server chain element that resets IP context configuration out of the settings scope
+package strictvl3ipam
 
 import (
 	"context"
@@ -30,7 +30,7 @@ import (
 	"github.com/networkservicemesh/sdk/pkg/tools/ippool"
 )
 
-type strictIPAMServer struct {
+type strictVl3IPAMServer struct {
 	ipPool *ippool.IPPool
 	m      sync.Mutex
 }
@@ -39,7 +39,7 @@ type strictIPAMServer struct {
 func NewServer(ctx context.Context, prefixCh <-chan *ipam.PrefixResponse) networkservice.NetworkServiceServer {
 	var ipPool = ippool.New(net.IPv4len)
 
-	s := &strictIPAMServer{ipPool: ipPool}
+	s := &strictVl3IPAMServer{ipPool: ipPool}
 	go func() {
 		for prefix := range prefixCh {
 			s.m.Lock()
@@ -51,7 +51,7 @@ func NewServer(ctx context.Context, prefixCh <-chan *ipam.PrefixResponse) networ
 	return s
 }
 
-func (n *strictIPAMServer) areAddressesValid(addresses []string) bool {
+func (n *strictVl3IPAMServer) areAddressesValid(addresses []string) bool {
 	n.m.Lock()
 	defer n.m.Unlock()
 
@@ -63,7 +63,7 @@ func (n *strictIPAMServer) areAddressesValid(addresses []string) bool {
 	return true
 }
 
-func (n *strictIPAMServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
+func (n *strictVl3IPAMServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
 	srcAddrs := request.GetConnection().GetContext().GetIpContext().GetSrcIpAddrs()
 	dstAddrs := request.GetConnection().GetContext().GetIpContext().GetDstIpAddrs()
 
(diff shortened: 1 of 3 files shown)


pkg/ipam/strictvl3ipam/server.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return s
+}
+
+func (n *strictVl3IPAMServer) areAddressesValid(addresses []string) bool {
+	n.m.Lock()
+	defer n.m.Unlock()
+
+	for _, addr := range addresses {
+		if !n.ipPool.ContainsNetString(addr) {
```
glazychev-art, line 59: What if we have multiple addresses? For example, IPv4 and IPv6? Or multiple IPv4?
I would suggest that addresses are valid if at least one is valid.

pkg/ipam/strictvl3ipam/server.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/networkservicemesh/sdk/pkg/tools/ippool"
+)
+
+type strictVl3IPAMServer struct {
+	ipPool *ippool.IPPool
+	m      sync.Mutex
+}
+
+// NewServer - returns a new ipam networkservice.NetworkServiceServer that validates the incoming IP context parameters and resets them based on the validation result.
+func NewServer(ctx context.Context, prefixCh <-chan *ipam.PrefixResponse) networkservice.NetworkServiceServer {
```
glazychev-art, line 39: Could you please look at strictipam?
https://github.com/networkservicemesh/sdk/blob/b1a3e26da8573763bc4c200d2aaffb873c7947db/pkg/networkservice/ipam/strictipam/server.go#L36
As you can see it acts as a wrapper for newIPAMServer.
Can we use the same approach here?

pkg/ipam/strictvl3ipam/server.go
```diff
@@ -0,0 +1,70 @@
...
+// NewServer - returns a new ipam networkservice.NetworkServiceServer that validates the incoming IP context parameters and resets them based on the validation result.
+func NewServer(ctx context.Context, vl3IPAMServer networkservice.NetworkServiceServer, ipPool *ippool.IPPool) networkservice.NetworkServiceServer {
+	return next.NewNetworkServiceServer(
+		&strictVl3IPAMServer{ipPool: ipPool},
+		vl3IPAMServer,
+	)
+}
+
+func (s *strictVl3IPAMServer) areAddressesValid(addresses []string) bool {
+	s.ipPool.Lock()
```
glazychev-art, line 43: Why do we need to use the mutex this way?
If there are concerns about concurrency, we can use a lock inside ContainsNetString, isn't it?

pkg/ipam/strictvl3ipam/server.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return false
+}
+
+func (s *strictVl3IPAMServer) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*networkservice.Connection, error) {
+	srcAddrs := request.GetConnection().GetContext().GetIpContext().GetSrcIpAddrs()
+	dstAddrs := request.GetConnection().GetContext().GetIpContext().GetDstIpAddrs()
+
+	if !s.areAddressesValid(srcAddrs) || !s.areAddressesValid(dstAddrs) {
+		request.Connection.Context.IpContext = &networkservice.IPContext{}
```
glazychev-art, line 63: This is a wrong way. 
A vl3 client can come to the vl3 server on the first request, and this request will already contain the src address of the vl3 client.
This way we will delete addresses that we should not have deleted. 
Perhaps we can only work with dst address.
Please run the vl3 example and see what requests vl3-server works with.

pkg/networkservice/chains/nsmgr/vl3_test.go
```diff
@@ -509,3 +511,73 @@ func Test_FloatingInterdomain_vl3_dns(t *testing.T) {
...
+	prefix2 := "10.10.0.0/24"
+
+	serverPrefixCh <- &ipam.PrefixResponse{Prefix: prefix1}
+	strictIPPool := ippool.NewWithNetString(prefix1)
+
+	_ = domain.Nodes[0].NewEndpoint(
+		ctx,
+		nseReg,
+		sandbox.GenerateTestToken,
+		strictvl3ipam.NewServer(ctx, vl3.NewServer(ctx, serverPrefixCh), strictIPPool),
```
glazychev-art, line 547: can we make it easier to use?

For example:
`strictvl3ipam.NewServer(ctx, vl3.NewServer, ipPool)`
Like this
https://github.com/networkservicemesh/sdk/blob/b1a3e26da8573763bc4c200d2aaffb873c7947db/pkg/networkservice/chains/nsmgr/suite_test.go#L244

pkg/networkservice/connectioncontext/ipcontext/vl3/metadata.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Cisco and/or its affiliates.
+// Copyright (c) 2022-2024 Cisco and/or its affiliates.
```
glazychev-art, line 1: unnecessary change?

NikitaSkrynnik: this is because I changed the name of this file

denis-tingaikin (approved):

pkg/tools/ippool/ippool.go
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) 2021-2022 Doc.ai and/or its affiliates.
 //
-// Copyright (c) 2022-2023 Cisco and/or its affiliates.
+// Copyright (c) 2022-2024 Cisco and/or its affiliates.
```
glazychev-art, line 3: Do we need changes in this file?

glazychev-art (approved):

Issue #1594, opened 2024-03-13
Option to disable control characters in logs


bszirtes:
It would be nice if there were an option to disable coloring and remove control characters from the logs.

denis-tingaikin:
Hello @bszirtes ,

We think that we can just remove some control characters in logs without options.

Could you also attach an example of logs with colouring and control characters that you would like to remove?

bszirtes:
Hello @denis-tingaikin,

The control characters we want to exclude from the logs are closely related to the coloring, to the best of my knowledge. We would like to avoid the following:

```
Feb 22 06:38:57.984.[36m [INFO] [cmd:/bin/forwarder] .[0mStartup completed in 7.874008959s
```

Instead, what we would like to have, are the plain logs:

```
Nov 15 14:39:42.385 [INFO] [cmd:/bin/forwarder] Startup completed in 5.65923051
```

denis-tingaikin:
As far as I know, we don't have colouring. Also, I suggest just removing all control symbols and not adding new options.

PR #1595, opened 2024-03-14
Disable log coloring

bszirtes:
## Description

Introduced an option to disable the coloring of the log messages and thus remove the control characters responsible for it, via an optionally defined environmental variable.

## Issue link

https://github.com/networkservicemesh/sdk/issues/1594

## How Has This Been Tested?

- [X] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [X] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/log/logruslogger/formatter.go
+++ b/pkg/tools/log/logruslogger/formatter.go
@@ -1,4 +1,6 @@
-// Copyright (c) 2021 Doc.ai and/or its affiliates.
+// Copyright (c) 2021-2024 Doc.ai and/or its affiliates.
+//
+// Copyright (c) 2024 Nordix Foundation.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -33,6 +35,7 @@ type formatter struct {
 func newFormatter() *formatter {
 	f := formatter{}
 	f.nf.FieldsOrder = []string{"id", "name"}
+	f.nf.NoColors = true
 	return &f
 }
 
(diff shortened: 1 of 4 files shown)


pkg/tools/log/logruslogger/formatter.go
```diff
@@ -33,6 +34,9 @@ type formatter struct {
 func newFormatter() *formatter {
 	f := formatter{}
 	f.nf.FieldsOrder = []string{"id", "name"}
+	if os.Getenv("LOG_NOCOLORS") == "true" {
+		f.nf.NoColors = true
+	}
```
denis-tingaikin, line 39: Let's use this behaviour by default. I don't see any problems on our side. So if it's also fine for your side, let's do it.

bszirtes: Are you suggesting removing the verification process and adopting the `f.nf.NoColors = true` behavior as the default?

denis-tingaikin: Yes, it's correct. I don't see any motivation for using these control symbols so we can just cut it.

bszirtes: That would be fine from our side, I will update my commit accordingly.

bszirtes: Hello @denis-tingaikin,

I  edited my commit and changed the default behavior based on the discussion. However, some of the CI tests failed, and upon reviewing the results, it seems that some expected outputs explicitly contain the control characters that I have removed.

```
...
--- FAIL: TestOutput (0.00s)
    common_test.go:73: 
        	Error Trace:	/home/runner/work/sdk/sdk/pkg/networkservice/core/trace/traceconcise/common_test.go:73
        	Error:      	Not equal: 
        	            	expected: "\x1b[36m [INFO] [id:conn-1] [type:networkService] \x1b[0mserver-reque...
...
 ```
Do you have any suggestions on how to proceed?

Comment (denis-tingaikin) [reply pkg/tools/log/logruslogger/formatter.go]: Since we're removing control characters from logs, we should also remove them from expectations in tests.

File: pkg/tools/log/logruslogger/formatter.go [hunk 2]
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Doc.ai and/or its affiliates.
+// Copyright (c) 2024 Doc.ai and/or its affiliates.
```
glazychev-art, line 1: It's better to use a year range (2021-2024) for these cases.
Please check other modified files

bszirtes: I have made updates to the headers.

glazychev-art (approved):

PR #1599, opened 2024-03-22
Pass dial timeout as option

glazychev-art:
## Description
Several options were missed

## Issue link
https://github.com/networkservicemesh/deployments-k8s/issues/11372

## How Has This Been Tested?

- [ ] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/registry/chains/client/option.go
+++ b/pkg/registry/chains/client/option.go
@@ -1,5 +1,7 @@
 // Copyright (c) 2021-2022 Doc.ai and/or its affiliates.
 //
+// Copyright (c) 2024 Cisco and/or its affiliates.
+//
 // SPDX-License-Identifier: Apache-2.0
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -18,6 +20,7 @@ package client
 
 import (
 	"net/url"
+	"time"
 
 	"google.golang.org/grpc"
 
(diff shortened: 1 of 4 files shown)


pkg/registry/chains/client/option.go
```diff
@@ -100,4 +110,5 @@ type clientOptions struct {
 	nsAdditionalFunctionality  []registry.NetworkServiceRegistryClient
 	nseAdditionalFunctionality []registry.NetworkServiceEndpointRegistryClient
 	dialOptions                []grpc.DialOption
+	dialTimeout                time.Duration
```
denis-tingaikin, line 113: Can we use a default value for the dialTimeout for the case when dialTimeout is not passed?

glazychev-art: We have default values ​​in the chain element. 
https://github.com/networkservicemesh/sdk/blob/f5e9a5c9a2290a69f80fd18f14b5604649574916/pkg/registry/common/dial/nse_client.go#L172
Do you think that additional ones need to be defined?

denis-tingaikin: As I can see, if `WithDialTimeout` is not passed to the chains/client then the dial timeout here  would be zero: https://github.com/networkservicemesh/sdk/blob/f5e9a5c9a2290a69f80fd18f14b5604649574916/pkg/registry/common/dial/nse_client.go#L172

denis-tingaikin (approved):

Issue #1621, opened 2024-05-07
Memory leak in authorize


denis-tingaikin:
## Motivation

![image](https://github.com/networkservicemesh/sdk/assets/49399980/45f24361-ce54-4e86-b362-90b060ec3b53)

Context https://github.com/networkservicemesh/cmd-nsmgr/issues/675

Affected: networkservice client/server, registry nse/ns client/server

denis-tingaikin:
Seems like done. Feel free to re-open if it's not.

PR #1623, opened 2024-05-14
Fix some leaks in authorize chain elements

NikitaSkrynnik:
## Description
Manage values in sync maps a bit better in case of errors

## Issue link
https://github.com/networkservicemesh/sdk/issues/1621

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [x] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI


denis-tingaikin (commented):
In general, LGTM


It seems like we have access to the` SpffeIDConnectionMap`, so we can add a few units to the test and make sure that in problem scenarios, the map has the expected count. It may also avoid regressions.

denis-tingaikin (approved):

PR #1638, opened 2024-05-30
Turn off tracing.

VitalyGushin:


denis-tingaikin (commented):
Consider just not using trace in the `chain` package based on TELEMETRY env. This is not looking perfect, but it could be plan B.

Issue #1652, opened 2024-07-29
Option to use Prometheus metrics without OpenTelemetry


bszirtes:
## Overview

It would be beneficial to have an option to use Prometheus Metrics natively, without requiring OpenTelemetry.

PR #1653, opened 2024-07-29
Add an option to expose Prometheus metrics via http/s server

bszirtes:
## Description

Added a new server that can be utilized to expose Prometheus metrics on via http or https.

## Issue link

https://github.com/networkservicemesh/sdk/issues/1652

## How Has This Been Tested?

- [X] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [X] New functionality
- [ ] Documentation
- [ ] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/prometheus/server.go
+++ b/pkg/tools/prometheus/server.go
@@ -1,0 +1,109 @@
+// Copyright (c) 2024 Nordix Foundation.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package prometheus provides a set of utilities for assisting with Prometheus data
+package prometheus
+
+import (
+	"context"
+	"crypto/tls"
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/spiffe/go-spiffe/v2/spiffetls/tlsconfig"
+	"github.com/spiffe/go-spiffe/v2/workloadapi"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/log"
+)
+
+// ListenAndServe gathers the certificate and initiates the server to begin handling incoming requests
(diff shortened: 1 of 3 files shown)


denis-tingaikin (commented):
I think we can add a separate option to export prometheus metrics.

Question: Have you tried to use https://opentelemetry.io/docs/languages/js/exporters/#use-otlp-exporter-with-httpjson-or-httpprotobuf?

Just wondering what other benefits there are to running it as a separate HTTP server.

bszirtes:
Hi @denis-tingaikin,

As I understand our situation, we intend to have an option to completely eliminate the usage of OpenTelemetry in our solution.

denis-tingaikin:
OK, that makes sense to me, and as I recall, @edwarnicke agrees to add a separate option for exporting metrics. 

Currently, we're thinking about the technical and design parts of the solution since we'll need to apply it for all repositories. I'll return back with more details in a few days.

bszirtes:
Hi @denis-tingaikin,

I've moved the Prometheus-related code to a new directory, as you suggested.

pkg/tools/prometheus/server.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/networkservicemesh/sdk/pkg/tools/log"
+)
+
```
denis-tingaikin, line 32: Since we plan to use it in all repos where we can grab metrics at this moment (forwarder ovs/sriov/vpp), I think it's important to move the logic with listen and serve to this package to make it easier to update:


```suggestion
func ListenAndServe(ctx context.Context, listenOn string) {

metricsServer := prometheus.Server{
			IP:            cfg.PrometheusIP,
			Port:          int(cfg.PrometheusPort),
			HeaderTimeout: cfg.PrometheusServerHeaderTimeout,
		}

		tlsConfig := &tls.Config{
			MinVersion: tls.VersionTLS12,
		}
		source, err := workloadapi.NewX509Source(ctx)
		if err != nil {
			log.FromContext(ctx).Fatalf("error getting x509 source: %v", err.Error())
		}
		tlsConfig.GetCertificate = tlsconfig.GetCertificate(source)
		metricsServer.TLSConfig = tlsConfig

		select {
		case <-ctx.Done():
			err = source.Close()
			log.FromContext(ctx).Errorf("unable to close x509 source: %v", err.Error())
		default:
		}

		
		err := metricsServer.Start(ctx)
		if err != nil {
			log.FromContext(ctx).Error(err.Error())
			cancel()
		}
}
```

Then we'll able to use this in other repos:
```go
if prometheus.IsEnabled() {
   go prometheus.ListenAndServe(ctx, listenOn, ...)
}
```

bszirtes:
Hi @denis-tingaikin,

I've moved the ListernAndServe function here, as you suggested.

denis-tingaikin (approved):
Looks good to me.

denis-tingaikin (approved):
I'm going to merge these.

If you get any comments, please feel free to leave them here, and we'll respond in a separate PR.

PR #1675, opened 2024-10-03
Add boundary check to SetupLogLevelChange

arp-est:
## Description
Add a few additional checks to the setup according to the comments below

Though I'm not super confident whether these changes make it any better, but I'm out of ideas.
It might be worth reducing the array size to fixed 2, as there's not much else than SIGUSR1, and SIGUSR2 for usage.
Also checking whether the values are the same in the map came out a bit too long, not sure if worth it.
Also there's only a single log line indicating that this thing is disabled, it could be a bit confusing, if someone gets used to the feature and doesn't see any log messages when he sends a signal.
So not sure what do you think?

## Issue link
https://github.com/networkservicemesh/cmd-forwarder-vpp/pull/1171#issuecomment-2376939709
https://github.com/networkservicemesh/deployments-k8s/issues/12296

## How Has This Been Tested?

- [x] Tested manually
- [ ] Have not tested

## Types of changes

- [ ] Bug fix
- [ ] New functionality
- [ ] Documentation
- [x] Refactoring
- [ ] CI

### DIFF
--- a/pkg/tools/log/logruslogger/levelchange.go
+++ b/pkg/tools/log/logruslogger/levelchange.go
@@ -28,6 +28,16 @@ import (
 
 // SetupLevelChangeOnSignal sets the loglevel to the one specified in the map when a signal assotiated to it arrives
 func SetupLevelChangeOnSignal(ctx context.Context, signals map[os.Signal]logrus.Level) {
+	var currentLevelCount int
+	for _, v := range signals {
+		if v == logrus.GetLevel() {
+			currentLevelCount++
+		}
+	}
+	if currentLevelCount == len(signals) {
+		log.FromContext(ctx).WithField("logruslogger", "SetupLevelChangeOnSignal").Warn("Detected that log level will never change, disabling loglevel change on signal")
+		return
+	}
 	sigChannel := make(chan os.Signal, len(signals))
 	for sig := range signals {
 		signal.Notify(sigChannel, sig)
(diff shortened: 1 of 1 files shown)


pkg/tools/log/logruslogger/levelchange.go
```diff
@@ -28,6 +29,21 @@ import (
...
+	for _, v := range signals {
+		vals = append(vals, v)
+	}
+	sort.Slice(vals, func(i, j int) bool { return i < j })
+	for i := 0; i < len(vals)-1; i++ {
+		if vals[i] == vals[i+1] {
+			log.FromContext(ctx).Info("duplicate values found in map, disabling loglevel change on signal")
+			return
+		}
+	}
```
denis-tingaikin, line 46:
```suggestion
I think duplicates are fine if we can do any other switch. I think we could bound this function only and only for cases when Goroutine will never do useful work.

```suggestion
var total int
for _, v := range signals{
  count[v]++
  total++
}

if total == count[logurs.GetLevel()] { 
 
 log.FromContext(ctx).WithField("logruslogger","SetupLevelChangeOnSignal").Warn("detected that log level will never changed, disabling loglevel change on signal")
  return
}
```

Thoughts?

arp-est: oh yeah, I'll edit it to get it through the ci

denis-tingaikin (approved):
LGTM

